PR #1, opened 2019-04-22
Forking network proxy proto-type into the apiserver-network-proxy repo.

cheftako:
This merges the initial prototype code for network proxy into the repo. 
The prototype is for https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190226-network-proxy.md
/sig cloud-provider
/assign @anfernee


cmd/agent/main.go
```diff
@@ -0,0 +1,185 @@
+package main
```
andrewsykim: Do we need license headers here?

cheftako: Yes. Want to get your first commit to this repo?

cheftako:
So the general idea is that there would be a proxy server co-located with each Kubernetes API Server (KAS). Currently we would have a proxy agent running in each Node connecting to each proxy server. One obvious improvement (which I will now go and file, thank you) is that a proxy agent should be able to connect to multiple proxy servers, so we don't need 3 proxy agents per Node. (One for each master in the HA). The slightly longer term issue to solve is that for LARGE clusters we do not need every Node to run a proxy agent. The routing portion of running sparse proxy agents on the cluster is a solved problem. The issue then is that for large clusters you want something like a DaemonSet but with sparse distribution. This is currently considered to be a Deployment issue on the cluster and as such not something which can be fixed in this repo.

PR #7, opened 2019-04-24
Run gofmt on all go files

mcrute:
This change runs gofmt on all of the go files (was a side-effect of vim-go and adding copyright statements) but seems worth merging since it's the standard for go style.

/kind cleanup
/assign @cheftako


andrewsykim:
Can we reuse the verify/update scripts in k/k here?

cheftako:
> Can we reuse the verify/update scripts in k/k for this here?

+1 But I think its a good idea to track that as a bug/issue.

PR #11, opened 2019-05-07
Adding basic image generation targets.

cheftako:
Adding simple build targets to create proxy-server and proxy-agent images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-.PHONY: gen clean certs build
+.PHONY: gen clean certs build docker/proxy-server docker/proxy-agent
 proto/agent/agent.pb.go: proto/agent/agent.proto
 	protoc -I proto proto/agent/agent.proto --go_out=plugins=grpc:proto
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -12,7 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-.PHONY: gen clean certs build
+ifndef REGISTRY
+$(error REGISTRY is not defined)
+endif
+ifndef PROJECT_ID
+$(error PROJECT_ID is not defined)
+endif
```
mcrute: Can you embed these checks into the targets that require them? The way this is written you'll have to specify these even if you have no need for them.

artifacts/images/agent-build.Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+COPY pkg/    pkg/
+COPY cmd/    cmd/
+COPY proto/  proto/
+COPY vendor/ vendor/
+
+# Build
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o proxy-agent sigs.k8s.io/apiserver-network-proxy/cmd/agent
+
+# Copy the loader into a thin image
+FROM ubuntu:latest
```
mcrute: I think these should work with `scratch` right?

Makefile
cheftako: Good point. Fixed.

artifacts/images/agent-build.Dockerfile
cheftako: Give it a try. At some point we may want an image with more tools in it.

PR #12, opened 2019-05-10
Added build target for image push.

cheftako:
Created push-images build target with a configurable docker style push utility.
Allows for both "docker" and "gcloud docker --".

Also switched from glog to klog.
This allows us to log to a file.
That allows debugging the image to be easier.
Also brings it in line with K8s processes.

Ran dep ensure to add klog which has updated all dependencies.
Should consider switching to gomod and lock down versions.


cheftako:
Let me know if you want this seperated into two commits, 1 for dep and 1 for code changes...

PR #13, opened 2019-05-15
Allow proxy agent to connect to non local server.

cheftako:
Added configuration for where the proxy server is.
Also improved cert related errors for easier debug.
Added comments to sample cert generation in make script to make it a little clearer.

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -55,9 +55,14 @@ func main() {
 }
 
 type GrpcProxyAgentOptions struct {
+	// Configuration for authenticating with the proxy-server
 	agentCert string
 	agentKey  string
 	caCert    string
+
+	// Configuration for connecting to the proxy-server
+	proxyServerHost string
+	proxyServerPort int
 }
 
 func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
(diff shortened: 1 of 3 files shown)


cmd/agent/main.go
```diff
@@ -55,47 +55,59 @@ func main() {
...
+	proxyServerHost string
+	proxyServerPort int
 }
 
 func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags := pflag.NewFlagSet("proxy-agent", pflag.ContinueOnError)
 	flags.StringVar(&o.agentCert, "agentCert", o.agentCert, "If non-empty secure communication with this cert.")
 	flags.StringVar(&o.agentKey, "agentKey", o.agentKey, "If non-empty secure communication with this key.")
 	flags.StringVar(&o.caCert, "caCert", o.caCert, "If non-empty the CAs we use to validate clients.")
+	flags.StringVar(&o.proxyServerHost, "proxyServerHost", o.proxyServerHost, "The hostname to use to connect to the proxy-server.")
```
andrewsykim: Should this be:
```
flags.StringVar(&o.proxyServerHost, "proxyServerHost", proxyServerHost, "The hostname to use to connect to the proxy-server.")
```

(default set to local host)

cmd/agent/main.go
```diff
@@ -55,47 +55,59 @@ func main() {
 }
 
 type GrpcProxyAgentOptions struct {
+	// Configuration for authenticating with the proxy-server
 	agentCert string
 	agentKey  string
 	caCert    string
+
+	// Configuration for connecting to the proxy-server
+	proxyServerHost string
```
andrewsykim: rename to `defaultProxyServerHost` to avoid confusion in above comment?

cmd/agent/main.go
cheftako: good catch

Issue #5, opened 2019-04-24
Implementing connection close


anfernee:
Proxy should be able to close connection properly, either expected or unexpected.

PR #14, opened 2019-05-23
Implement Close

anfernee:
- When agent receives CLOSE_REQ, it closes the corresponding connection identified by ConnID
- When agent's connection failed, it sends CLOSE_RESP to server to notify it.

Fixes: #5


pkg/agent/client/conn.go
```diff
@@ -102,7 +105,32 @@ func (c *conn) SetWriteDeadline(t time.Time) error {
 
 func (c *conn) Close() error {
 	klog.Info("conn.Close()")
-	return nil
+	// TODO: implement Close
```
cheftako: What is left to do?

pkg/agent/agentserver/server.go
```diff
@@ -117,31 +117,41 @@ func (s *ProxyServer) serveRecvFrontend(stream agent.ProxyService_ProxyServer, r
...
 			klog.Info("DIAL_REQ sent to backend") // got this. but backend didn't receive anything.
-		case agent.PacketType_DIAL_RSP:
-			klog.Info("received DIAL_RSP;; ignore on Server")
-		case agent.PacketType_ACK:
-			klog.Info("received ACK")
-		case agent.PacketType_CLOSE:
-			klog.Info("received CLOSE")
+
+		case agent.PacketType_CLOSE_REQ:
+			klog.Info("received CLOSE_REQ")
```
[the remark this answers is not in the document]
anfernee: This is tricky because we simply pass the packet thru. I will leave it as it. we could add trace log later.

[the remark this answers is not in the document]
cheftako: Not sure I understand. It seems like we can just print the value of pkt.Payload.CloseRequest.ConnectID. What am I missing?

pkg/agent/agentserver/server.go
```diff
@@ -117,31 +117,41 @@ func (s *ProxyServer) serveRecvFrontend(stream agent.ProxyService_ProxyServer, r
...
 			klog.Info("DIAL_REQ sent to backend") // got this. but backend didn't receive anything.
-		case agent.PacketType_DIAL_RSP:
-			klog.Info("received DIAL_RSP;; ignore on Server")
-		case agent.PacketType_ACK:
-			klog.Info("received ACK")
-		case agent.PacketType_CLOSE:
-			klog.Info("received CLOSE")
+
+		case agent.PacketType_CLOSE_REQ:
+			klog.Infof("received CLOSE_REQ(id=%d)", pkt.GetCloseRequest().ConnectID)
```
cheftako: Tempted to suggest we up the log level of close. It seems like it should be higher than normal data packets. Maybe just file an issue to review our log levels.

pkg/agent/agentserver/tunnel.go
```diff
@@ -101,6 +103,7 @@ func (t *Tunnel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	for {
 		n, err := conn.Read(pkt[:])
 		if err == io.EOF {
+			// TODO: Close remote..
```
cheftako: We should file an issue for this todo.

PR #22, opened 2019-06-05
Several improvements: unit test, logging, comments

anfernee:


cmd/client/main.go
```diff
@@ -202,6 +196,8 @@ func (c *Client) run(o *GrpcProxyClientOptions) error {
 	if err != nil {
 		return fmt.Errorf("failed to send request to client, got %v", err)
 	}
+	defer response.Body.Close() // TODO: proxy server should handle the case where Body isn't closed.
```
cheftako: If we aren't going to handle it in this PR, then we should file an issue.

pkg/agent/agentclient/client.go
```diff
@@ -47,15 +60,14 @@ func (c *connContext) cleanup() {
...
-func NewAgentClient(address string) *AgentClient {
-	a := &AgentClient{
-		connContext: make(map[int64]*connContext),
-		address:     address,
-	}
-
-	return a
-}
-
+// Connect connnects to proxy server to establish a gRPC stream,
```
cheftako: We need to think about abstracting away the gRPC piece of this. We already know we will need to support other types of streams in the near future.

pkg/agent/agentserver/server.go
```diff
@@ -39,12 +39,13 @@ func (c *ProxyClientConnection) send(pkt *agent.Packet) error {
...
+		if pkt.Type == agent.PacketType_CLOSE_RSP {
+			return c.Http.Close()
+		} else if pkt.Type == agent.PacketType_DATA {
+			_, err := c.Http.Write(pkt.GetData().Data)
+			return err
 		}
-		writer := c.Http
-		_, err := writer.Write(pkt.GetData().Data)
-		return err
+
```
cheftako: It seems like you are missing a return statement here. Does this compile?

pkg/agent/agentclient/client.go
```diff
@@ -201,11 +216,13 @@ func (a *AgentClient) remoteToProxy(conn net.Conn, connID int64) {
 
 	for {
 		n, err := conn.Read(buf[:])
+		klog.Infof("received %d bytes from proxy server", n)
```
cheftako: Might be worth switching some of these to verbose log levels? Eg. klog.V(4).Infof("received %d bytes from proxy server", n)

pkg/agent/agentclient/client.go
anfernee: right. we don't have a common interface for client. this method only works for gRPC.

cmd/client/main.go
anfernee: created #27

pkg/agent/agentclient/client.go
anfernee: created #28

pkg/agent/agentclient/client.go
cheftako: Agreed. However it is not actually part of the gRPC client code. So I believe it will eventually have to support non gRPC streams as well.

pkg/agent/agentserver/tunnel.go
```diff
@@ -99,16 +99,18 @@ func (t *Tunnel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	klog.Infof("Starting proxy to %q", r.Host)
 	pkt := make([]byte, 1<<12)
+
 	for {
 		n, err := conn.Read(pkt[:])
 		if err == io.EOF {
-			// TODO: Close remote..
+			klog.Infof("EOF from %v", r.Host)
```
cheftako: Warning?

pkg/agent/client/conn.go
```diff
@@ -50,6 +56,7 @@ func (c *conn) Write(data []byte) (n int, err error) {
 	}
 
 	klog.Infof("[tracing] send req %+v", req)
+	fmt.Printf("[tracing] send req %+v\n", req)
```
cheftako: Seems like we should be consistently using klog. If you do not direct the output to a file it should be going to stdout.

PR #23, opened 2019-06-25
Migrate to go modules

### LINKED ISSUE #17: Transition from dep to gomod
K8s has transitioned from dep to gomod. It would be good if we did as well.

mvladev:
Fixes #17


cheftako:
Waiting 1 day for any comments, will then merge.

PR #31, opened 2019-09-03
Agent supports reconnect

### LINKED ISSUE #21: Make sure proxy-agent reconnects to the proxy-server.
Make sure we handle cases like the proxy-agent coming up before the proxy-server. Also it should reconnect if the connection is broken. We should also have tests for this.

anfernee:
When the dialback is interrupted, the agent connects back to the proxy
server and continue to serve the proxy requests.

Fixes #21

The change also adds integration test for basic proxy functionality
    - The test covers both grpc and http-connect
    - The remote server is http based, so it won't send traffic volunteerly
      unless there is a request
    - The test traffic is not over TLS

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-.PHONY: gen clean certs build docker/proxy-server docker/proxy-agent push-images
+.PHONY: gen clean certs build docker/proxy-server docker/proxy-agent push-images test
 proto/agent/agent.pb.go: proto/agent/agent.proto
 	protoc -I proto proto/agent/agent.proto --go_out=plugins=grpc:proto
 
@@ -104,3 +104,6 @@ push-images: docker/proxy-agent docker/proxy-server
 
 clean:
 	rm -rf proto/agent/agent.pb.go proto/proxy.pb.go easy-rsa.tar.gz easy-rsa-master cfssl cfssljson certs bin
+
+test:
+	go test ./...
--- a/pkg/agent/agentclient/stream.go
+++ b/pkg/agent/agentclient/stream.go
@@ -40,7 +40,8 @@ type RedialableAgentClient struct {
 	opts          []grpc.DialOption
 	conn          *grpc.ClientConn
 	stopCh        chan struct{}
-	reconnTrigger chan error
+	reconnOngoing bool
+	reconnWaiters []chan error
 
 	// locks
 	sendLock   sync.Mutex
@@ -54,7 +55,7 @@ type RedialableAgentClient struct {
 	Interval time.Duration
 }
 
-func NewRedialableAgentClient(address string, opts ...grpc.DialOption) *RedialableAgentClient {
+func NewRedialableAgentClient(address string, opts ...grpc.DialOption) (*RedialableAgentClient, error) {
 	c := &RedialableAgentClient{
 		address:  address,
 		opts:     opts,
@@ -63,11 +64,7 @@ func NewRedialableAgentClient(address string, opts ...grpc.DialOption) *Redialab
 		stopCh:   make(chan struct{}),
 	}
 
-	_ = <-c.triggerReconnect()
-
-	go c.probe()
-
-	return c
+	return c, c.Connect()
 }
 
 func (c *RedialableAgentClient) probe() {
@@ -129,22 +126,26 @@ func (c *RedialableAgentClient) triggerReconnect() <-chan error {
 	c.reconnLock.Lock()
 	defer c.reconnLock.Unlock()
 
-	if c.reconnTrigger != nil {
-		return c.reconnTrigger
+	errch := make(chan error)
+	c.reconnWaiters = append(c.reconnWaiters, errch)
+
+	if !c.reconnOngoing {
+		go c.reconnect()
+		c.reconnOngoing = true
 	}
 
-	c.reconnTrigger = make(chan error)
-	go c.reconnect()
-
-	return c.reconnTrigger
+	return errch
 }
 
 func (c *RedialableAgentClient) doneReconnect(err error) {
 	c.reconnLock.Lock()
 	defer c.reconnLock.Unlock()
 
-	c.reconnTrigger <- err
-	c.reconnTrigger = nil
+	for _, ch := range c.reconnWaiters {
+		ch <- err
+	}
+	c.reconnOngoing = false
+	c.reconnWaiters = nil
 }
 
 func (c *RedialableAgentClient) Recv() (*agent.Packet, error) {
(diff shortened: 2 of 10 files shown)


yamt:
isn't grpc.Dial asynchronous by default?
i thought it dealt with reconnect by itself as well.

anfernee:
@yamt [dial](https://godoc.org/google.golang.org/grpc#Dial) doesn't seem to redial, it just returns a Conn

Makefile
```diff
@@ -1,5 +1,8 @@
 # Copyright 2019 The Kubernetes Authors.
 #
+#test:
```
cheftako: These 3 lines do no seem correct.

pkg/agent/agentclient/stream.go
```diff
@@ -0,0 +1,156 @@
...
+		klog.Info("start reconnecting")
+
+		if err2 := c.reconnect(); err2 != nil {
+			klog.Infof("reconnect failed: %v", err2)
+			return pkt, err2
+		}
+
+		return c.stream.Recv()
+	} else {
+		return pkt, err
```
cheftako: Shouldn't err be nil here? Seems more clear to return pkt, nil

pkg/agent/agentclient/stream.go
```diff
@@ -0,0 +1,156 @@
...
+
+	for retry < c.Retry {
+		if err = c.tryConnect(); err == nil {
+			klog.Info("connected")
+			return nil
+		}
+		retry++
+		klog.V(5).Infof("Failed to connect to proxy server, retry %d in %v: %v", retry, c.Interval, err)
+		klog.Infof("Failed to connect to proxy server, retry %d in %v: %v", retry, c.Interval, err)
+		time.Sleep(c.Interval)
```
cheftako: We are sleeping while holding a lock. Wondering if we always want to block everyone here. Please note that this can lead to arbitrarily long waits. For things like probe we may want it to fail faster.

pkg/agent/agentclient/stream.go
```diff
@@ -0,0 +1,156 @@
...
+	} else {
+		return pkt, err
+	}
+}
+
+func (c *RedialableAgentClient) reconnect() error {
+	c.reconnLock.Lock()
+	defer c.reconnLock.Unlock()
+
+	if c.conn != nil && c.conn.GetState() == connectivity.Ready {
```
cheftako: Why is this connectivity check inside the reconnLock, the send connectivity check in the sendLock and the probe connectivity check not in a lock at all?

pkg/agent/agentclient/stream.go
anfernee: the lock is to prevent concurrent reconnect. but you are right, we should trigger reconnect in probe instead of direct calling it.

pkg/agent/agentclient/stream.go
anfernee: This is a bit confusing, I refactor a little bit to clear it up.

anfernee:
@cheftako updated. not exactly sure what's wrong with commit message tho.

cheftako:
Have we thought through the client behaviors in various error cases? A few that come to mind, what if the client attempts to connect to a hostname which will not DNS resolve? What if nothing is listening on the relevant port, will we eventually give up and give back an error?

anfernee:
@cheftako that's a good point. we could differentiate "transient" errors from those that are not. the only non-transient error now is `EOF`. We retry all others, which I think it's fine. If nothing listens on a given port, we still retry until the port is open, this will survive server crash and come back.

cheftako:
So I think we should either extend our non transient error support or file a bug to handle it. If we misconfigure the client to connect to the proxy server which is not reachable (eg bad hostname or any other "permenant" error), I think we should blow up loudly and immediately to ease debugging.

We should also discuss how to handle "permenant" transient errors. If there is something sufficiently wrong with an agent that happens to surface as a transient error (Eg. wrong port number for the proxy server). I think we should eventually escalate the error. Obviously we could accomplish that through the healthcheck... We again should either fix it here or file a bug to track that work needs to be done.

pkg/agent/agentclient/stream.go
```diff
@@ -0,0 +1,203 @@
...
+	go c.reconnect()
+
+	return c.reconnTrigger
+}
+
+func (c *RedialableAgentClient) doneReconnect(err error) {
+	c.reconnLock.Lock()
+	defer c.reconnLock.Unlock()
+
+	c.reconnTrigger <- err
```
cheftako: I have some concerns about what happens when we have multiple clients with the channel.
Please see https://play.golang.org/p/iKP5oETRahJ.
As far as I can tell we ONLY awake the first listener on a given channel instance.

anfernee:
How about we doing this: when agent starts, we do `connect` and treat all errors critical. After that, we `reconnect` on failed connection.

pkg/agent/agentclient/stream.go
anfernee: seems like a bug. the the chan needs to be closed.

anfernee: the error msg chan needs to go thru a 1:n fanout basically. good catch.

pkg/agent/agentclient/stream.go
```diff
@@ -0,0 +1,213 @@
...
+	internalErr error
+	errChan     <-chan error
+}
+
+func (e *ReconnectError) Error() string {
+	return "transient error: " + e.internalErr.Error()
+}
+
+func (e *ReconnectError) Wait() error {
+	return <-e.errChan
```
cheftako: So I see code reading from the error channel but nothing writing to it?

anfernee: it's queued in `reconnWaiters`, and `doneReconnect()` writes to all the waiters channels.

pkg/agent/agentclient/stream.go
```diff
@@ -0,0 +1,213 @@
...
+
+	return errch
+}
+
+func (c *RedialableAgentClient) doneReconnect(err error) {
+	c.reconnLock.Lock()
+	defer c.reconnLock.Unlock()
+
+	for _, ch := range c.reconnWaiters {
+		ch <- err
```
anfernee: this is where it writes to.

pkg/agent/agentclient/stream.go
cheftako: Sorry missed that. That leaves me with the question where is it closed?

anfernee: it's not closed. if it is not used, it will be gc'ed by system.

justinsb:
As far as I can tell, this is the default GRPC behaviour: https://grpc.io/blog/grpc_on_http2/

Is there a link to the GRPC bug that we're working around here?

anfernee:
@justinsb We use a bi-directional stream on top of the gRPC connection. When the connection is down, we need to reset the stream as well: https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/agent/agentclient/stream.go#L209

PR #35, opened 2019-10-17
Add concurrency test

### LINKED ISSUE #26: e2e test with concurrency

anfernee:
Use 10 concurrent clients to send http request to remote server and download 10M bytes content from the server. Found an issue where we have concurrent write to a map. 

We could test protocols other than http as well in future.

Fix #26

### DIFF
--- a/pkg/agent/client/client.go
+++ b/pkg/agent/client/client.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"math/rand"
 	"net"
+	"sync"
 	"time"
 
 	"google.golang.org/grpc"
@@ -43,9 +44,11 @@ type dialResult struct {
 
 // grpcTunnel implements Tunnel
 type grpcTunnel struct {
-	stream      agent.ProxyService_ProxyClient
-	pendingDial map[int64]chan<- dialResult
-	conns       map[int64]*conn
+	stream          agent.ProxyService_ProxyClient
+	pendingDial     map[int64]chan<- dialResult
+	conns           map[int64]*conn
+	pendingDialLock sync.RWMutex
+	connsLock       sync.RWMutex
 }
 
 // CreateGrpcTunnel creates a Tunnel to dial to a remote server through a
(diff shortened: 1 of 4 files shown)


Sh4d1:
@anfernee do you mind rebasing to get the Travis CI running?

anfernee:
@Sh4d1 this has no conflict with master, so it should be ready to be merged.

pkg/agent/client/client.go
```diff
@@ -90,7 +93,11 @@ func (t *grpcTunnel) serve() {
 		switch pkt.Type {
 		case agent.PacketType_DIAL_RSP:
 			resp := pkt.GetDialResponse()
-			if ch, ok := t.pendingDial[resp.Random]; !ok {
+			t.pendingDialLock.RLock()
+			ch, ok := t.pendingDial[resp.Random]
+			t.pendingDialLock.RUnlock()
```
cheftako: Please use defer for the unlock. If we need to build a helper to minimize how long the lock is head, that's fine.

Sh4d1:
@anfernee it's not to resolve conflict but to ensure that the PR is run by Travis CI

anfernee:
@Sh4d1 checked.

pkg/agent/client/client.go
anfernee: @cheftako the lock is in a loop so it's not `defer`able. it could be a method to get the map, but it's just a couple of occurrences. If getting the map becomes repeated pattern, let's put it in a method.

cheftako: You could either write a getter method or an anonymous method to allow you to defer here. If something happened, such that we errored here and failed to unlock the lock, this would become very hard to debug. I would rather we just add belt and suspenders to protect ourselves.

anfernee: it's a map lookup, which shouldn't error out. but i am fine either way.

anfernee: I figured it is a little bit too verbose to add 3 methods for every map that is protected. golang doesn't error out like an exception is thrown in other languages, so as long as we don't `return` or `break` in between, it's fine.

anfernee:
@cheftako @Sh4d1 any updates?

Sh4d1:
@anfernee just need an /ok-to-test and a /testall I guess (sorry for the noise with Travis 🙄)

cheftako:
Hi @anfernee Can you remove the '#' from your commit message? I believe thats why your getting the do no merge flag.

anfernee:
thanks @cheftako ! works like a magic 👍

anfernee:
missing /lgtm ?

PR #42, opened 2019-11-26
Remove duplicate log message

caesarxuchao:
/assign @cheftako


pkg/agent/agentclient/stream.go
```diff
@@ -191,7 +191,6 @@ func (c *RedialableAgentClient) reconnect() {
 		}
 		retry++
 		klog.V(5).Infof("Failed to connect to proxy server, retry %d in %v: %v", retry, c.Interval, err)
```
Sh4d1: v4 would be better here no?

PR #43, opened 2019-12-03
Add mutex to pass "go test -race"

anfernee:
Add mutex to internal state so it's thread-safe (goroutine-safe) to access and monitor in different goroutines. Pass "go test -race".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,4 +10,4 @@ before_script:
   - go install github.com/golangci/golangci-lint/cmd/golangci-lint
 script:
   - golangci-lint run --skip-files ".*.pb.go"
-  - go test -v ./... # go test -v -race ./... is failing, should be fixed
+  - make test
--- a/pkg/agent/agentclient/client.go
+++ b/pkg/agent/agentclient/client.go
@@ -32,6 +32,7 @@ import (
 type AgentClient struct {
 	nextConnID  int64
 	connContext map[int64]*connContext
+	connLock    sync.RWMutex
 
 	stream *RedialableAgentClient
 }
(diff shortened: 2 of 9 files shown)


cheftako:
cc @caesarxuchao

pkg/agent/agentclient/client.go
```diff
@@ -153,9 +155,13 @@ func (a *AgentClient) Serve(stopCh <-chan struct{}) {
 					}
 
 					close(dataCh)
+
+					a.connLock.Lock()
 					delete(a.connContext, connID)
+					a.connLock.Unlock()
 				},
 			}
+			a.connLock.Unlock()
```
cheftako: We need to get better about using defer.

.travis.yml
```diff
@@ -10,4 +10,4 @@ before_script:
   - go install github.com/golangci/golangci-lint/cmd/golangci-lint
 script:
   - golangci-lint run --skip-files ".*.pb.go"
-  - go test -v ./... # go test -v -race ./... is failing, should be fixed
+  - go test -v -race ./...
```
Sh4d1: Move the race test inside the makefile. The prow job (travis is not used) launches `make test`

pkg/agent/agentclient/client.go
anfernee: the only way is to have a method for every state change. I hear you. Let me address it later.

pkg/agent/agentclient/client.go
```diff
@@ -171,7 +177,11 @@ func (a *AgentClient) Serve(stopCh <-chan struct{}) {
 			klog.Infof("received DATA(id=%d)", data.ConnectID)
 			klog.Infof("[tracing] %v", data)
 
-			if ctx, ok := a.connContext[data.ConnectID]; ok {
+			a.connLock.RLock()
```
zanetworker, line 180: can we somehow reduce duplicate code by putting the wrapper code in a function? 
```suggestion
			a.wrapLocks(connContext)
```

Jefftree:
This has been fixed by https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/106
/close

PR #45, opened 2019-12-11
Adding support for UDS.

cheftako:
Added flags to allow customer to specify using UDS between test-client
and server. Making sure that UDS works for both http-connect and grpc.
Also added support for clean shutdown on proxy server. This allows for
clean up of the UDS and prevents errors on restart.

### DIFF
--- a/cmd/client/main.go
+++ b/cmd/client/main.go
@@ -28,6 +28,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -68,6 +69,7 @@ type GrpcProxyClientOptions struct {
 	requestPort  int
 	proxyHost    string
 	proxyPort    int
+	proxyUdsName string
 	mode         string
 }
 
@@ -82,6 +84,7 @@ func (o *GrpcProxyClientOptions) Flags() *pflag.FlagSet {
 	flags.IntVar(&o.requestPort, "request-port", o.requestPort, "The port the request server is listening on.")
 	flags.StringVar(&o.proxyHost, "proxy-host", o.proxyHost, "The host of the proxy server.")
 	flags.IntVar(&o.proxyPort, "proxy-port", o.proxyPort, "The port the proxy server is listening on.")
+	flags.StringVar(&o.proxyUdsName, "proxy-uds", o.proxyHost, "The UDS name to connect to.")
 	flags.StringVar(&o.mode, "mode", o.mode, "Mode can be either 'grpc' or 'http-connect'.")
 
 	return flags
@@ -97,6 +100,7 @@ func (o *GrpcProxyClientOptions) Print() {
 	klog.Warningf("RequestPort set to %d.\n", o.requestPort)
 	klog.Warningf("ProxyHost set to %q.\n", o.proxyHost)
 	klog.Warningf("ProxyPort set to %d.\n", o.proxyPort)
+	klog.Warningf("ProxyUdsName set to %q.\n", o.proxyUdsName)
 	klog.Warningf("Mode set to %q.\n", o.mode)
 }
 
@@ -134,8 +138,17 @@ func (o *GrpcProxyClientOptions) Validate() error {
 	if o.proxyPort > 49151 {
 		return fmt.Errorf("please do not try to use ephemeral port %d for the proxy server port", o.proxyPort)
 	}
-	if o.proxyPort < 1024 {
+	if o.proxyPort < 1024 && o.proxyUdsName == "" {
 		return fmt.Errorf("please do not try to use reserved port %d for the proxy server port", o.proxyPort)
+	}
+	if o.proxyUdsName != "" {
+		if o.proxyPort != 0 {
+			return fmt.Errorf("please do set proxy server port to 0 not %d when using UDS", o.proxyPort)
+		}
+		if o.clientKey != "" || o.clientCert != "" || o.caCert != "" {
+			return fmt.Errorf("please do set cert materials when using UDS, key = %s, cert = %s, CA = %s",
+				o.clientKey, o.clientCert, o.caCert)
+		}
 	}
 	return nil
 }
@@ -151,6 +164,7 @@ func newGrpcProxyClientOptions() *GrpcProxyClientOptions {
 		requestPort:  8000,
 		proxyHost:    "localhost",
 		proxyPort:    8090,
+		proxyUdsName: "",
 		mode:         "grpc",
 	}
 	return &o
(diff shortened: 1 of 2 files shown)


cmd/client/main.go
```diff
@@ -28,6 +28,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"time"
```
cheftako: Please remove

cmd/proxy/main.go
```diff
@@ -17,24 +17,26 @@ limitations under the License.
...
-	"net"
-	"net/http"
-	"os"
-
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"io/ioutil"
```
cheftako: Please fix ordering

cmd/client/main.go
```diff
@@ -135,7 +139,14 @@ func (o *GrpcProxyClientOptions) Validate() error {
 		return fmt.Errorf("please do not try to use ephemeral port %d for the proxy server port", o.proxyPort)
 	}
 	if o.proxyPort < 1024 {
-		return fmt.Errorf("please do not try to use reserved port %d for the proxy server port", o.proxyPort)
+		if o.proxyUdsName == "" {
+			return fmt.Errorf("please do not try to use reserved port %d for the proxy server port", o.proxyPort)
+		}
+	}
+	if o.proxyUdsName != "" {
```
caesarxuchao: nit: move the two if's to one line to reduce indention

cmd/client/main.go
```diff
@@ -135,7 +139,14 @@ func (o *GrpcProxyClientOptions) Validate() error {
 		return fmt.Errorf("please do not try to use ephemeral port %d for the proxy server port", o.proxyPort)
 	}
 	if o.proxyPort < 1024 {
-		return fmt.Errorf("please do not try to use reserved port %d for the proxy server port", o.proxyPort)
+		if o.proxyUdsName == "" {
```
caesarxuchao: nit: move the two if's to one line to reduce indention

cmd/client/main.go
```diff
@@ -151,6 +162,7 @@ func newGrpcProxyClientOptions() *GrpcProxyClientOptions {
 		requestPort:  8000,
 		proxyHost:    "localhost",
 		proxyPort:    8090,
```
caesarxuchao: nit: Can you set the default proxy port to 0 if proxyUdsName!="", to save the user from explicitly setting --proxy-port=0?

cmd/client/main.go
```diff
@@ -211,6 +223,84 @@ func (c *Client) run(o *GrpcProxyClientOptions) error {
...
+	var proxyConn net.Conn
+	var err error
+
+	// Setup signal handler
+	ch := make(chan os.Signal, 1)
+	signal.Notify(ch)
+
+	go func() {
+		<-ch
+		err := proxyConn.Close()
```
caesarxuchao: Check if proxyConn is nil?

cmd/client/main.go
```diff
@@ -211,6 +223,84 @@ func (c *Client) run(o *GrpcProxyClientOptions) error {
...
+		klog.Infof("connection closed: %v", err)
+	}()
+
+	switch o.mode {
+	case "grpc":
+		dialOption := grpc.WithDialer(func(string, time.Duration) (net.Conn, error) {
+			// Ignoring addr and timeout arguments:
+			// addr - comes from the closure
+			// timeout - is ignored since we are connecting in a non-blocking configuration
```
caesarxuchao: `client.CreateGrpcTunnel` is blocking, right?

cmd/client/main.go
```diff
@@ -211,6 +223,84 @@ func (c *Client) run(o *GrpcProxyClientOptions) error {
 }
 
 func (c *Client) getDialer(o *GrpcProxyClientOptions) (func(ctx context.Context, network, addr string) (net.Conn, error), error) {
+	if o.proxyUdsName != "" {
+		return c.getUDSDialer(o)
+	}
+	return c.getMTLSDialer(o)
+}
+
+func (c *Client) getUDSDialer(o *GrpcProxyClientOptions) (func(ctx context.Context, network, addr string) (net.Conn, error), error) {
```
caesarxuchao: This has a lot of code duplication. To support UDS, for the grpc mode, we need to create a different dialOption, and for the http-Connect mode, we should use `net.Dial("unix", o.proxyUdsName)` instead of the `tls.Dial`. Can we just do those?

cmd/client/main.go
caesarxuchao: Add a check to verify that clientCert, CA, key are not set.

cmd/proxy/main.go
```diff
@@ -180,6 +202,7 @@ func newProxyRunOptions() *ProxyRunOptions {
 		clusterKey:    "",
 		clusterCaCert: "",
 		mode:          "grpc",
+		udsName:       "",
 		serverPort:    8090,
```
caesarxuchao: Same comment here, can we default the serverPort to 0 if udsName is not empty?

cmd/proxy/main.go
```diff
@@ -202,15 +225,18 @@ func newProxyCommand(p *Proxy, o *ProxyRunOptions) *cobra.Command {
...
 	o.Print()
 	if err := o.Validate(); err != nil {
 		return fmt.Errorf("failed to validate server options with %v", err)
 	}
+	ctx, cancel := context.WithCancel(context.Background())
 	server := agentserver.NewProxyServer()
 
 	klog.Info("Starting master server for client connections.")
-	err := p.runMasterServer(o, server)
+	masterStop, err := p.runMasterServer(ctx, o, server)
```
caesarxuchao: Is the masterStop a necessary piece for UDS to clean up, or is that a separate improvement?

cmd/proxy/main.go
```diff
@@ -227,25 +253,95 @@ func (p *Proxy) run(o *ProxyRunOptions) error {
 		return fmt.Errorf("failed to run the admin server: %v", err)
 	}
 
-	stopCh := make(chan struct{})
+	stopCh := SetupSignalHandler()
```
caesarxuchao: Why not set it up the same way as in the client code? That seems cleaner.
```golang
ch := make(chan os.Signal, 1)
signal.Notify(ch)
go func() {
	<-ch
	klog.Info("Shutting down server.")
	if masterStop != nil {
		masterStop()
	}
        cancel()
}()

cmd/client/main.go
cheftako: Done

cmd/client/main.go
cheftako: Not merging if's as I am following your second suggestion :D

cmd/client/main.go
cheftako: Goot catch

cmd/proxy/main.go
cheftako: Done

cmd/proxy/main.go
cheftako: Stealing from controller runtime. I think its a good idea for us to move in that direction.

cmd/client/main.go
cheftako: While I agree there is a bunch of code duplication, there are also a lot of small differences, more that just the Dial method. I would suggest we merge what works and then look at trying to unify it later.

cmd/client/main.go
cheftako: Missed the recv(). Fixed the comment. Leaving the 0 timeout.

cmd/proxy/main.go
cheftako: It is necessary for the graceful shutdown of UDS. It also improves the shutdown for grpc but that is a side effect.

cmd/client/main.go
cheftako: I like the idea but would rather we did it in a separate PR. We should restructure the start to have a defaulting layer for that.

PR #46, opened 2019-12-14
Make proxy work even if proxy server is HA and is behind a LB

caesarxuchao:
This PR makes the proxy works for the following setup:

We have an HA k8s master, each kube-apiserver instance is accompanied by a proxy server instance. We need to ensure that every proxy server instance is connected to at least one proxy agent, so that every kube-apiserver instance can reach to the cluster nodes.

In some k8s deployments, the proxy server instances are behind a LB. Upon receiving a Connect RPC call from the proxy agent, the LB randomly picks an instance of the HA proxy server, and forwards the request to it. Thus, the current code cannot guarantee that every proxy server has connections to at least one proxy agents.

This PR makes the proxy agent establish connection with all instances of the proxy server. To overcome the randomness caused by the LB, the proxy agent tracks the proxy server ID when it makes connection, and it keeps making connections until it has connections to all proxy servers. If there are multiple connections to the same proxy server instance, the proxy agent only keeps the oldest one.

/assign @Jefftree @cheftako 
/cc @anfernee

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -64,8 +64,21 @@ type GrpcProxyAgentOptions struct {
 	proxyServerHost string
 	proxyServerPort int
 
-	agentID      string
-	syncInterval int
+	agentID           string
+	syncInterval      int
+	probeInterval     int
+	reconnectInterval int
+}
+
+func (o *GrpcProxyAgentOptions) ClientSetConfig(dialOption grpc.DialOption) *agentclient.ClientSetConfig {
+	return &agentclient.ClientSetConfig{
+		Address:           fmt.Sprintf("%s:%d", o.proxyServerHost, o.proxyServerPort),
+		AgentID:           o.agentID,
+		SyncInterval:      time.Duration(o.syncInterval) * time.Second,
+		ProbeInterval:     time.Duration(o.probeInterval) * time.Second,
+		ReconnectInterval: time.Duration(o.reconnectInterval) * time.Second,
+		DialOption:        dialOption,
+	}
 }
 
 func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
@@ -77,6 +90,8 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.IntVar(&o.proxyServerPort, "proxy-server-port", o.proxyServerPort, "The port the proxy server is listening on.")
 	flags.StringVar(&o.agentID, "agent-id", o.agentID, "The unique ID of this agent. Default to a generated uuid if not set.")
 	flags.IntVar(&o.syncInterval, "sync-interval", o.syncInterval, "The seconds by which the agent periodically checks that it has connections to all instances of the proxy server.")
+	flags.IntVar(&o.probeInterval, "probe-interval", o.probeInterval, "The seconds by which the agent periodically checks if its connections to the proxy server are ready.")
+	flags.IntVar(&o.reconnectInterval, "reconnect-interval", o.reconnectInterval, "The seconds by which the agent tries to reconnect.")
 	return flags
 }
 
@@ -88,6 +103,8 @@ func (o *GrpcProxyAgentOptions) Print() {
 	klog.Warningf("ProxyServerPort set to %d.\n", o.proxyServerPort)
 	klog.Warningf("AgentID set to %s.\n", o.agentID)
 	klog.Warningf("SyncInterval set to %d.\n", o.syncInterval)
+	klog.Warningf("ProbeInterval set to %d.\n", o.probeInterval)
+	klog.Warningf("ReconnectInterval set to %d.\n", o.reconnectInterval)
 }
 
 func (o *GrpcProxyAgentOptions) Validate() error {
@@ -120,13 +137,15 @@ func (o *GrpcProxyAgentOptions) Validate() error {
 
 func newGrpcProxyAgentOptions() *GrpcProxyAgentOptions {
 	o := GrpcProxyAgentOptions{
-		agentCert:       "",
-		agentKey:        "",
-		caCert:          "",
-		proxyServerHost: "127.0.0.1",
-		proxyServerPort: 8091,
-		agentID:         uuid.New().String(),
-		syncInterval:    5,
+		agentCert:         "",
+		agentKey:          "",
+		caCert:            "",
+		proxyServerHost:   "127.0.0.1",
+		proxyServerPort:   8091,
+		agentID:           uuid.New().String(),
+		syncInterval:      5,
+		probeInterval:     5,
+		reconnectInterval: 5,
 	}
 	return &o
 }
@@ -173,7 +192,8 @@ func (a *Agent) runProxyConnection(o *GrpcProxyAgentOptions) error {
 		return err
 	}
 	dialOption := grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig))
-	cs := agentclient.NewAgentClientSet(fmt.Sprintf("%s:%d", o.proxyServerHost, o.proxyServerPort), o.agentID, time.Duration(o.syncInterval)*time.Second, dialOption)
+	cc := o.ClientSetConfig(dialOption)
+	cs := cc.NewAgentClientSet()
 	cs.Serve()
 
 	return nil
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -22,7 +22,9 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"time"
 
+	"github.com/google/uuid"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -61,6 +63,9 @@ type GrpcProxyAgentOptions struct {
 	// Configuration for connecting to the proxy-server
 	proxyServerHost string
 	proxyServerPort int
+
+	agentID      string
+	syncInterval int
 }
 
 func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
@@ -70,6 +75,8 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.caCert, "ca-cert", o.caCert, "If non-empty the CAs we use to validate clients.")
 	flags.StringVar(&o.proxyServerHost, "proxy-server-host", o.proxyServerHost, "The hostname to use to connect to the proxy-server.")
 	flags.IntVar(&o.proxyServerPort, "proxy-server-port", o.proxyServerPort, "The port the proxy server is listening on.")
+	flags.StringVar(&o.agentID, "agent-id", o.agentID, "The unique ID of this agent. Default to a generated uuid if not set.")
+	flags.IntVar(&o.syncInterval, "sync-interval", o.syncInterval, "The seconds by which the agent periodically checks that it has connections to all instances of the proxy server.")
 	return flags
 }
 
@@ -79,6 +86,8 @@ func (o *GrpcProxyAgentOptions) Print() {
 	klog.Warningf("CACert set to \"%s\".\n", o.caCert)
 	klog.Warningf("ProxyServerHost set to \"%s\".\n", o.proxyServerHost)
 	klog.Warningf("ProxyServerPort set to %d.\n", o.proxyServerPort)
+	klog.Warningf("AgentID set to %s.\n", o.agentID)
+	klog.Warningf("SyncInterval set to %d.\n", o.syncInterval)
 }
 
 func (o *GrpcProxyAgentOptions) Validate() error {
@@ -116,6 +125,8 @@ func newGrpcProxyAgentOptions() *GrpcProxyAgentOptions {
 		caCert:          "",
 		proxyServerHost: "127.0.0.1",
 		proxyServerPort: 8091,
+		agentID:         uuid.New().String(),
+		syncInterval:    5,
 	}
 	return &o
 }
@@ -162,14 +173,8 @@ func (a *Agent) runProxyConnection(o *GrpcProxyAgentOptions) error {
 		return err
 	}
 	dialOption := grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig))
-	client, err := agentclient.NewAgentClient(fmt.Sprintf("%s:%d", o.proxyServerHost, o.proxyServerPort), dialOption)
-	if err != nil {
-		return err
-	}
-
-	stopCh := make(chan struct{})
-
-	go client.Serve(stopCh)
+	cs := agentclient.NewAgentClientSet(fmt.Sprintf("%s:%d", o.proxyServerHost, o.proxyServerPort), o.agentID, time.Duration(o.syncInterval)*time.Second, dialOption)
+	cs.Serve()
 
 	return nil
 }
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -102,7 +102,7 @@ func (o *GrpcProxyAgentOptions) Print() {
 	klog.Warningf("ProxyServerHost set to \"%s\".\n", o.proxyServerHost)
 	klog.Warningf("ProxyServerPort set to %d.\n", o.proxyServerPort)
 	klog.Warningf("AgentID set to %s.\n", o.agentID)
-	klog.Warningf("SyncInterval set to %d.\n", o.syncInterval)
+	klog.Warningf("SyncInterval set to %d seconds.\n", o.syncInterval)
 	klog.Warningf("ProbeInterval set to %d.\n", o.probeInterval)
 	klog.Warningf("ReconnectInterval set to %d.\n", o.reconnectInterval)
 }
--- a/cmd/proxy/main.go
+++ b/cmd/proxy/main.go
@@ -23,13 +23,15 @@ import (
 	"flag"
 	"fmt"
 	"io/ioutil"
-	"k8s.io/klog"
 	"net"
 	"net/http"
 	"os"
 	"os/signal"
 	"syscall"
 
+	"k8s.io/klog"
+
+	"github.com/google/uuid"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -79,6 +81,11 @@ type ProxyRunOptions struct {
 	agentPort uint
 	// Port we listen for admin connections on.
 	adminPort uint
+
+	// ID of this server.
+	serverID string
+	// Number of proxy server instances, should be 1 unless it is a HA server.
+	serverCount uint
 }
 
 func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
@@ -94,6 +101,8 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.UintVar(&o.serverPort, "server-port", o.serverPort, "Port we listen for server connections on. Set to 0 for UDS.")
 	flags.UintVar(&o.agentPort, "agent-port", o.agentPort, "Port we listen for agent connections on.")
 	flags.UintVar(&o.adminPort, "admin-port", o.adminPort, "Port we listen for admin connections on.")
+	flags.StringVar(&o.serverID, "server-id", o.serverID, "The unique ID of this server.")
+	flags.UintVar(&o.serverCount, "server-count", o.serverCount, "The number of proxy server instances, should be 1 unless it is an HA server.")
 	return flags
 }
 
@@ -109,6 +118,8 @@ func (o *ProxyRunOptions) Print() {
 	klog.Warningf("Server port set to %d.\n", o.serverPort)
 	klog.Warningf("Agent port set to %d.\n", o.agentPort)
 	klog.Warningf("Admin port set to %d.\n", o.adminPort)
+	klog.Warningf("ServerID set to %s.\n", o.serverID)
+	klog.Warningf("ServerCount set to %d.\n", o.serverCount)
 }
 
 func (o *ProxyRunOptions) Validate() error {
@@ -207,6 +218,8 @@ func newProxyRunOptions() *ProxyRunOptions {
 		serverPort:    8090,
 		agentPort:     8091,
 		adminPort:     8092,
+		serverID:      uuid.New().String(),
+		serverCount:   1,
 	}
 	return &o
 }
@@ -234,7 +247,7 @@ func (p *Proxy) run(o *ProxyRunOptions) error {
 		return fmt.Errorf("failed to validate server options with %v", err)
 	}
 	ctx, cancel := context.WithCancel(context.Background())
-	server := agentserver.NewProxyServer()
+	server := agentserver.NewProxyServer(o.serverID, int(o.serverCount))
 
 	klog.Info("Starting master server for client connections.")
 	masterStop, err := p.runMasterServer(ctx, o, server)
--- a/pkg/agent/agentclient/client.go
+++ b/pkg/agent/agentclient/client.go
@@ -33,21 +33,33 @@ type AgentClient struct {
 	nextConnID  int64
 	connContext map[int64]*connContext
 
-	stream *RedialableAgentClient
+	stream      *RedialableAgentClient
+	serverID    string // the proxy server this client connects to
+	serverCount int    // number of proxy servers, 1 if server is not HA.
+	stopCh      <-chan struct{}
 }
 
-// NewAgentClient creates an AgentClient
-func NewAgentClient(address string, opts ...grpc.DialOption) (*AgentClient, error) {
-	stream, err := NewRedialableAgentClient(address, opts...)
+func newAgentClient(address, agentID string, cs *ClientSet, opts ...grpc.DialOption) (*AgentClient, error) {
+	stream, err := NewRedialableAgentClient(address, agentID, cs, opts...)
 	if err != nil {
 		return nil, err
 	}
+	return newAgentClientWithRedialableAgentClient(stream), nil
+}
 
-	a := &AgentClient{
+func newAgentClientWithRedialableAgentClient(rac *RedialableAgentClient) *AgentClient {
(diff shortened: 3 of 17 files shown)


cmd/agent/main.go
```diff
@@ -81,6 +85,7 @@ func (o *GrpcProxyAgentOptions) Print() {
 	klog.Warningf("CACert set to \"%s\".\n", o.caCert)
 	klog.Warningf("ProxyServerHost set to \"%s\".\n", o.proxyServerHost)
 	klog.Warningf("ProxyServerPort set to %d.\n", o.proxyServerPort)
+	klog.Warningf("agentID set to %s.\n", o.agentID)
```
cheftako: Lets keep the capitalization consistent.

cmd/agent/main.go
```diff
@@ -72,6 +75,7 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.caCert, "ca-cert", o.caCert, "If non-empty the CAs we use to validate clients.")
 	flags.StringVar(&o.proxyServerHost, "proxy-server-host", o.proxyServerHost, "The hostname to use to connect to the proxy-server.")
 	flags.IntVar(&o.proxyServerPort, "proxy-server-port", o.proxyServerPort, "The port the proxy server is listening on.")
+	flags.StringVar(&o.agentID, "agent-id", o.agentID, "The unique ID of this agent.")
```
cheftako: Maybe add something about defaulting to a generated uuid if not set?

cmd/proxy/main.go
```diff
@@ -88,6 +94,8 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.UintVar(&o.serverPort, "server-port", 8090, "Port we listen for server connections on.")
 	flags.UintVar(&o.agentPort, "agent-port", 8091, "Port we listen for agent connections on.")
 	flags.UintVar(&o.adminPort, "admin-port", 8092, "Port we listen for admin connections on.")
+	flags.StringVar(&o.serverID, "server-id", o.serverID, "The unique ID of this server.")
+	flags.UintVar(&o.serverCount, "server-count", o.serverCount, "The number of proxy server instances, should be 1 unless it is a HA serer.")
```
cheftako: serer -> server or maybe cluster?

pkg/agent/agentclient/clientset.go
```diff
@@ -0,0 +1,158 @@
...
+	"fmt"
+	"sync"
+	"time"
+
+	"google.golang.org/grpc"
+	"k8s.io/klog"
+)
+
+const (
+	syncInterval = 5 * time.Second
```
cheftako: Should be tunable

pkg/agent/agentclient/clientset.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+func (cs *ClientSet) newAgentClient() (*AgentClient, error) {
+	return newAgentClient(cs.address, cs.agentID, cs, cs.dialOption)
+}
+
+// sync makes sure that #clients >= #proxy servers
+func (cs *ClientSet) sync() {
+	for {
+		time.Sleep(syncInterval)
```
cheftako: Should probably add some form of jitter

pkg/agent/agentclient/clientset.go
```diff
@@ -0,0 +1,158 @@
...
+			c.Serve()
+		}
+	}
+}
+
+func (cs *ClientSet) Serve() {
+	cs.mu.Lock()
+	defer cs.mu.Unlock()
+	for _, c := range cs.clients {
+		go c.Serve()
```
cheftako: Doesn't this mean we are going to have redundant goroutines running Serve for the same client?

pkg/agent/agentserver/server.go
```diff
@@ -56,22 +60,89 @@ func (c *ProxyClientConnection) send(pkt *agent.Packet) error {
...
+			}
+		}
+	}
+	if !found {
+		klog.Warningf("can't find conn %v for agentID %s in the backends", conn, agentID)
+	}
+}
+
+func (s *ProxyServer) randomBackend() (agent.AgentService_ConnectServer, error) {
+	s.mu.Lock()
```
Jefftree: Using a rw lock here might help with performance down the line.

pkg/agent/agentclient/stream.go
```diff
@@ -168,49 +187,142 @@ func (c *RedialableAgentClient) Recv() (*agent.Packet, error) {
...
-	for retry < c.Retry {
-		if err = c.tryConnect(); err == nil {
-			klog.Info("connected")
+	var retry, limit int
+
+	limit = retryLimit(c.serverCount)
+	for retry < limit {
+		serverID, _, err, conn, stream := c.tryConnect()
+		switch {
+		case err != nil:
```
Jefftree: Can we nest this in an if statement block? Some of the latter switch conditions are a bit difficult to read grouped within one statement.

```
if err != nil {
 ...
} else {
 ...
}
```

pkg/agent/agentclient/clientset.go
```diff
@@ -0,0 +1,158 @@
...
+	c, err := cs.newAgentClient()
+	if err != nil {
+		return err
+	}
+	cs.serverCount = c.serverCount
+	klog.Infof("added client connecting to proxy server %s", c.serverID)
+	klog.Infof("server count is %d", cs.serverCount)
+	cs.clients[c.serverID] = c
+
+	for len(cs.clients) != cs.serverCount {
```
anfernee: this for loop inside a mutex lock could cause deadlock. it could hold lock for a very long time.

pkg/agent/agentclient/clientset.go
```diff
@@ -0,0 +1,158 @@
...
+				klog.Error(err)
+				continue
+			}
+			if err := cs.AddClient(c.serverID, c); err != nil {
+				klog.Infof("closing connection: %v", err)
+				c.Close()
+				continue
+			}
+			klog.Infof("sync added client connecting to proxy server %s", c.serverID)
+			c.Serve()
```
anfernee: ```go c.Serve()```?

cmd/agent/main.go
```diff
@@ -25,6 +25,7 @@ import (
 	"net/http"
 	"os"
 
+	"github.com/google/uuid"
```
anfernee: update go.mod for the new depedency.

pkg/agent/agentserver/server.go
```diff
@@ -56,22 +60,89 @@ func (c *ProxyClientConnection) send(pkt *agent.Packet) error {
...
+	defer s.mu.Unlock()
+	if len(s.backends) == 0 {
+		return nil, fmt.Errorf("No backend available")
+	}
+	agentID := s.agentIDs[rand.Intn(len(s.agentIDs))]
```
anfernee: make rand.Rand part of server, and set seed using time.Now().

caesarxuchao:
Thank you for the reviews. I'll work on this again in a week.

anfernee:
Thanks @caesarxuchao I'll rebase my change on top of this PR.

anfernee:
There is a test failure too, it's okay to me to merge after the test is fixed.

pkg/agent/agentclient/stream.go
```diff
@@ -168,49 +187,142 @@ func (c *RedialableAgentClient) Recv() (*agent.Packet, error) {
...
 	c.doneReconnect(fmt.Errorf("Failed to connect to proxy server: %v", err))
 }
 
-func (c *RedialableAgentClient) tryConnect() error {
+func serverCount(stream agent.AgentService_ConnectClient) (int, error) {
+	md, err := stream.Header()
+	if err != nil {
+		return 0, err
+	}
+	scounts := md.Get("serverCount")
```
dberkov: Could you make it as a constant

pkg/agent/agentclient/stream.go
```diff
@@ -168,49 +187,142 @@ func (c *RedialableAgentClient) Recv() (*agent.Packet, error) {
...
-	c.conn, err = grpc.Dial(c.address, c.opts...)
+	conn, err := grpc.Dial(c.address, c.opts...)
 	if err != nil {
-		return err
+		return "", 0, nil, nil, err
 	}
 
-	c.stream, err = agent.NewAgentServiceClient(c.conn).Connect(context.Background())
-	return err
+	ctx := metadata.AppendToOutgoingContext(context.Background(), "agentID", c.agentID)
```
dberkov: could you make "agentID" as a constant

cmd/proxy/main.go
dberkov: or proxy

pkg/agent/agentclient/clientset.go
dberkov: +1

dberkov:
Proposed implementation managed clients list in 2 places:
1. At the beginning it tries to create all required connections
2. It has a sync method that continue to resync it all the time.

This approach has an issue, that if one of the masters is down during first stage, agent will never come up. ( same issue exists in current implementation. If agent start first, it immediately failed).  

Is it possible to keep "sync" method only. So all the logic starts with "sync" that always tries to keep up to amount of required connections.

pkg/agent/agentclient/clientset.go
caesarxuchao: Can you elaborate on how there will be redundant Serve()?

pkg/agent/agentclient/clientset.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+func (cs *ClientSet) clientsCount() int {
+	cs.mu.Lock()
+	defer cs.mu.Unlock()
+	return len(cs.clients)
+}
+
+func (cs *ClientSet) hasIDLocked(serverID string) bool {
+	for k, _ := range cs.clients {
```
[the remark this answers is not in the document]
caesarxuchao: Good catch.

pkg/agent/agentserver/server.go
```diff
@@ -154,12 +234,12 @@ func (s *ProxyServer) serveRecvFrontend(stream agent.ProxyService_ProxyServer, r
...
-			if s.Backend == nil {
-				klog.Info(">>> No backend found; drop")
+			if backend == nil {
+				klog.Errorf("backend has not been initialized for connID %d. Client should send a Dial Request first.", connID)
 				continue
 			}
-
-			if err := s.Backend.Send(pkt); err != nil {
+			if err := backend.Send(pkt); err != nil {
+				// TODO: retry with other backends connecting to this agent.
```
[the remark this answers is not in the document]
caesarxuchao: I think that depends on how backend.Send (which is a grpc stream's [SendMsg](https://github.com/grpc/grpc-go/blob/master/stream.go#L114)) handles error. It claims that it aborts the stream on error, so we can safely retry non-idempotent packets.

pkg/agent/agentclient/clientset.go
caesarxuchao: CreateClients is called when the agent starts, no other goroutine is going to contend for the lock, so I think it's ok.

pkg/agent/agentclient/clientset.go
caesarxuchao: Good catch.

Jefftree:
Tested this on a GKE regional cluster and works fine for the most part :)

A couple small things:

- Can we log both agent and server ID when connecting/dropping connections? It makes debugging easier and helps us identify situations where the konnectivity pod may have restarted.
- This branch has a merge conflict with @dberkov's auth PR, maybe one of you can rebase off the other?
- Add the missing library to vendor

pkg/agent/agentclient/stream.go
```diff
@@ -168,48 +207,164 @@ func (c *RedialableAgentClient) Recv() (*agent.Packet, error) {
...
+			time.Sleep(c.Interval)
+		case r.serverID != c.serverID && !c.cs.HasID(r.serverID):
+			// create a new client
+			cc := copyRedialableAgentClient(*c)
+			cc.stream = r.agentServiceClient
+			cc.conn = r.grpcConn
+			ac := &AgentClient{
+				connContext: make(map[int64]*connContext),
+				stream:      &cc,
+				serverID:    cc.serverID,
```
Jefftree: This should be `r.serverID`, else the new client would still try to reconnect to the old `cc.serverID` when attempting another reconnect.

pkg/agent/agentclient/stream.go
```diff
@@ -168,48 +207,164 @@ func (c *RedialableAgentClient) Recv() (*agent.Packet, error) {
...
+			}
+			retry++
+			klog.Infof("Trying to reconnect to proxy server %s, got connected to proxy server %s, for which there is already a connection, retry %d in %v", c.serverID, r.serverID, retry, c.Interval)
+			time.Sleep(c.Interval)
+		case r.serverID != c.serverID && !c.cs.HasID(r.serverID):
+			// create a new client
+			cc := copyRedialableAgentClient(*c)
+			cc.stream = r.agentServiceClient
+			cc.conn = r.grpcConn
+			ac := &AgentClient{
```
Jefftree: We need to initialize the `serverCount` and `stopCh` similar to https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/09c3ae4ccc1ae2a9345ddfa7aa1b91f022bad69c/pkg/agent/agentclient/client.go#L48. 

Perhaps a constructor `newAgentClient(stream * RedialableAgentClient) (*AgentClient, error)` can help us centralize the creation.

pkg/agent/agentclient/stream.go
```diff
@@ -168,48 +207,164 @@ func (c *RedialableAgentClient) Recv() (*agent.Packet, error) {
...
+		case r.serverID != c.serverID && c.cs.HasID(r.serverID):
+			// reset the connection
+			err := r.grpcConn.Close()
+			if err != nil {
+				klog.Infof("failed to close connection to %s: %v", r.serverID, err)
+			}
+			retry++
+			klog.Infof("Trying to reconnect to proxy server %s, got connected to proxy server %s, for which there is already a connection, retry %d in %v", c.serverID, r.serverID, retry, c.Interval)
+			time.Sleep(c.Interval)
+		case r.serverID != c.serverID && !c.cs.HasID(r.serverID):
```
Jefftree: In the case that a master is restarted with a different ID, we'd create a new agent and would have connections to all the masters. However, the old agent would still be attempting to reach the old master for another 15 retries (servercount=3). Should we add additional logic to break out of this loop if the desired number of master connections is reached?

caesarxuchao: I had considered this case. The cost of the connect attempt is small, so I decided not to add extra complexity to handle this case.

Jefftree: ack. It pollutes the logs a bit, but yeah I agree doesn't affect the performance by much.

Jefftree:
Can you squash commits?

/lgtm

pkg/agent/agentclient/clientset.go
```diff
@@ -0,0 +1,128 @@
...
+
+// ClientSet consists of clients connected to each instance of an HA proxy server.
+type ClientSet struct {
+	mu      sync.Mutex              //protects the following
+	clients map[string]*AgentClient // map between serverID and the client
+	// connects to this server.
+
+	agentID     string // ID of this agent
+	address     string // proxy server address. Assuming HA proxy server
+	serverCount int    // number of proxy server instances, should be 1
```
dberkov: comment is incorrect, since value "0" is a valid case

pkg/agent/agentclient/clientset.go
```diff
@@ -0,0 +1,128 @@
...
+			sleep := cs.syncInterval + time.Duration(rand.Float64()*jitter*float64(cs.syncInterval))
+			time.Sleep(sleep)
+		}
+		if cs.serverCount == 0 || cs.clientsCount() < cs.serverCount {
+			c, err := cs.newAgentClient()
+			if err != nil {
+				klog.Error(err)
+				continue
+			}
+			if cs.serverCount == 0 {
```
dberkov: is "if" case required? We can always update serverCount after new connection is established. It will allow in the future, dynamically change amount of servers

pkg/agent/agentclient/clientset.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (cs *ClientSet) newAgentClient() (*AgentClient, error) {
+	return newAgentClient(cs.address, cs.agentID, cs, cs.dialOption)
+}
+
+// sync makes sure that #clients >= #proxy servers
+func (cs *ClientSet) sync() {
+	jitter := float64(0.2)
+	for {
+		if cs.serverCount != 0 {
```
dberkov: We need add some metric around this code, so we can tune the syncInterval during beta prior going to GA, especially for regional clusters

pkg/agent/agentclient/stream.go
dberkov: Why do we need retry logic at all? What if we  just deregister server, If connection is closed.   pkg/agent/agentclient/clientset.go:sync supposed to recreate missing connections periodically

pkg/agent/agentclient/clientset.go
caesarxuchao: I don't think the comment is incorrect. "0" means it's not initialized yet.

pkg/agent/agentclient/clientset.go
caesarxuchao: Do you have any metrics in mind?

caesarxuchao:
> Why do we need retry logic at all? What if we just deregister server, If connection is closed. pkg/agent/agentclient/clientset.go:sync supposed to recreate missing connections periodically

Somehow github webpage doesn't have a reply box for this comment.

The retry logic here allows the proxy to retryin sending a packet. Otherwise it will require the upper-level application to retry after the sync loop re-establish the connection.

caesarxuchao:
Added a unit test in the last commit. PTAL.

tests/ha_proxy_server_test.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+		// go lb.handleConnection(conn, lb.randomBackend())
+		back := lb.randomBackend()
+		go lb.handleConnection(conn, back)
+	}
+}
+
+func (lb *tcpLB) addBackend(backend string) {
+	lb.mu.Lock()
+	lb.mu.Unlock()
```
Jefftree: Missing `defer`?

cmd/agent/main.go
```diff
@@ -61,6 +63,22 @@ type GrpcProxyAgentOptions struct {
 	// Configuration for connecting to the proxy-server
 	proxyServerHost string
 	proxyServerPort int
+
+	agentID           string
+	syncInterval      int
```
cheftako: Not going to hold up merge but we should decide soon if we really want this to be int seconds or a duration.

cmd/agent/main.go
```diff
@@ -79,6 +101,10 @@ func (o *GrpcProxyAgentOptions) Print() {
 	klog.Warningf("CACert set to \"%s\".\n", o.caCert)
 	klog.Warningf("ProxyServerHost set to \"%s\".\n", o.proxyServerHost)
 	klog.Warningf("ProxyServerPort set to %d.\n", o.proxyServerPort)
+	klog.Warningf("AgentID set to %s.\n", o.agentID)
+	klog.Warningf("SyncInterval set to %d.\n", o.syncInterval)
```
cheftako: To be more clear we should specify this as seconds.

pkg/agent/agentclient/client.go
```diff
@@ -33,21 +33,33 @@ type AgentClient struct {
...
+	return newAgentClientWithRedialableAgentClient(stream), nil
+}
 
-	a := &AgentClient{
+func newAgentClientWithRedialableAgentClient(rac *RedialableAgentClient) *AgentClient {
+	return &AgentClient{
 		connContext: make(map[int64]*connContext),
-		stream:      stream,
+		stream:      rac,
+		serverID:    rac.serverID,
```
cheftako: serverID, serverCount and stopCh seem to be duplicated across AgentClient and RedialableAgentClient. However the RedialableAgentClient is a field in the AgentClient. I do not believe the inner and outer struct are allowed to have different values for these fields. As such it seems like extra complexity and a source of potential bugs without providing any value.

pkg/agent/agentclient/clientset.go
```diff
@@ -0,0 +1,155 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
cheftako: 2020

pkg/agent/agentclient/clientset.go
```diff
@@ -0,0 +1,155 @@
...
+	"time"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/connectivity"
+	"k8s.io/klog"
+)
+
+// ClientSet consists of clients connected to each instance of an HA proxy server.
+type ClientSet struct {
+	mu      sync.Mutex              //protects the following
```
cheftako: All of the following or specifically the clients map?

cmd/proxy/main.go
```diff
@@ -79,6 +81,11 @@ type ProxyRunOptions struct {
 	agentPort uint
 	// Port we listen for admin connections on.
 	adminPort uint
+
+	// ID of this server.
+	serverID string
```
cheftako: What is the serverID set to when serverCount > 1?

pkg/agent/agentclient/client.go
```diff
@@ -33,21 +33,33 @@ type AgentClient struct {
 	nextConnID  int64
 	connContext map[int64]*connContext
 
-	stream *RedialableAgentClient
+	stream      *RedialableAgentClient
+	serverID    string // the proxy server this client connects to
+	serverCount int    // number of proxy servers, 1 if server is not HA.
```
cheftako: AgentClient and RedialableAgentClient seem like the wrong level for keeping the serverCount. I realize the value comes from Connect() but I think it makes more sense to plumb the value up to the AgentClientSet.

cheftako: Might make more sense to pass a admin or metadata channel in to the Clients. They can then push things like server count out to the appropriate Set or controller when something like ServerCount is set (changed)

pkg/agent/agentclient/client.go
```diff
@@ -34,20 +34,28 @@ type AgentClient struct {
...
-		stream:      stream,
+		stream:      rac,
+		stopCh:      rac.stopCh,
 	}
-	return a, nil
+}
+
+// Close closes the underlying stream.
+func (c *AgentClient) Close() {
+	c.stream.Close()
```
cheftako: What about when stream is null?

pkg/agent/agentclient/stream.go
```diff
@@ -168,48 +208,162 @@ func (c *RedialableAgentClient) Recv() (*agent.Packet, error) {
...
+		grpcConn:           conn,
+		agentServiceClient: stream,
+	}
+	return r, err
 }
 
-// interrupt interrupt the stream connection. (For testing purpose)
-func (c *RedialableAgentClient) interrupt() {
+func (c *RedialableAgentClient) Close() {
 	c.conn.Close()
```
cheftako: What about when conn is null?

cmd/agent/main.go
caesarxuchao: This field is going to hold the parsed flag value so it has to be int. I converted it to time.Duration when constructing the clients to simplify the code.

cmd/agent/main.go
caesarxuchao: Done.

pkg/agent/agentclient/client.go
caesarxuchao: Good catch. Removed the fields in AgentClient.

cmd/proxy/main.go
caesarxuchao: Server count is the number of servers in the HA setup.

ServerID is the id of this particular instance.

Do you have suggestions on how to make it more clear?

pkg/agent/agentclient/client.go
caesarxuchao: > I realize the value comes from Connect() but I think it makes more sense to plumb the value up to the AgentClientSet.

That requires plumbing through 4 functions. And because the server count is only useful to the clientset, we will have many "_, err := xxx()" in our code.

Using a channel is cleaner, and is required to deal with HA proxy servers that can dynamically resize, but also makes the code more complex. As it's not needed yet, I'll keep the current code.

caesarxuchao:
@cheftako I addressed the remaining comments. PTAL. Thanks.

cmd/agent/main.go
cheftako: I believe we do have support for parsing Duration flag fields as you can see here https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-apiserver/app/options/options.go#L161.

cmd/proxy/main.go
cheftako: proxyServerID maybe? Knowing its an HA cluster made we wonder if this might have to do with the Kube API Server.

pkg/agent/agentclient/clientset.go
cheftako: Either tries or time needed to reach desired state (cs.ClientsCount() == cs.serverCount) seems like it would be desirable. It may be thrown off if you have something like a crash looping proxy or network issue. OTOH have a metric which shows a connectivity problem of that ilk is not a bad thing.

cheftako: Maybe file a bug for this? This PR is already large.

pkg/agent/agentclient/stream.go
```diff
@@ -168,48 +208,165 @@ func (c *RedialableAgentClient) Recv() (*agent.Packet, error) {
...
+	case 2:
+		return 3 + 7
+	case 3:
+		return 3 + 12
+	case 4:
+		return 3 + 17
+	case 5:
+		return 3 + 21
+	default:
+		// we don't expect HA server with more than 5 instances.
```
cheftako: We may not expect it but the 5 limit is primarily about etcd and the proxy server count is not directly related to etcd. I'm still fine with this but we may want to add a warning log for this case.

PR #49, opened 2020-01-14
Remove vendor and change Dockerfile to support go.mod

Jefftree:
Also change grpc.WithDialer to grpc.WithContextDialer due to deprecation.


proto/proxy.pb.go
```diff
@@ -36,7 +37,7 @@ var _ = math.Inf
 // is compatible with the proto package it is being compiled against.
 // A compilation error at this line likely means your copy of the
 // proto package needs to be updated.
-const _ = proto.ProtoPackageIsVersion2 // please upgrade the proto package
+const _ = proto.ProtoPackageIsVersion3 // please upgrade the proto package
```
Jefftree: @cheftako @anfernee: I noticed that this was changed from `proto.ProtoPackageIsVersion2` to `proto.ProtoPackageIsVersion3` after removing vendor. Is there any implication of this I should be aware of? Still able to run proxy server and agent without any issue.

anfernee: Probably we move to the new proto version. I think it should be fine as long as it's consistent between agent and server

cheftako: Better to do it now while we are alpha

Jefftree:
Docker caches the installation of the modules, so after the first build (and assuming no dependency changes), all subsequent builds should use the cached image. I don't think this should add any additional time to the build. The main goal is to tidy up the code and reduce PRs where a big portion of the diff is adding vendor files.

anfernee:
Thanks for explaining @Jefftree That makes sense. 
Can you also change `make build` so that it also uses the golang container with the cached vendor to build binary?

Sh4d1:
@Jefftree could you also add go module env variable in the test rule of the makefile? Should make the CI happy 😄

Jefftree:
@anfernee: I think this is more of a design decision we need to tackle. Using the docker builds in `make build` means that we will no longer have a executable when we build (locally), and we'll need to run the containers with `docker run ...` instead of `./bin/proxy-server ...`. Technically local builds (not using docker) are already fast since all the libraries needed would already be installed in $GOPATH so I'm not sure if build time will even be an issue.

Right now we have both `make build` and `make docker-build` for building an executable and docker executable image respectively. I think it's fine to keep them separate for now.

anfernee:
it's not about build speed. Right now local build isn't working without `vendor` folder.

PR #47, opened 2019-12-18
Adding support for agent authenticate server by using kubeapi-server ca 

dberkov:
The PR allows to reuse certificate generated for kubelet by proxy/agent/test-client, instead of generating dedicated set of certificates for proxy / agent / test-agent.

### Server
Proxy-server runs as a static  pod on master and initializes a grpc server with   /etc/srv/kubernetes/pki/apiserver.crt & /etc/srv/kubernetes/pki/apiserver.key (no CA certificate)

### Agent
Agent runs on node as a regular pod. It uses existing /var/run/secrets/kubernetes.io/serviceaccount/ca.crt (ca.crt bounds in every pod by k8s) to validate proxy-server's certificate returns during tls handshake.

### Test-Clinet
As part of the tests, test-client runs on master VM as a static pod and communicates with proxy-server via UDS.

PR #51, opened 2020-01-22
Adding support for server to authenticate agent

dberkov:
The PR allows to proxy-server authenticate proxy-agent.

## Agent:
Agent sends in GRPC metadata token associated to the pod by kubernetes system.

## Server
Server as part of the connection opening step, reads the token sent by agent, invokes kubernetes.TokenReviews API, checks the token is valid and belongs to agent's pod by validating namespace + service account of the token's owner .

## General
All examples/kubernetes/* templates and READ.me procedure has been updated for supporting fully working e2e test of this feature in kubernetes.


cmd/proxy/main.go
```diff
@@ -191,22 +211,52 @@ func (o *ProxyRunOptions) Validate() error {
...
+		}
+		if o.agentServiceAccount != "" {
+			return fmt.Errorf("agentServiceAccount must be empty when authenticateAgent is set to false")
+		}
+		if o.authenticationAudience != "" {
+			return fmt.Errorf("authenticationAudience must be empty when authenticateAgent is set to false")
+		}
+	}
+
+	if o.authenticateAgent {
```
Sh4d1: nit: use else

cmd/proxy/main.go
```diff
@@ -234,7 +284,29 @@ func (p *Proxy) run(o *ProxyRunOptions) error {
 		return fmt.Errorf("failed to validate server options with %v", err)
 	}
 	ctx, cancel := context.WithCancel(context.Background())
-	server := agentserver.NewProxyServer()
+	defer cancel()
```
Sh4d1: You can remove the one line 336 then

cmd/proxy/main.go
```diff
@@ -234,7 +284,29 @@ func (p *Proxy) run(o *ProxyRunOptions) error {
...
+
+	var k8sClient *kubernetes.Clientset
+	if o.authenticateAgent {
+		config, err := clientcmd.BuildConfigFromFlags("", o.kubeconfigPath)
```
Sh4d1: Maybe add a validation for kubeconfig to fail in validate if it is empty/non existent?

pkg/agent/agentclient/stream.go
```diff
@@ -197,6 +205,32 @@ func (c *RedialableAgentClient) reconnect() {
 	c.doneReconnect(fmt.Errorf("Failed to connect to proxy server: %v", err))
 }
 
+func (c *RedialableAgentClient) initializeAuthContext(ctx context.Context) (context.Context, error) {
+	if c.saTokenFile == "" {
```
Sh4d1: This test seems useless since it's only call when the file path is not empty.

cmd/client/main.go
```diff
@@ -164,7 +164,7 @@ func newGrpcProxyClientOptions() *GrpcProxyClientOptions {
 		requestPath:  "/",
 		requestHost:  "localhost",
 		requestPort:  8000,
-		proxyHost:    "localhost",
+		proxyHost:    "127.0.0.1",
```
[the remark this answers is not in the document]
dberkov: It was a leftover from https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/47. I removed it

cmd/agent/main.go
```diff
@@ -70,6 +73,7 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.caCert, "ca-cert", o.caCert, "If non-empty the CAs we use to validate clients.")
 	flags.StringVar(&o.proxyServerHost, "proxy-server-host", o.proxyServerHost, "The hostname to use to connect to the proxy-server.")
 	flags.IntVar(&o.proxyServerPort, "proxy-server-port", o.proxyServerPort, "The port the proxy server is listening on.")
+	flags.StringVar(&o.saToken, "authorization-token", o.saToken, "If non-empty authorize each call to the proxy-server with token file's content.")
```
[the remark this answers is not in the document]
dberkov: Done. renamed to service-account-token

cmd/proxy/main.go
dberkov: done

cmd/proxy/main.go
dberkov: done

cmd/proxy/main.go
dberkov: Empty string is a valid case, since clientcmd.BuildConfigFromFlags take care for it, however, I added a validation case, to check the file exists if kubeconfigPath is provided

pkg/agent/agentclient/stream.go
dberkov: good catch. removed

pkg/agent/agentclient/stream.go
```diff
@@ -205,7 +239,13 @@ func (c *RedialableAgentClient) tryConnect() error {
 		return err
 	}
 
-	c.stream, err = agent.NewAgentServiceClient(c.conn).Connect(context.Background())
+	ctx := context.Background()
+	if c.saTokenFile != "" {
+		if ctx, err = c.initializeAuthContext(ctx); err != nil {
```
caesarxuchao: We can remove line 242 and 243

pkg/agent/agentserver/server.go
```diff
@@ -196,6 +216,73 @@ func (s *ProxyServer) serveSend(stream agent.ProxyService_ProxyServer, sendCh <-
...
+	if err != nil {
+		return fmt.Errorf("Failed to authenticate request. err:%v", err)
+	}
+
+	if r.Status.Error != "" {
+		return fmt.Errorf("lookup failed: %s", r.Status.Error)
+	}
+
+	if !r.Status.Authenticated {
+		return fmt.Errorf("lookup failed: service account jwt not valid")
```
caesarxuchao: Is this a jwt?

proto/agent/agent.proto
```diff
@@ -98,3 +98,8 @@ message Data {
     bytes data = 3;
 }
 
+// AuthenticationTokens contains a definition of auth tokens to be passed in grpc calls
+message AuthenticationTokens  {
```
caesarxuchao: Why do we need to define this as a proto message? Can we just send the token itself?

pkg/util/auth.go
```diff
@@ -0,0 +1,6 @@
+package util
+
+const (
+	// AuthenticationContextName will be used as a key to store authentication tokens in grpc call
+	AuthenticationContextName = "X-Auth"
```
caesarxuchao: Out of curiosity, is "X-Auth" a wellknown header?

proto/agent/agent.pb.go
```diff
@@ -37,7 +36,7 @@ var _ = math.Inf
 // is compatible with the proto package it is being compiled against.
 // A compilation error at this line likely means your copy of the
 // proto package needs to be updated.
-const _ = proto.ProtoPackageIsVersion3 // please upgrade the proto package
+const _ = proto.ProtoPackageIsVersion2 // please upgrade the proto package
```
caesarxuchao: Is this downgrade intentional?

examples/kubernetes/README.md
```diff
@@ -30,6 +31,30 @@ CLUSTER_CERT=/etc/srv/kubernetes/pki/apiserver.crt
 CLUSTER_KEY=/etc/srv/kubernetes/pki/apiserver.key
 ```
 
+# Register SERVER_TOKEN in [static-token-file](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#static-token-file)
```
caesarxuchao: Is "static-token" the standard way to authenticate a process running in the master node?

I think we can run the proxy server as a static pod, and then use a service account to authenticate it.

examples/kubernetes/kubeconfig
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: Config
+users:
+- name: konnectivity-server
+  user:
+    token: ${SERVER_TOKEN}
+clusters:
+- name: local
+  cluster:
+    insecure-skip-tls-verify: true
```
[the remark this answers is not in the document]
Sh4d1: This is only for the client side no?

pkg/agent/agentserver/server.go
Sh4d1: Yep sa are jwt

proto/agent/agent.proto
Sh4d1: It may be future proof to have it in it's own message IMHO

caesarxuchao: IMHO since we are not building a grpc service around this message, making it a message only makes it less flexible, so it's the opposite of future proof. @Sh4d1 can you elaborate?

examples/kubernetes/kubeconfig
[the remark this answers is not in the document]
caesarxuchao: You are right. This means that the client won't verify the server's identity, so I think it's ok.

proto/agent/agent.proto
Sh4d1: @caesarxuchao yep sorry I had a public API in mind, I deleted the comment when I realized my mistake 😅

cmd/agent/main.go
```diff
@@ -79,6 +83,7 @@ func (o *GrpcProxyAgentOptions) Print() {
 	klog.Warningf("CACert set to \"%s\".\n", o.caCert)
 	klog.Warningf("ProxyServerHost set to \"%s\".\n", o.proxyServerHost)
 	klog.Warningf("ProxyServerPort set to %d.\n", o.proxyServerPort)
+	klog.Warningf("Service account authorizationToken set to \"%s\".\n", o.saToken)
```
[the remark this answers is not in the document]
dberkov: done

cmd/agent/main.go
```diff
@@ -106,6 +111,11 @@ func (o *GrpcProxyAgentOptions) Validate() error {
 	if o.proxyServerPort <= 0 {
 		return fmt.Errorf("proxy server port %d must be greater than 0", o.proxyServerPort)
 	}
+	if o.saToken != "" {
+		if _, err := os.Stat(o.saToken); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
dberkov: We do exactly same validation for all other files (agentCert, agentKey, etc..). I would prefer to keep it as is and we can refactor entire project's validation methods in separate PR

pkg/util/auth.go
mikedanese: Why not use "Authorization: Bearer [token]"

https://tools.ietf.org/html/rfc6750#section-2.1

examples/kubernetes/konnectivity-server.yaml
```diff
@@ -19,13 +19,18 @@ spec:
       "--log-file=/var/log/konnectivity-server.log",
       "--logtostderr=false",
       "--log-file-max-size=0",
-      "--uds-name=/etc/srv/kubernetes/konnectivity/konnectivity-server.socket",
+      "--uds-name=/etc/srv/kubernetes/konnectivity-server/konnectivity-server.socket",
```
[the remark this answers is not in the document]
dberkov: We will have kubeconfig file generated in this folder (based on the pattern the folder name supposed to match component name), therefore  wanted to combine both files (kubeconfig & uds) to same folder

examples/kubernetes/konnectivity-server.yaml
```diff
@@ -19,13 +19,18 @@ spec:
...
       "--cluster-key=/etc/srv/kubernetes/pki/apiserver.key",
       "--server-port=0",
       "--agent-port=8091",
       "--admin-port=8092",
-      "--mode=http-connect"
+      "--mode=http-connect",
+      "--authenticate-agent=true",
+      "--agent-namespace=kube-system",
+      "--agent-service-account=konnectivity-agent-sa",
+      "--kubeconfig=/etc/srv/kubernetes/konnectivity-server/kubeconfig",
```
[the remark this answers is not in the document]
dberkov: It will be generated by kubernetes host service (aka: GKE in GCP)

examples/kubernetes/kubeconfig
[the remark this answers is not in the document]
dberkov: This config will be used by konnectivity-server static pod for hitting local kube-apiserver. Without specifying  insecure-skip-tls-verify. tls will fail, since we need to pass ca.crt, but even by passing ca.crt tls fail, since ca.crt (at least in GKE) has cluster's IPs only (local host is not valid). Moreover, this is a pattern used by all static pods running in master.

pkg/agent/agentclient/stream.go
dberkov: I removed validation from inside initializeAuthContext method, so will keep it here

proto/agent/agent.proto
dberkov: At early stages of feature development, I passed 2 values instead of one, but found a way how to pass single token only, so yes, I will remove it, since it no more required

examples/kubernetes/README.md
dberkov: Yes, this pattern used across all other static pods. Ex: https://github.com/kubernetes/kubernetes/blob/c14106ad1234742da80eb8f12ddcbf19dba61284/cluster/gce/gci/configure-helper.sh#L613-L615

proto/agent/agent.pb.go
dberkov: Nope

cmd/agent/main.go
```diff
@@ -61,6 +61,9 @@ type GrpcProxyAgentOptions struct {
 	// Configuration for connecting to the proxy-server
 	proxyServerHost string
 	proxyServerPort int
+
+	// file contains service account authorization token for enabling proxy-server token based authorization
+	serviceAccountToken string
```
caesarxuchao: Can you name it serviceAccountTokenPath, in consistent with kubeconfigPath?

cmd/agent/main.go
```diff
@@ -70,6 +73,7 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.caCert, "ca-cert", o.caCert, "If non-empty the CAs we use to validate clients.")
 	flags.StringVar(&o.proxyServerHost, "proxy-server-host", o.proxyServerHost, "The hostname to use to connect to the proxy-server.")
 	flags.IntVar(&o.proxyServerPort, "proxy-server-port", o.proxyServerPort, "The port the proxy server is listening on.")
+	flags.StringVar(&o.serviceAccountToken, "service-account-token", o.serviceAccountToken, "If non-empty authorize each call to the proxy-server with token file's content.")
```
caesarxuchao: nit: how about "If non-empty proxy agent uses this token to prove its identity to the proxy server"?

cmd/agent/main.go
```diff
@@ -79,6 +83,7 @@ func (o *GrpcProxyAgentOptions) Print() {
 	klog.Warningf("CACert set to \"%s\".\n", o.caCert)
 	klog.Warningf("ProxyServerHost set to \"%s\".\n", o.proxyServerHost)
 	klog.Warningf("ProxyServerPort set to %d.\n", o.proxyServerPort)
+	klog.Warningf("Authorization Token set to \"%s\".\n", o.serviceAccountToken)
```
caesarxuchao: update "Authorization Token"

cmd/proxy/main.go
```diff
@@ -94,6 +104,11 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.UintVar(&o.serverPort, "server-port", o.serverPort, "Port we listen for server connections on. Set to 0 for UDS.")
 	flags.UintVar(&o.agentPort, "agent-port", o.agentPort, "Port we listen for agent connections on.")
 	flags.UintVar(&o.adminPort, "admin-port", o.adminPort, "Port we listen for admin connections on.")
+	flags.BoolVar(&o.authenticateAgent, "authenticate-agent", o.authenticateAgent, "True if agent needs to be authenticated.")
```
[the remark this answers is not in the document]
caesarxuchao: Also, the naming of the flag is not accurate, because if clusterCaCert is specified, proxy server will authenticate the agent as well.

examples/kubernetes/konnectivity-agent.yaml
```diff
@@ -33,3 +40,16 @@ spec:
...
+  serviceAccountName: konnectivity-agent-sa
+  volumes:
+  - name: konnectivity-agent-token
+    projected:
+      sources:
+      - serviceAccountToken:
+          path: konnectivity-agent-token
+          expirationSeconds: 7200
+          audience: konnectivity
+
```
caesarxuchao: nit: Please remove the empty line.

pkg/agent/agentclient/client.go
```diff
@@ -37,8 +37,8 @@ type AgentClient struct {
 }
 
 // NewAgentClient creates an AgentClient
-func NewAgentClient(address string, opts ...grpc.DialOption) (*AgentClient, error) {
-	stream, err := NewRedialableAgentClient(address, opts...)
+func NewAgentClient(address, saTokenFile string, opts ...grpc.DialOption) (*AgentClient, error) {
```
caesarxuchao: nit:saTokenPath to be consisent with kubeconfigPath.

pkg/util/auth.go
caesarxuchao: Also, can you move it to proto/header/header.go, like in this [PR](https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/46/files#diff-6989bdcc4293a974aef57ca42635665d)? This essentially is part of our api, not a util.

cmd/proxy/main.go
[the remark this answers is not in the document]
dberkov: flag has been removed

examples/kubernetes/README.md
```diff
@@ -30,6 +31,30 @@ CLUSTER_CERT=/etc/srv/kubernetes/pki/apiserver.crt
...
+Append following line to the [static-token-file](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#static-token-file) and restart **kube-apiserver** on the master
+```bash
+echo "${SERVER_TOKEN},system:konnectivity-server,uid:system:konnectivity-server"
+```
+
+#### GKE specific configuration
+1. [static-token-file](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#static-token-file) location is: **/etc/srv/kubernetes/known_tokens.csv**
+
+1. Restart kube-api
```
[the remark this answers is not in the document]
dberkov: fixed

pkg/agent/agentclient/stream.go
```diff
@@ -48,20 +52,24 @@ type RedialableAgentClient struct {
 	recvLock   sync.Mutex
 	reconnLock sync.Mutex
 
+	// file name contains service account token
```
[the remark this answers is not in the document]
dberkov: renamed

pkg/util/auth.go
dberkov: 1. moved to proto/header/header.go 2. Aligned with https://tools.ietf.org/html/rfc6750#section-2.1 standard

cmd/agent/main.go
dberkov: renamed

cmd/agent/main.go
dberkov: string has been replaced as suggested.

cmd/agent/main.go
dberkov: done

examples/kubernetes/konnectivity-agent.yaml
dberkov: removed

pkg/agent/agentclient/client.go
dberkov: renamed

cmd/proxy/main.go
```diff
@@ -79,6 +82,12 @@ type ProxyRunOptions struct {
 	agentPort uint
 	// Port we listen for admin connections on.
 	adminPort uint
+
+	// Agent authentication
```
caesarxuchao: nit: Token-based agent authentication

cmd/proxy/main.go
```diff
@@ -94,6 +103,10 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.UintVar(&o.serverPort, "server-port", o.serverPort, "Port we listen for server connections on. Set to 0 for UDS.")
 	flags.UintVar(&o.agentPort, "agent-port", o.agentPort, "Port we listen for agent connections on.")
 	flags.UintVar(&o.adminPort, "admin-port", o.adminPort, "Port we listen for admin connections on.")
+	flags.StringVar(&o.agentNamespace, "agent-namespace", o.agentNamespace, "Expected agent's namespace during agent authentication (used with agent-service-account, authentication-audience, kubeconfig).")
+	flags.StringVar(&o.agentServiceAccount, "agent-service-account", o.agentServiceAccount, "Expected agent's service account during agent authentication (used with agent-namespace, authentication-audience, kubeconfig).")
+	flags.StringVar(&o.kubeconfigPath, "kubeconfig", o.kubeconfigPath, "absolute path to the kubeconfig file specifying the apiserver instance (used with agent-namespace, agent-service-account, authentication-audience).")
```
caesarxuchao: nit: s/apiserver instance/kube apiserver which will review the service account token

pkg/agent/agentserver/server.go
```diff
@@ -61,17 +67,30 @@ type ProxyServer struct {
 	// connID track
 	Frontends   map[int64]*ProxyClientConnection
 	PendingDial map[int64]*ProxyClientConnection
+
+	// agent authentication
+	AgentAuthenticationOptions *AgentAuthenticationOptions
+}
+
+// AgentAuthenticationOptions contains list of parameters required for agent authentication
```
caesarxuchao: Can you rename it to AgentTokenAuthenticationOptions and update the comment, since we still support the cert based authentication?

Jefftree:
@dberkov @caesarxuchao: Since network proxy promises MTLS, can we enforce that either a cert or token must be sent by the proxy agent? If no token is sent, the konnectivity-server will reject the request and unregister the connection, but the konnectivity-agent pod will be in a infinite crash loop state, and keep trying to connect to the server with no token info supplied.

examples/kubernetes/konnectivity-agent.yaml
```diff
@@ -20,6 +26,7 @@ spec:
       "--ca-cert=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
       "--proxy-server-host=${CLUSTER_IP}",
       "--proxy-server-port=8091",
+      "--service-account-token-path=/var/run/secrets/tokens/konnectivity-agent-token",
```
Jefftree: afaik a pod can only be bound to one service account, and the service account cannot be changed after pod creation. I noticed that the token info is mounted to `var/run/secrets/kubernetes.io/serviceaccount/token` by default. I think we can directly reference that token instead of having to mount the volumes below?

cc: @caesarxuchao

pkg/agent/agentserver/server.go
```diff
@@ -196,6 +215,68 @@ func (s *ProxyServer) serveSend(stream agent.ProxyService_ProxyServer, sendCh <-
...
+	}
+
+	if s.AgentAuthenticationOptions.AgentServiceAccount != sa {
+		return fmt.Errorf("lookup failed: incoming request from %q service account. Expected %q", sa, s.AgentAuthenticationOptions.AgentServiceAccount)
+	}
+
+	return nil
+}
+
+func (s *ProxyServer) authenticateAgent(ctx context.Context) error {
```
Jefftree: Can you rename this to authenticateAgentViaToken or something similar?

proto/header/header.go
```diff
@@ -0,0 +1,27 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package header
+
+const (
+	// AuthenticationContextKey will be used as a key to store authentication tokens in grpc call
+	// (https://tools.ietf.org/html/rfc6750#section-2.1)
+	AuthenticationContextKey = "Authorization"
```
Jefftree: maybe `AuthenticationTokenContextKey`?

pkg/agent/agentserver/server.go
```diff
@@ -196,6 +215,68 @@ func (s *ProxyServer) serveSend(stream agent.ProxyService_ProxyServer, sendCh <-
...
+	authContext := md.Get(header.AuthenticationContextKey)
+	if len(authContext) == 0 {
+		return fmt.Errorf("Authentication context was not found in metadata")
+	}
+
+	if err := s.validateAuthToken(strings.TrimPrefix(authContext[0], header.AuthenticationContextSchemePrefix)); err != nil {
+		return fmt.Errorf("Failed to validate authentication token, err:%v", err)
+	}
+
+	klog.Infof("Client successfully authenticated")
```
Jefftree: authenticated via token

pkg/agent/agentserver/server.go
```diff
@@ -196,6 +215,68 @@ func (s *ProxyServer) serveSend(stream agent.ProxyService_ProxyServer, sendCh <-
 	}
 }
 
+func (s *ProxyServer) validateAuthToken(token string) error {
+	trReq := &authv1.TokenReview{
+		Spec: authv1.TokenReviewSpec{
+			Token:     token,
+			Audiences: []string{s.AgentAuthenticationOptions.AuthenticationAudience},
```
Jefftree: You mentioned that Audiences is mandatory, is it something standardized across other system pods? I noticed that it's optional in the api (https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/authentication/v1/types.go#L73) and didn't see any other instances of it being used in k/k.

proto/header/header.go
dberkov: Authorization - this is a standard by https://tools.ietf.org/html/rfc6750#section-2.1

examples/kubernetes/konnectivity-agent.yaml
dberkov: cc: @mikedanese

pkg/agent/agentserver/server.go
dberkov: cc @mikedanese

proto/header/header.go
Jefftree: (Referring to the name, eg: `AuthenticationTokenContextKey = "Authorization"`)

cmd/agent/main.go
```diff
@@ -106,16 +111,22 @@ func (o *GrpcProxyAgentOptions) Validate() error {
 	if o.proxyServerPort <= 0 {
 		return fmt.Errorf("proxy server port %d must be greater than 0", o.proxyServerPort)
 	}
+	if o.serviceAccountTokenPath != "" {
+		if _, err := os.Stat(o.serviceAccountTokenPath); os.IsNotExist(err) {
+			return fmt.Errorf("error checking service account authorization token %s, got %v", o.serviceAccountTokenPath, err)
```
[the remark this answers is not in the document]
dberkov: done

cmd/proxy/main.go
dberkov: done

cmd/proxy/main.go
dberkov: done

pkg/agent/agentserver/server.go
dberkov: done

pkg/agent/agentserver/server.go
dberkov: done

proto/header/header.go
dberkov: done

pkg/agent/agentserver/server.go
dberkov: done

dberkov:
> A few more nits.
> 
> @Jefftree @dberkov have you manually tested it in GCE/GKE?
> 
> @dberkov I understand that it's difficult to write a complete test because we don't have the test framework that runs a k8s cluster. Can you add an integration test to tests/ to verify that the proxy-server denies the Connect request if the agent doesn't send a bearer token at all? That doesn't require a k8s cluster running.

I have created pkg/agent/agentserver/server_test.go with full test coverage on new behavior on server side

pkg/agent/agentserver/server.go
mikedanese: Since bearer tokens can be replayed by the relying party, security is predicated on the assumption that a single bearer token is only valid against the relying party that they it was originally intended for.

> To deal with token redirect, it is important for the authorization
> server to include the identity of the intended recipients (the
> audience), typically a single resource server (or a list of resource
> servers), in the token.

https://tools.ietf.org/html/rfc6750#section-5.2

All uses of tokens in k/k are to communicate with the kube-apiserver.  The audience in a TokenRequest and TokenReview defaults to the kube-apiserver. Since the konnectivity server is not the kube-apiserver, it should not accept tokens that are audience bound the kube-apiserver (just like the kube-apiserver should not accept tokens that are audience bound to the konnectivity server).

Jefftree: That makes sense. Thanks!

Makefile
```diff
@@ -75,6 +75,12 @@ proto/agent/agent.pb.go: proto/agent/agent.proto
 	cat hack/go-license-header.txt proto/agent/agent.pb.go > proto/agent/agent.licensed.go
 	mv proto/agent/agent.licensed.go proto/agent/agent.pb.go
 
+	mkdir -p mocks
```
Jefftree: We should put these commands as part of `test:` rather than here.

mocks/agent_mock.go
```diff
@@ -0,0 +1,162 @@
+/*
```
Jefftree: This file doesn't need to be committed since we're already generating it in the Makefile?

dberkov: removed from source control

Makefile
dberkov: done

cmd/agent/main.go
```diff
@@ -106,16 +111,26 @@ func (o *GrpcProxyAgentOptions) Validate() error {
...
 		return fmt.Errorf("proxy server port %d must be greater than 0", o.proxyServerPort)
 	}
+	if o.serviceAccountTokenPath != "" {
+		if _, err := os.Stat(o.serviceAccountTokenPath); os.IsNotExist(err) {
+			return fmt.Errorf("error checking service account token path %s, got %v", o.serviceAccountTokenPath, err)
+		}
+	}
+
+	if o.agentCert == "" && o.agentKey == "" && o.serviceAccountTokenPath == "" {
+		return fmt.Errorf("agent must enable certificate based or token based authentication")
```
cheftako: I believe this should be enforced from the server side and not the client side. There are legitimate non production use cases for turning off client authentication.

cheftako: In addition if we allow this through I think we don't need the mock agent as the real agent can do what is necessary.

cheftako: We may want to add an enum setting to the proxy/main.go at this point for requiredAgentAuth? That way we can detect if the agent is meeting the minimum authentication requirements.

cheftako:
/assign @mikedanese 
I would like to make sure Mike takes a look and approves prior to merge.

cmd/agent/main.go
Jefftree: I brought this up because an incorrectly configured konnectivity-agent would send a large number of denied requests to the konnectivity-server. (agent retries + multiple threads hitting LB for regional clusters + CrashLoop retries).

Since there are use cases for turning off client authentication, removing this check is fine, but we should still think of ways to limit the outgoing requests of an incorrectly configured agent.

dberkov: I removed this validation

pkg/agent/agentclient/stream.go
```diff
@@ -332,6 +337,20 @@ type connectResult struct {
 	agentServiceClient agent.AgentService_ConnectClient
 }
 
+func (c *RedialableAgentClient) initializeAuthContext(ctx context.Context) (context.Context, error) {
+	var err error
+	var b []byte
+
+	// load service account's token
+	if b, err = ioutil.ReadFile(c.serviceAccountTokenPath); err != nil {
```
caesarxuchao: I think this token won't change in the life time of the agent, then can we read the token when constructing the clientset, and pass down to the rediableAgentClient, instead of read the file every time retry connection?
I'm ok if you want to do this in a followup PR.

pkg/agent/agentserver/server_test.go
```diff
@@ -67,3 +81,147 @@ func TestAddRemoveBackends(t *testing.T) {
...
+			wantNamespace:      ns,
+			wantServiceAccount: sa,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			kcs := k8sfake.NewSimpleClientset()
+
+			kcs.AuthenticationV1().(*fakeauthenticationv1.FakeAuthenticationV1).Fake.PrependReactor("*", "*", func(action k8stesting.Action) (handled bool, ret runtime.Object, err error) {
```
caesarxuchao: For the first two parameters of PrependReactor, can you refine the scope? I guess they should be "create" and "tokenReviews".

mocks/agent_mock.go
caesarxuchao: The generated proto go file is commited as well, let's commit this one as well.

Two suggestions:
1. we need a verify srcipt to make sure the generated code is up-to-date, similar to this [file](https://github.com/kubernetes/kubernetes/blob/master/hack/verify-codegen.sh) in k/k. We don't need to do it in this PR, but if you agree, let's create an issue to track it.
2. let's move this package next to the real code, i.e., proto/agent/mock/agent_mock.go.

pkg/agent/agentclient/stream.go
dberkov: It changed every 1 hour (https://kubernetes.io/docs/concepts/storage/volumes/#projected)

cmd/proxy/main.go
```diff
@@ -202,24 +219,48 @@ func (o *ProxyRunOptions) Validate() error {
...
+	if o.agentNamespace != "" || o.agentServiceAccount != "" || o.authenticationAudience != "" || o.kubeconfigPath != "" {
+		if o.agentNamespace == "" {
+			return fmt.Errorf("agentNamespace cannot be empty when agent authentication is enabled")
+		}
+		if o.agentServiceAccount == "" {
+			return fmt.Errorf("agentServiceAccount cannot be empty when agent authentication is enabled")
+		}
+		if o.authenticationAudience == "" {
+			return fmt.Errorf("authenticationAudience cannot be empty when agent authentication is enabled")
+		}
```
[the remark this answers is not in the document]
dberkov: It cannot be nil, since we have newProxyRunOptions().

examples/kubernetes/konnectivity-agent.yaml
```diff
@@ -33,3 +40,14 @@ spec:
...
+      - mountPath: /var/run/secrets/tokens
+        name: konnectivity-agent-token
+  serviceAccountName: konnectivity-agent
+  volumes:
+  - name: konnectivity-agent-token
+    projected:
+      sources:
+      - serviceAccountToken:
+          path: konnectivity-agent-token
+          audience: system:konnectivity-server
```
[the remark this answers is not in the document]
dberkov: KubernetesClient.AuthenticationV1().TokenReviews() gets the audience is the parameters, so k8s API validates that token is issued with this audience

pkg/agent/agentserver/server_test.go
dberkov: done

examples/kubernetes/README.md
```diff
@@ -30,6 +31,31 @@ CLUSTER_CERT=/etc/srv/kubernetes/pki/apiserver.crt
 CLUSTER_KEY=/etc/srv/kubernetes/pki/apiserver.key
 ```
 
+# Register SERVER_TOKEN in [static-token-file](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#static-token-file)
+Append following line to the [static-token-file](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#static-token-file) and restart **kube-apiserver** on the master
```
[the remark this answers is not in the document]
dberkov: done

Makefile
```diff
@@ -34,9 +34,14 @@ DOCKER_CLI_EXPERIMENTAL ?= enabled
...
 ## --------------------------------------
+mock_gen:
+	mkdir -p mocks
+	mockgen sigs.k8s.io/apiserver-network-proxy/proto/agent AgentService_ConnectServer > mocks/agent_mock.go
+	cat hack/go-license-header.txt mocks/agent_mock.go > mocks/agent_mock.licensed.go
+	mv mocks/agent_mock.licensed.go mocks/agent_mock.go
 
 .PHONY: test
-test:
+test:
```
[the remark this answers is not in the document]
dberkov: done

examples/kubernetes/konnectivity-agent.yaml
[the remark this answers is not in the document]
caesarxuchao: But konnectivity-server calls TokenReviews, and this is the yaml for the konnectivity-agent.

My guess is the token data mounted by the agent will contain the "audience", and apiserver will be able to extract the "audience" out from the token.

pkg/agent/agentclient/stream.go
caesarxuchao: Good to know. Can you add a comment to document this fact?

examples/kubernetes/konnectivity-server.yaml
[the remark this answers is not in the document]
Jefftree: Did some testing on GCE and the kubeconfig file will be generated to `/mnt/disks/master-pd/srv/kubernetes/konnectivity-server/kubeconfig`. Technically this doesn't need to be renamed, but I'll leave the decision to you.

mocks/agent_mock.go
dberkov: Moved to proto/agent/mock/agent_mock.go

pkg/agent/agentclient/stream.go
dberkov: done

cmd/proxy/main.go
```diff
@@ -86,6 +87,14 @@ type ProxyRunOptions struct {
 	serverID string
 	// Number of proxy server instances, should be 1 unless it is a HA proxy server.
 	serverCount uint
+	// Agent pod's namespace for token-based agent authentication
+	agentNamespace string
+	// Agent pod's service account for token-based agent authentication
```
cheftako: Will we ever want different service accounts for different agents? Eg. agent service account per failure domain?

cmd/proxy/main.go
```diff
@@ -202,24 +219,48 @@ func (o *ProxyRunOptions) Validate() error {
 	if o.adminPort < 1024 {
 		return fmt.Errorf("please do not try to use reserved port %d for the admin port", o.adminPort)
 	}
+
+	// validate agent authentication params
+	// all 4 parametes must be empty or must have value (except kubeconfigPath that might be empty)
+	if o.agentNamespace != "" || o.agentServiceAccount != "" || o.authenticationAudience != "" || o.kubeconfigPath != "" {
+		if o.agentNamespace == "" {
+			return fmt.Errorf("agentNamespace cannot be empty when agent authentication is enabled")
```
cheftako: For future we should consider accumulating these errors. Sort of annoying to be given and error I need a service account on run1 and then be given an error I need an audience on run2.

examples/kubernetes/README.md
```diff
@@ -30,6 +31,31 @@ CLUSTER_CERT=/etc/srv/kubernetes/pki/apiserver.crt
 CLUSTER_KEY=/etc/srv/kubernetes/pki/apiserver.key
 ```
 
+# Register SERVER_TOKEN in [static-token-file](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#static-token-file)
+Append the output of the following line to the [static-token-file](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#static-token-file) and restart **kube-apiserver** on the master
+```bash
+echo "${SERVER_TOKEN},system:konnectivity-server,uid:system:konnectivity-server"
+```
+
+#### GKE specific configuration
```
cheftako: GCE? Should not have anything GKE specific here.

PR #52, opened 2020-01-27
minor fixes

caesarxuchao:
Addressing some remaining comments in #46 made by @cheftako 

/assign @Jefftree @cheftako

PR #53, opened 2020-01-27
Do exponential backoff in clientset sync

caesarxuchao:
... to avoid hot loops.

First commit is #52.

### DIFF
--- a/pkg/agent/agentclient/clientset.go
+++ b/pkg/agent/agentclient/clientset.go
@@ -18,12 +18,12 @@ package agentclient
 
 import (
 	"fmt"
-	"math/rand"
 	"sync"
 	"time"
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/connectivity"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog"
 )
 
@@ -31,7 +31,7 @@ import (
 type ClientSet struct {
 	mu      sync.Mutex              //protects the clients.
 	clients map[string]*AgentClient // map between serverID and the client
-	// connects to this proxy server.
+	// connects to this server.
 
 	agentID     string // ID of this agent
 	address     string // proxy server address. Assuming HA proxy server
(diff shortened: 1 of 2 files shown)


pkg/agent/agentclient/clientset.go
```diff
@@ -124,27 +124,37 @@ func (cs *ClientSet) newAgentClient() (*AgentClient, error) {
...
 		if cs.serverCount != 0 {
-			sleep := cs.syncInterval + time.Duration(rand.Float64()*jitter*float64(cs.syncInterval))
-			time.Sleep(sleep)
+			time.Sleep(backoff.Duration)
 		}
 		if cs.serverCount == 0 || cs.ClientsCount() < cs.serverCount {
 			c, err := cs.newAgentClient()
 			if err != nil {
 				klog.Error(err)
+				backoff.Step()
```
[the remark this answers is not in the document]
caesarxuchao: Good catch. I accidentally deleted the "Factor" line. Added back.

caesarxuchao:
>  Should this scenario warrant an exponential backoff?

No. We should only exponential backoff when we fail to connect to any server. Updated. Thanks.

pkg/agent/agentclient/clientset.go
```diff
@@ -124,20 +124,30 @@ func (cs *ClientSet) newAgentClient() (*AgentClient, error) {
 	return newAgentClient(cs.address, cs.agentID, cs, cs.dialOption)
 }
 
+func (cs *ClientSet) resetBackoff() *wait.Backoff {
+	return &wait.Backoff{
+		Steps:    3,
+		Jitter:   0.1,
```
Jefftree: It still looks like this parameter is not used, should we remove it?

caesarxuchao:
@Jefftree I refactored the code. PTAL. Thank you.

pkg/agent/agentclient/clientset.go
```diff
@@ -128,30 +128,48 @@ func (cs *ClientSet) newAgentClient() (*AgentClient, error) {
...
+		Factor:   1.5,
+		Duration: cs.syncInterval,
+	}
```
dberkov: Do you want initialize "Cap" too, just in  case it grows too much?

pkg/agent/agentclient/clientset.go
```diff
@@ -128,30 +128,48 @@ func (cs *ClientSet) newAgentClient() (*AgentClient, error) {
...
-				continue
-			}
-			klog.Infof("sync added client connecting to proxy server %s", c.stream.serverID)
-			go c.Serve()
+		if err := cs.syncOnce(); err != nil {
+			klog.Error(err)
+			duration = backoff.Step()
+		} else {
+			backoff = cs.resetBackoff()
+			duration = wait.Jitter(backoff.Duration, backoff.Jitter)
```
dberkov: based on Jitter documentation, we can set second parameter as 0.0, so default value will be selected

pkg/agent/agentclient/clientset.go
Jefftree: +1

caesarxuchao: Done

pkg/agent/agentclient/clientset.go
caesarxuchao: I think using the value defined in this file is more readable than using the default value.

caesarxuchao:
PTAL. Thank you for the reviews!

PR #58, opened 2020-02-05
Let the proxy server handle concurrent client requests correctly

caesarxuchao:
Terms:
* connection: the persistent grpc connection between the proxy server and a proxy agent.
* stream: a dial request marks the begin of a stream, and a close request marks the end of a stream. There are multiple streams multiplexed in one connection. 
* client: the entity that sends the dial request to the proxy server. For example, kube-apiserver is a client.

Bug fixed by this PR:
* The proxy agent uses connID, an int64 counter, to index the streams it handles
* The proxy server also uses connID. However, different agents can have the same connID, causing incorrect routing among streams.

Contents of this PR:
* The first commit adds a unit test to demonstrate the bug
* The second commit fixes the proxy server to use agentID + connID to uniquely index a stream.

Alternatives:
* we can solve the problem from the agent side as well by making the connID unique across agents. I don't think it makes much difference.

/assign @Jefftree @cheftako @dberkov

### DIFF
--- a/pkg/agent/agentserver/backend_manager.go
+++ b/pkg/agent/agentserver/backend_manager.go
@@ -1,0 +1,139 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package agentserver
+
+import (
+	"math/rand"
+	"sync"
+	"time"
+
+	"k8s.io/klog"
+	"sigs.k8s.io/apiserver-network-proxy/proto/agent"
+)
+
+// BackendManager is an interface to manage backend connections, i.e.,
+// connection to the proxy agents.
+type BackendManager interface {
+	// Backend returns a single backend.
+	Backend() (agent.AgentService_ConnectServer, error)
+	// AddBackend adds a backend.
+	AddBackend(agentID string, conn agent.AgentService_ConnectServer)
+	// RemoveBackend removes a backend.
+	RemoveBackend(agentID string, conn agent.AgentService_ConnectServer)
+}
+
+var _ BackendManager = &DefaultBackendManager{}
+
+// DefaultBackendManager is the default backend manager.
+type DefaultBackendManager struct {
+	mu sync.RWMutex //protects the following
+	// A map between agentID and its grpc connections.
+	// For a given agent, ProxyServer prefers backends[agentID][0] to send
+	// traffic, because backends[agentID][1:] are more likely to be closed
+	// by the agent to deduplicate connections to the same server.
+	backends map[string][]agent.AgentService_ConnectServer
+	// agentID is tracked in this slice to enable randomly picking an
+	// agentID in the Backend() method. There is no reliable way to
+	// randomly pick a key from a map (in this case, the backends) in
+	// Golang.
+	agentIDs []string
+	random   *rand.Rand
+}
+
+// NewDefaultBackendManager returns a DefaultBackendManager.
+func NewDefaultBackendManager() *DefaultBackendManager {
+	return &DefaultBackendManager{
+		backends: make(map[string][]agent.AgentService_ConnectServer),
+		random:   rand.New(rand.NewSource(time.Now().UnixNano())),
+	}
+}
+
+// AddBackend adds a backend.
+func (s *DefaultBackendManager) AddBackend(agentID string, conn agent.AgentService_ConnectServer) {
+	klog.Infof("register Backend %v for agentID %s", conn, agentID)
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	_, ok := s.backends[agentID]
+	if ok {
+		for _, v := range s.backends[agentID] {
+			if v == conn {
+				klog.Warningf("this should not happen. Adding existing connection %v for agentID %s", conn, agentID)
+				return
+			}
+		}
+		s.backends[agentID] = append(s.backends[agentID], conn)
+		return
+	}
+	s.backends[agentID] = []agent.AgentService_ConnectServer{conn}
+	s.agentIDs = append(s.agentIDs, agentID)
+}
+
+// RemoveBackend removes a backend.
+func (s *DefaultBackendManager) RemoveBackend(agentID string, conn agent.AgentService_ConnectServer) {
+	klog.Infof("remove Backend %v for agentID %s", conn, agentID)
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	backends, ok := s.backends[agentID]
+	if !ok {
+		klog.Warningf("can't find agentID %s in the backends", agentID)
+		return
+	}
+	var found bool
+	for i, c := range backends {
+		if c == conn {
+			s.backends[agentID] = append(s.backends[agentID][:i], s.backends[agentID][i+1:]...)
+			if i == 0 && len(s.backends) != 0 {
+				klog.Warningf("this should not happen. Removed connection %v that is not the first connection, remaining connections are %v", conn, s.backends[agentID])
+			}
+			found = true
+		}
+	}
+	if len(s.backends[agentID]) == 0 {
+		delete(s.backends, agentID)
+		for i := range s.agentIDs {
+			if s.agentIDs[i] == agentID {
+				s.agentIDs[i] = s.agentIDs[len(s.agentIDs)-1]
+				s.agentIDs = s.agentIDs[:len(s.agentIDs)-1]
+				break
+			}
+		}
+	}
+	if !found {
+		klog.Errorf("can't find conn %v for agentID %s in the backends", conn, agentID)
+	}
+}
+
+// ErrNotFound indicates that no backend can be found.
+type ErrNotFound struct{}
+
+// Error returns the error message.
+func (e *ErrNotFound) Error() string {
+	return "No backend available"
+}
+
+// Backend returns a random backend.
+func (s *DefaultBackendManager) Backend() (agent.AgentService_ConnectServer, error) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	if len(s.backends) == 0 {
+		return nil, &ErrNotFound{}
+	}
+	agentID := s.agentIDs[s.random.Intn(len(s.agentIDs))]
+	// always return the first connection to an agent, because the agent
+	// will close later connections if there are multiple.
+	return s.backends[agentID][0], nil
+}
--- a/pkg/agent/agentserver/server.go
+++ b/pkg/agent/agentserver/server.go
@@ -20,12 +20,10 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"math/rand"
 	"net"
 	"strconv"
 	"strings"
 	"sync"
-	"time"
 
 	"google.golang.org/grpc/metadata"
 	authv1 "k8s.io/api/authentication/v1"
@@ -67,17 +65,14 @@ func (c *ProxyClientConnection) send(pkt *client.Packet) error {
 
 // ProxyServer
 type ProxyServer struct {
-	mu sync.RWMutex //protects the following
-	// A map between agentID and its grpc connections.
-	// For a given agent, ProxyServer prefers backends[agentID][0] to send
-	// traffic, because backends[agentID][1:] are more likely to be closed
-	// by the agent to deduplicate connections to the same server.
-	backends map[string][]agent.AgentService_ConnectServer
-	agentIDs []string
-	random   *rand.Rand
+	// BackendManager manages the backends.
+	BackendManager BackendManager
 
-	// connID track
-	Frontends   map[int64]*ProxyClientConnection
+	// fmu protects frontends.
+	fmu sync.RWMutex
+	// conn = Frontend[agentID][connID]
+	frontends map[string]map[int64]*ProxyClientConnection
+
 	PendingDial map[int64]*ProxyClientConnection
 
 	serverID    string // unique ID of this server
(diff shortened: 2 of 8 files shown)


pkg/agent/agentserver/server.go
```diff
@@ -70,9 +70,14 @@ type ProxyServer struct {
 	backends map[string][]agent.AgentService_ConnectServer
 	agentIDs []string
 	random   *rand.Rand
+	// This interface makes testing easier.
+	BackendGetter BackendGetter
```
Jefftree: I actually really like the way you do this. Perhaps this can be used not only for testing, but also when we want look at new strategies for load balancing. I'm wondering if it makes sense to add this interface as an optional parameter to the `agentserver.NewProxyServer` function.

pkg/agent/agentserver/server.go
```diff
@@ -201,7 +273,7 @@ func (s *ProxyServer) serveRecvFrontend(stream agent.ProxyService_ProxyServer, r
 			// the address, then we can send the Dial_REQ to the
 			// same agent. That way we save the agent from creating
 			// a new connection to the address.
-			backend, err = s.randomBackend()
+			backend, err = s.BackendGetter.Backend()
```
Jefftree: Maybe `getSingleBackend`? Just something to indicate that we're picking one backend out of multiple.

tests/proxy_test.go
```diff
@@ -218,10 +218,11 @@ type proxy struct {
 }
 
 func runGRPCProxyServer() (proxy, func(), error) {
-	return runGRPCProxyServerWithServerCount(1)
+	p, _, cleanup, err := runGRPCProxyServerWithServerCount(1)
+	return p, cleanup, err
```
Jefftree: Can we make the parameters returned consistent even though it'll probably be another _ assignment?

pkg/agent/agentserver/server.go
dberkov: +1

pkg/agent/agentserver/server.go
```diff
@@ -84,6 +89,48 @@ var _ agent.AgentServiceServer = &ProxyServer{}
 
 var _ agent.ProxyServiceServer = &ProxyServer{}
 
+func (s *ProxyServer) addFrontend(agentID string, connID int64, p *ProxyClientConnection) {
+	s.fmu.Lock()
```
Jefftree: Similar to add/removeBackend, can we add a log line when we add and remove frontends?

pkg/agent/agentserver/server.go
caesarxuchao: We can add the parameter when it's necessary.

pkg/agent/agentserver/server.go
caesarxuchao: I made the BackendManager interface, which has `Backend()` and `AllBackends()` methods, let me know if their meaning is still not obvious.

tests/proxy_test.go
caesarxuchao: I was trying to reduce the ripple caused by the signature change. Otherwise we would simply use runGRPCProxyServerWithServerCount every where :)

pkg/agent/agentserver/server.go
Jefftree: That looks good. Thanks!

Jefftree:
lgtm from me. Deferring to @cheftako for the approval.

pkg/agent/agentserver/backend_manager.go
```diff
@@ -0,0 +1,130 @@
...
+	backends map[string][]agent.AgentService_ConnectServer
+	agentIDs []string
+	random   *rand.Rand
+}
+
+// NewDefaultBackendManager returns a DefaultBackendManager.
+func NewDefaultBackendManager() *DefaultBackendManager {
+	return &DefaultBackendManager{
+		backends: make(map[string][]agent.AgentService_ConnectServer),
+		random:   rand.New(rand.NewSource(time.Now().UTC().UnixNano())),
```
cheftako: Why bother with UTC() ?

pkg/agent/agentserver/backend_manager.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+}
+
+// AddBackend adds a backend.
+func (s *DefaultBackendManager) AddBackend(agentID string, conn agent.AgentService_ConnectServer) {
+	klog.Infof("register Backend %v for agentID %s", conn, agentID)
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	_, ok := s.backends[agentID]
```
cheftako: Why not 
list, ok := s.backends[agentID]
if ok {
		s.backends[agentID] = append(list, conn)

caesarxuchao: This is a style thing. I stick to the form of "a=append(a, ...)" style to keep the code away from the [golang append to a slice gotcha](https://allegro.tech/2017/07/golang-slices-gotcha.html). In k/k we once did a sweep in the code base to enforce this style.

pkg/agent/agentserver/backend_manager.go
```diff
@@ -0,0 +1,130 @@
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	backends, ok := s.backends[agentID]
+	if !ok {
+		klog.Warningf("can't find agentID %s in the backends", agentID)
+		return
+	}
+	var found bool
+	for i, c := range backends {
+		if c == conn {
```
cheftako: Please make sure we have a test for mismatching conn and agentID. We should never be asked for a mismatching pair.

pkg/agent/agentserver/backend_manager.go
```diff
@@ -0,0 +1,130 @@
...
+		for i := range s.agentIDs {
+			if s.agentIDs[i] == agentID {
+				s.agentIDs[i] = s.agentIDs[len(s.agentIDs)-1]
+				s.agentIDs = s.agentIDs[:len(s.agentIDs)-1]
+				break
+			}
+		}
+	}
+	if !found {
+		klog.Warningf("can't find conn %v for agentID %s in the backends", conn, agentID)
```
cheftako: Error maybe?

pkg/agent/agentserver/backend_manager.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	agentID := s.agentIDs[s.random.Intn(len(s.agentIDs))]
+	return s.backends[agentID][0], nil
+}
+
+// AllBackends returns all backends, categorized based on agentID.
+func (s *DefaultBackendManager) AllBackends() map[string][]agent.AgentService_ConnectServer {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	ret := make(map[string][]agent.AgentService_ConnectServer)
```
cheftako: Why does making the map need to be in the lock?

pkg/agent/agentserver/server.go
```diff
@@ -453,38 +435,45 @@ func (s *ProxyServer) serveRecvBackend(stream agent.AgentService_ConnectServer,
...
-				}
+			client, err := s.getFrontend(agentID, resp.ConnectID)
+			if err != nil {
+				klog.Warning(err)
+				break
+			}
+			if err := client.send(pkt); err != nil {
+				klog.Warningf("<<< DATA send to client stream error: %v", err)
+			} else {
+				klog.Infof("<<< DATA sent to frontend")
```
cheftako: We should use something like v=5 for these.

cheftako: Also data sent with indication of agent or client seems useless.

pkg/agent/agentserver/backend_manager.go
cheftako: I believe you need to repeatedly append to the same reference to hit this issue, which is not currently the case. OTOH it would not be hard to add a loop here and have a problem.

pkg/agent/agentserver/backend_manager.go
caesarxuchao: added a invalid removal in the test to verify mismatching conn and agentID will be a no-op.

caesarxuchao:
Updated. PTAL. Thank you!

pkg/agent/agentserver/backend_manager.go
```diff
@@ -0,0 +1,130 @@
...
+	klog.Infof("remove Backend %v for agentID %s", conn, agentID)
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	backends, ok := s.backends[agentID]
+	if !ok {
+		klog.Warningf("can't find agentID %s in the backends", agentID)
+		return
+	}
+	var found bool
+	for i, c := range backends {
```
cheftako: Backend() assumes it will return element 0 of the slice. That means that if len(backends) > 0 and we remove i == 0 we have a problem. We should probably add a warning to that effect in the code.

tests/concurrent_client_request_test.go
```diff
@@ -0,0 +1,141 @@
...
+		Transport: &http.Transport{
+			Dial: tunnel.Dial,
+		},
+		Timeout: 2 * time.Second,
+	}
+}
+
+// singleTimeManager makes sure that a backend only serves one request.
+type singleTimeManager struct {
+	mu sync.Mutex
```
cheftako: Please change name to not match mu in DefaultBackendManager.

pkg/agent/agentserver/backend_manager.go
```diff
@@ -0,0 +1,130 @@
...
+	"k8s.io/klog"
+	"sigs.k8s.io/apiserver-network-proxy/proto/agent"
+)
+
+// BackendManager is an interface to manage backend connections, i.e.,
+// connection to the proxy agents.
+type BackendManager interface {
+	// Backend returns a single backend.
+	Backend() (agent.AgentService_ConnectServer, error)
+	// AllBackends returns all backends, categorized based on agentID.
```
cheftako: Can we doc the method as for test only?

Its returning a slice of AgentService_ConnectServer which my evaporate out from under the caller.

pkg/agent/agentserver/server.go
```diff
@@ -67,17 +65,14 @@ func (c *ProxyClientConnection) send(pkt *client.Packet) error {
...
-	// by the agent to deduplicate connections to the same server.
-	backends map[string][]agent.AgentService_ConnectServer
-	agentIDs []string
-	random   *rand.Rand
+	// BackendManager manages the backends.
+	BackendManager BackendManager
 
 	// connID track
-	Frontends   map[int64]*ProxyClientConnection
+	fmu sync.RWMutex
```
cheftako: Please doc what it is intended to protect.

pkg/agent/agentserver/backend_manager.go
```diff
@@ -0,0 +1,130 @@
...
+
+// DefaultBackendManager is the default backend manager.
+type DefaultBackendManager struct {
+	mu sync.RWMutex //protects the following
+	// A map between agentID and its grpc connections.
+	// For a given agent, ProxyServer prefers backends[agentID][0] to send
+	// traffic, because backends[agentID][1:] are more likely to be closed
+	// by the agent to deduplicate connections to the same server.
+	backends map[string][]agent.AgentService_ConnectServer
+	agentIDs []string
```
dberkov: Could you please add a comment why agentIDs is required. It used inside Backend() method only, so I assume this is for optimization,  to save time on extracting list of keys from map, right?

pkg/agent/agentserver/backend_manager.go
```diff
@@ -0,0 +1,130 @@
+/*
```
dberkov: Could you please add unit tests for this file, to cover cases like adding same connection, removing first, last connection, etc...

pkg/agent/agentserver/backend_manager.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// AddBackend adds a backend.
+func (s *DefaultBackendManager) AddBackend(agentID string, conn agent.AgentService_ConnectServer) {
+	klog.Infof("register Backend %v for agentID %s", conn, agentID)
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	_, ok := s.backends[agentID]
+	if ok {
+		s.backends[agentID] = append(s.backends[agentID], conn)
```
dberkov: It adds conn to the end of the list,  while  Backend() uses a first entry.    backends description says:
 "For a given agent, ProxyServer prefers backends[agentID][0] to send
	// traffic, because backends[agentID][1:] are more likely to be closed
	// by the agent to deduplicate connections to the same server."

Which entry is supposes to be likely used:  the last added or first added?

tests/concurrent_client_request_test.go
caesarxuchao: Removed the embedding altogether.

pkg/agent/agentserver/backend_manager.go
caesarxuchao: Removed.

pkg/agent/agentserver/backend_manager.go
caesarxuchao: Added.

pkg/agent/agentserver/backend_manager.go
caesarxuchao: Added. Otherewise we can't randomly pick a key from the backends map.

pkg/agent/agentserver/backend_manager.go
caesarxuchao: first added connection would be backends[agentID][0], so the two comments are consistent.

pkg/agent/agentserver/backend_manager.go
caesarxuchao: The test was in server_test.go. Moved it to backend_manager_test.go to be clear.

caesarxuchao: Also added test for "adding same connection".

pkg/agent/agentserver/backend_manager.go
```diff
@@ -0,0 +1,132 @@
...
+	backends, ok := s.backends[agentID]
+	if !ok {
+		klog.Warningf("can't find agentID %s in the backends", agentID)
+		return
+	}
+	var found bool
+	for i, c := range backends {
+		if c == conn {
+			s.backends[agentID] = append(s.backends[agentID][:i], s.backends[agentID][i+1:]...)
+			if i != 0 {
```
cheftako: i == 0

pkg/agent/agentserver/backend_manager.go
```diff
@@ -0,0 +1,132 @@
...
+			}
+		}
+	}
+	if !found {
+		klog.Errorf("can't find conn %v for agentID %s in the backends", conn, agentID)
+	}
+}
+
+// Backend returns a random backend.
+func (s *DefaultBackendManager) Backend() (agent.AgentService_ConnectServer, error) {
```
cheftako: maybe ok rather than error?

pkg/agent/agentserver/server.go
```diff
@@ -453,38 +435,45 @@ func (s *ProxyServer) serveRecvBackend(stream agent.AgentService_ConnectServer,
...
-					klog.Infof("<<< CLOSE_RSP sent to frontend")
-				}
+			client, err := s.getFrontend(agentID, resp.ConnectID)
+			if err != nil {
+				klog.Warning(err)
+				break
 			}
+			if err := client.send(pkt); err != nil {
+				// Normal when frontend closes it.
+				klog.Warningf("<<< CLOSE_RSP send to client stream error: %v", err)
```
cheftako: I agree that this would be normal but then we should lower the log level. Info and at least 4.

pkg/agent/agentserver/backend_manager.go
cheftako: ErrNotFound is good

PR #59, opened 2020-02-05
Move client libraries into separate module

Jefftree:
- This is required due to https://github.com/kubernetes/kubernetes/pull/87179#discussion_r374747999
- We will create a separate module for the client and reduce downstream dependency imports for libraries only requiring the client and not the entire network proxy.
- Delete `proto/proxy.proto` since it was not being used.
- I was planning to separate `agent.proto` into two files but realized that we only save [3 lines](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/proto/agent/agent.proto#L25), so I am just moving the entire `agent.proto` into the client directory for now.

Tagging works a bit differently for multi module repos. We will have a tag for the base `apiserver-network-proxy` (Eg: v0.0.4) and a separate tag for for the version of `apiserver-network-proxy/konnectivity-client`. This tag does not need to have the same version as the base tag, but must have a konnectivity-client prefix (Eg: konnectivity-client/v0.0.1). When we do releases, we will most likely have two releases with a tag for both v0.0.x and konnectivity-client/v0.0.y


caesarxuchao:
Thanks for the making the change @Jefftree.

I still think we should move the `Connect` service out of the client module. The headers.go also doesn't belong to the client module. I would move the `Connect` proto and the headers .go to /proto/agent.

I would also rename konnectivity-client/proto/agent -> konnectivity-client/proto/client.

go.sum
```diff
@@ -93,10 +97,15 @@ google.golang.org/grpc v1.19.0/go.mod h1:mqu4LbDTu4XGKhr4mRzUsmM4RtVoemTSY81AxZi
...
 gopkg.in/alecthomas/kingpin.v2 v2.2.6/go.mod h1:FMv+mEhP44yOT+4EoQTLFTRgOQ1FBLkstjWtayDeSgw=
 gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
 gopkg.in/yaml.v2 v2.2.1/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
 honnef.co/go/tools v0.0.0-20190102054323-c2f93a96b099/go.mod h1:rf3lG4BRIbNafJWhAfAdb/ePZxsR/4RtNHQocxwk9r4=
 honnef.co/go/tools v0.0.0-20190523083050-ea95bdfd59fc/go.mod h1:rf3lG4BRIbNafJWhAfAdb/ePZxsR/4RtNHQocxwk9r4=
 k8s.io/klog v0.3.0 h1:0VPpR+sizsiivjIfIAQH/rl8tan6jvWkS7lU+0di3lE=
 k8s.io/klog v0.3.0/go.mod h1:Gq+BEi5rUBO/HRz0bTSXDUcqjScdoY3a9IHpCEIOOfk=
+k8s.io/klog v1.0.0 h1:Pt+yjF5aB1xDSVbau4VsWe+dQNzA0qv1LlXdC2dF6Q8=
+k8s.io/klog v1.0.0/go.mod h1:4Bi6QPql/J/LkTDqv7R/cd3hPo4k2DG6Ptcz060Ez5I=
+sigs.k8s.io/apiserver-network-proxy v0.0.4/go.mod h1:4CYp06cVRHIXzhiPCXP6p54bLOAWrJ0q+WnhS6lPH+Q=
```
Jefftree: We have a chicken and egg problem with this. We can't reference the new library since it's not merged on master yet. We can't create the library without any references since the code would have to be duplicated elsewhere.

This temporarily references konnectivity-client as a directory rather than a go module.

cc @caesarxuchao

Makefile
```diff
@@ -73,18 +75,18 @@ bin/proxy-server: bin cmd/proxy/main.go proto/agent/agent.pb.go proto/proxy.pb.g
...
+gen: proto/agent/agent.pb.go proto/client/client.pb.go mock_gen
+
+proto/client/client.pb.go: konnectivity-client/proto/client/client.proto
+	protoc -I . konnectivity-client/proto/client/client.proto --go_out=plugins=grpc:.
+	cat hack/go-license-header.txt konnectivity-client/proto/client/client.pb.go > konnectivity-client/proto/client/client.licensed.go
+	mv konnectivity-client/proto/client/client.licensed.go konnectivity-client/proto/client/client.pb.go
 
 proto/agent/agent.pb.go: proto/agent/agent.proto
-	protoc -I proto proto/agent/agent.proto --go_out=plugins=grpc:proto
+	protoc -I ${GOPATH}/src sigs.k8s.io/apiserver-network-proxy/proto/agent/agent.proto --go_out=plugins=grpc:${GOPATH}/src
```
caesarxuchao: See if you like this better: https://github.com/caesarxuchao/apiserver-network-proxy/commit/47d2b878dc877d87f2fd9992c5da0b3d3056a44c

I can't avoid using "GOPATH" either.

Jefftree: Adopted your changes. :smile:

go.mod
```diff
@@ -5,18 +5,22 @@ go 1.12
...
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/pflag v1.0.5
 	golang.org/x/net v0.0.0-20191004110552-13f9640d40b9
-	google.golang.org/grpc v1.26.0
+	google.golang.org/grpc v1.27.0
 	k8s.io/api v0.17.1
 	k8s.io/apimachinery v0.17.1
 	k8s.io/client-go v0.17.1
 	k8s.io/klog v1.0.0
+	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.0
```
caesarxuchao: Do you know why it has to be "v0.0.0"?

Jefftree: A version number must exist (for go.mod file parsing), although it doesn't necessarily have to be valid. I'm following k/k's conventions of v0.0.0 for local references.

go.sum
```diff
@@ -236,10 +250,17 @@ honnef.co/go/tools v0.0.0-20190106161140-3f1c8253044a/go.mod h1:rf3lG4BRIbNafJWh
...
 k8s.io/api v0.17.1/go.mod h1:zxiAc5y8Ngn4fmhWUtSxuUlkfz1ixT7j9wESokELzOg=
+k8s.io/api v0.17.2 h1:NF1UFXcKN7/OOv1uxdRz3qfra8AHsPav5M93hlV9+Dc=
+k8s.io/api v0.17.2/go.mod h1:BS9fjjLc4CMuqfSO8vgbHPKMt5+SF0ET6u/RVDihTo4=
 k8s.io/apimachinery v0.17.1 h1:zUjS3szTxoUjTDYNvdFkYt2uMEXLcthcbp+7uZvWhYM=
 k8s.io/apimachinery v0.17.1/go.mod h1:b9qmWdKlLuU9EBh+06BtLcSf/Mu89rWL33naRxs1uZg=
+k8s.io/apimachinery v0.17.2 h1:hwDQQFbdRlpnnsR64Asdi55GyCaIP/3WQpMmbNBeWr4=
+k8s.io/apimachinery v0.17.2/go.mod h1:b9qmWdKlLuU9EBh+06BtLcSf/Mu89rWL33naRxs1uZg=
 k8s.io/client-go v0.17.1 h1:LbbuZ5tI7OYx4et5DfRFcJuoojvpYO0c7vps2rgJsHY=
 k8s.io/client-go v0.17.1/go.mod h1:HZtHJSC/VuSHcETN9QA5QDZky1tXiYrkF/7t7vRpO1A=
+k8s.io/client-go v1.5.1 h1:XaX/lo2/u3/pmFau8HN+sB5C/b4dc4Dmm2eXjBH4p1E=
```
cheftako: Would be nice to understand why we have such disparate versions of our dependencies in here and see if some of this can be cleaned up.

Jefftree: I did an upgrade when initializing the go.mod for konnectivity-client so the versions were mismatched between the client module and the network proxy module. Reverted those changes and there shouldn't be any modifications to go.sum

proto/agent/agent.proto
```diff
@@ -14,87 +14,11 @@
...
 
-// Retransmit?
-// Sliding windows?
+option go_package = "sigs.k8s.io/apiserver-network-proxy/proto/agent";
 
-
-service ProxyService {
-  rpc Proxy(stream Packet) returns (stream Packet) {}
-}
+import "konnectivity-client/proto/client/client.proto";
```
cheftako: Eventually I think we are going to want to break out the agent.proto in a similar way. We should consider making a .proto which is the actual protocol and two service level files which each import the the protocol.

Issue #62, opened 2020-02-12
Use sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.5?


liggitt:
Now that the client is tagged, should the root go.mod switch to using that version?

Jefftree:
We are planning to keep the version numbers in sync between the root and client code. Since a commit in the root repository tagged v0.0.x contains the code for a v0.0.x client, pointing to the local directory should achieve the same effect as specifying the corresponding version. This also gives us more flexibility as the master branches of the two modules will be able to reference each other without needing to cut a new release.

/cc @caesarxuchao

liggitt:
> pointing to the local directory should achieve the same effect as specifying the corresponding version

it also doesn't work for anyone using `go get` to fetch this module

liggitt:
Using v0.0.0 and a replace directive effectively makes the root module unsuitable for anyone to depend on. If the intent is to only support the client module as a library, and only support the top-level module as a built binary (e.g. to not support `go get` or use as a library), that's worth calling out

Jefftree:
> If the intent is to only support the client module as a library, and only support the top-level module as a built binary (e.g. to not support go get or use as a library), that's worth calling out

ack, I believe this was the intent with creating a separate client library for others to import.

PR #63, opened 2020-02-12
Point konnectivity-client to v0.0.5

Jefftree:
Fixes https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/62

/assign @caesarxuchao

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,6 @@ require (
 	k8s.io/apimachinery v0.17.1
 	k8s.io/client-go v0.17.1
 	k8s.io/klog v1.0.0
-	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.0
+	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.5
 )
 
-replace sigs.k8s.io/apiserver-network-proxy/konnectivity-client => ./konnectivity-client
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -19,7 +19,6 @@ require (
 	k8s.io/apimachinery v0.17.1
 	k8s.io/client-go v0.17.1
 	k8s.io/klog v1.0.0
-	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.0
+	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.5
 )
 
-replace sigs.k8s.io/apiserver-network-proxy/konnectivity-client => ./konnectivity-client
```
caesarxuchao: I think we should keep this "replace", following the [example](https://github.com/kubernetes/kubernetes/blob/f7eafa1a838c097c9b6fff118e24bf8334c34f32/go.mod#L537) of Kubernetes staging repositories. This makes sure the code in master branch works together.

We need to have a skew test between the client and the server, but that's a separate issue.

Jefftree: I imagined a scenario where the proxy and client code have different versions, but in theory that should never happen since we synchronize the releases. Hmm, if the master branch of the two libraries are always compatible, I guess we have no need for this PR since any version of the proxy code contains the same version client code.

caesarxuchao: If we check in more code into the proxy client but haven't yet cut a release, this PR will leave the code untested.

Jefftree:
/close

PR seems redundant.

Issue #66, opened 2020-02-13
Create README for konnectivity-client


Jefftree:
We split the proxy and client code. A README should be added documenting how to use the client in other modules.

PR #67, opened 2020-02-13
Add note about importing root module

Jefftree:
Fixes https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/62

Also created https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/66 to add documentation for the client library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ build docker/proxy-agent
 
 ## Examples
 
-The current examples run two actual services as well as a sample client on one end and a sample destination for 
+The current examples run two actual services as well as a sample client on one end and a sample destination for
 requests on the other.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,7 +109,11 @@ curl -v -p --proxy-key certs/master/private/proxy-client.key --proxy-cert certs/
 ```
 
 ### Running on kubernetes
-See following [README.md](examples/kubernetes/README.md) 
+See following [README.md](examples/kubernetes/README.md)
+
+### Clients
+
+`apiserver-network-proxy` is intended to be run as standalone binaries, and should not be imported as a library. Clients communicating with the network proxy can import the `konnectivity-client` module.
```
Sh4d1: `apiserver-network-proxy` components are intended
makes more sense no?

PR #68, opened 2020-02-14
Avoid only listening on localhost for admin server

Jefftree:
The admin server listens solely on localhost, and the assumption that the IP address of 127.0.0.1 is used to health check the konnectivity agent should not be made. 

when `hostNetwork` for the agent pod is turned on in a kubernetes cluster, kubelet can health check the agent by hitting 127.0.0.1. However, with `hostNetwork` off, kubelet is unable to reach the health check since it'll be pinging pod IP, but the admin server will only be listening on 127.0.0.1.

This is only a problem for the agent since the proxy server requires `hostNetwork`.

/assign @caesarxuchao

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -64,6 +64,10 @@ type GrpcProxyAgentOptions struct {
 	proxyServerHost string
 	proxyServerPort int
 
+	// Ports for the health and admin server
+	healthServerPort int
+	adminServerPort  int
+
 	agentID           string
 	syncInterval      time.Duration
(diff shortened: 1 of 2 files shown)


cheftako:
You make a compelling case for wanting health check to NOT be localhost only. However many of the other admin features we may want to lock down. We should consider either adding auth to these or splitting off health check before listening beyond localhost.

Jefftree:
@cheftako Split the health checks into a separate server. The admin server still listens on localhost. PTAL, Thanks!

PR #74, opened 2020-02-26
No-op to trigger CI

Jefftree:
I'm looking to trigger the push images CI job (https://k8s-testgrid.appspot.com/sig-cloud-provider-apiserver-network-proxy#apiserver-network-proxy-push-images) with the new v0.0.6 tag. It uses the old v0.0.5 tag for the push since we added the tag after the branch was merged.


Jefftree:
Will merge another PR instead of creating an empty commit.

/close

PR #75, opened 2020-03-02
Fix warning message if agent disconnects

Jefftree:
There is a valid scenario where there is exactly one agent, and it disconnects and reconnects. In this case, the first backend is removed, however I don't think a warning should be logged here.

### DIFF
--- a/pkg/agent/agentserver/backend_manager.go
+++ b/pkg/agent/agentserver/backend_manager.go
@@ -96,7 +96,7 @@ func (s *DefaultBackendManager) RemoveBackend(agentID string, conn agent.AgentSe
 	for i, c := range backends {
 		if c == conn {
 			s.backends[agentID] = append(s.backends[agentID][:i], s.backends[agentID][i+1:]...)
-			if i == 0 && len(s.backends) != 0 {
+			if i == 0 && len(s.backends[agentID]) != 0 {
 				klog.Warningf("this should not happen. Removed connection %v that is not the first connection, remaining connections are %v", conn, s.backends[agentID])
 			}
 			found = true


pkg/agent/agentserver/backend_manager.go
```diff
@@ -96,7 +96,7 @@ func (s *DefaultBackendManager) RemoveBackend(agentID string, conn agent.AgentSe
 	for i, c := range backends {
 		if c == conn {
 			s.backends[agentID] = append(s.backends[agentID][:i], s.backends[agentID][i+1:]...)
-			if i == 0 && len(s.backends) != 0 {
+			if i == 0 && len(s.backends) > 1 {
```
dberkov: how is it possible that connection remains the same in this case?

cheftako: > 0. The warning is for removing conn 0 when its not the last conn. Line 98 has already removed the ith conn. So for the length now to be greater than 1 it must have been greater than 2 prior to removing the ith element.

Jefftree: Ahh, this was a draft PR and I wasn't sure if changing this fixes it. I think this should be `len(s.backends[agentID]) != 0` instead of `len(s.backends) != 0`

Jefftree: Ahh this was a draft PR and I wasn't sure if the change fixes it. So the actual fix should be `len(s.backends[agentID]) != 0` instead of `len(s.backends) != 0`

cheftako: +1

PR #77, opened 2020-03-03
Only exit on os.Interrupt

anfernee:
Don't quit on other signals like SIGURG.

### DIFF
--- a/tests/slow_http_test.go
+++ b/tests/slow_http_test.go
@@ -1,0 +1,151 @@
+package tests
+
+import (
+	"bytes"
+	"crypto/tls"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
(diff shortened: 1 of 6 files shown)


tests/slow_http_test.go
```diff
@@ -0,0 +1,151 @@
...
+import (
+	"bytes"
+	"crypto/tls"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"google.golang.org/grpc"
+	"k8s.io/klog"
```
zanetworker, line 12: I guess this needs to be updated to klog/v2 (i.e., needs rebase).

cheftako:
This actually seems like a fairly valuable change. Especially if we add a thread dump on something like SIGINT or SIGUSR

PR #78, opened 2020-03-03
Fix races

### LINKED ISSUE #76: Network Proxy hangs when large amount of data is transferred
This was observed with `kubectl cp` on a large binary file. I attempted to run kubectl cp on a 57M file.

To reproduce, start a kubernetes cluster with network proxy enabled

`KUBE_ENABLE_EGRESS_VIA_KONNECTIVITY_SERVICE=true ./cluster/kube-up.sh`

SSH onto the master and try copying the `kubectl` binary to a random pod:

Eg: 

```
kubectl apply -f https://k8s.io/examples/application/shell-demo.yaml
kubectl cp $(which kubectl) shell-demo:kubectl
```

The `konnectivity-server.log` file shows data transferred, but it stops partially through the data transfer and hangs

EDIT: This can be reproduced without kubernetes

1. Build network proxy with certs & protos generated. (`make build && make gen && make certs`)
1. Copy the kubectl binary (or a similar sized file) to the network proxy dir. (cp $(`which kubectl) .`)
1. Run 4 terminals:

HTTP Connect mode fails:

```
./bin/proxy-server --uds-name=konnectivity-server.socket --mode=http-connect --cluster-cert=certs/agent/issued/proxy-master.crt --cluster-key=certs/agent/private/proxy-master.key --server-port=
./bin/proxy-agent --ca-cert=certs/agent/issued/ca.crt --agent-cert=certs/agent/issued/proxy-agent.crt --agent-key=certs/agent/private/proxy-agent.key
python -m SimpleHTTPServer 8001
./bin/proxy-test-client --proxy-uds=konnectivity-server.socket --proxy-host= --proxy-port=0 --mode=http-connect --request-port=8001 --request-host=localhost --request-path=/kubectl
```

GRPC mode also fails

```
./bin/proxy-server --uds-name=konnectivity-server.socket --mode=grpc --cluster-cert=certs/agent/issued/proxy-master.crt --cluster-key=certs/agent/private/proxy-master.key --server-port=
./bin/proxy-agent --ca-cert=certs/agent/issued/ca.crt --agent-cert=certs/agent/issued/proxy-agent.crt --agent-key=certs/agent/private/proxy-agent.key
python -m SimpleHTTPServer 8001
./bin/proxy-test-client --proxy-uds=konnectivity-server.socket --mode=grpc --proxy-host= --proxy-port=0 --mode=http-connect --request-port=8001 --request-host=localhost --request-path=/kubectl
```

caesarxuchao:
Add more logging.

Add lock to protect backend. It is not safe to call Send on a GRPC
stream from multiple Goroutines.

Add lock to protect pendingDials.

Seed the random generator before using it.
/assign @Jefftree @anfernee  @cheftako 

`kubectl cp` ran successfully 11 times in a row with this fix.

Opened issues for these two followup tasks:
1. #85: update the concurrency test, it should have caught the issue
2. #84: we use a mutex to protect the backend stream. Need to investigate if there is a more efficient implementation with channels.

Fixes #76.

### DIFF
--- a/pkg/agent/agentclient/client.go
+++ b/pkg/agent/agentclient/client.go
@@ -74,7 +74,7 @@ func (c *connContext) cleanup() {
 	c.cleanOnce.Do(c.cleanFunc)
 }
 
-// Connect connnects to proxy server to establish a gRPC stream,
+// Connect connects to proxy server to establish a gRPC stream,
 // on which the proxied traffic is multiplexed through the stream
 // and piped to the local connection. It register itself as a
 // backend from proxy server, so proxy server will route traffic
@@ -230,7 +230,7 @@ func (a *AgentClient) remoteToProxy(conn net.Conn, connID int64) {
 
 	for {
 		n, err := conn.Read(buf[:])
-		klog.Infof("received %d bytes from proxy server", n)
+		klog.Infof("received %d bytes from remote for connID[%d]", n, connID)
 
 		if err == io.EOF {
 			klog.Info("connection EOF")
@@ -264,8 +264,10 @@ func (a *AgentClient) proxyToRemote(conn net.Conn, connID int64) {
 		for {
 			n, err := conn.Write(d[pos:])
 			if err == nil {
+				klog.Infof("[connID: %d] write last %d data to remote", connID, n)
 				break
 			} else if n > 0 {
+				klog.Infof("[connID: %d] write %d data to remote with error: %v", connID, n, err)
 				pos += n
 			} else {
 				klog.Errorf("conn write error: %v", err)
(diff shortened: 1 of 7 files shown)


pkg/agent/agentclient/client.go
```diff
@@ -264,8 +264,10 @@ func (a *AgentClient) proxyToRemote(conn net.Conn, connID int64) {
 		for {
 			n, err := conn.Write(d[pos:])
 			if err == nil {
+				klog.Infof("[connID: %d] write last %d data to remote", connID, n)
```
cheftako: Might be worth logging len(d). Default implementation seems like it won't return other than err or n == len(d) but in case...

pkg/agent/agentclient/client.go
```diff
@@ -264,8 +264,10 @@ func (a *AgentClient) proxyToRemote(conn net.Conn, connID int64) {
...
 				break
 			} else if n > 0 {
+				klog.Infof("[connID: %d] write %d data to remote", connID, n)
```
cheftako: err != nil in this case correct? Might be worth logging the err.

pkg/agent/agentclient/client.go
caesarxuchao: After reading this carefully, I think we shouldn't log len(d). The write is not guaranteed to write the entire d[pos:] to the conn. Writing d in pieces is the purpose of this for loop.

pkg/agent/agentclient/client.go
caesarxuchao: Done.

cheftako:
> /lgtm thanks for catching that. I am still wondering why concurrent test didn't capture this. I had more tests . I hope the new test could capture this kind of errors

I think the large data transfer test jying ran caused a long lived request/connection. That dramatically increased the window to actually generate the issue.

caesarxuchao:
Fixed the test for the type change. PTAL. Thanks.

PR #89, opened 2020-03-05
Protect agent connContext

caesarxuchao:
ctx.cleanup are running in go routines and it modifies the map.

I noticed this when debugging e2e issue. This one doesn't really cause any e2e failure.

### DIFF
--- a/pkg/agent/agentclient/client.go
+++ b/pkg/agent/agentclient/client.go
@@ -27,11 +27,54 @@ import (
 	"sigs.k8s.io/apiserver-network-proxy/konnectivity-client/proto/client"
 )
 
+// connContext tracks a connection from agent to node network.
+type connContext struct {
+	conn      net.Conn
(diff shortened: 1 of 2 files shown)


pkg/agent/agentclient/client.go
```diff
@@ -27,11 +27,54 @@ import (
...
+func (cm *connectionManager) Add(connID int64, ctx *connContext) {
+	cm.mu.Lock()
+	defer cm.mu.Unlock()
+	cm.connections[connID] = ctx
+}
+
+func (cm *connectionManager) Get(connID int64) (*connContext, bool) {
+	cm.mu.RLock()
+	defer cm.mu.RUnlock()
+	ctx, ok := cm.connections[connID]
```
Jefftree: nit: Could do just `return cm.connections[connID]`. May affect readability so I'm not too sure

PR #90, opened 2020-03-05
Quick return

caesarxuchao:
First commit is #89.

The data is not complete send() fails once. There is no point sending the remaining data to the backend. If the backend is permanently broken, this would actually cause an infinite loop.


Jefftree:
The break makes a lot of sense. We have a lot of tests we need to add related to connection drops. lgtm from me

PR #91, opened 2020-03-09
Restructure the directories

caesarxuchao:
pkg/agent/agentclient -> pkg/agent
pkg/agent/agentserver -> pkg/server
cmd/proxy -> cmd/server

Also opened https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/92 for further restructure.

/assign @cheftako @Jefftree

### DIFF
--- a/tests/proxy_test.go
+++ b/tests/proxy_test.go
@@ -14,11 +14,11 @@ import (
 
 	"github.com/google/uuid"
 	"google.golang.org/grpc"
-	"sigs.k8s.io/apiserver-network-proxy/pkg/agent/agentclient"
-	"sigs.k8s.io/apiserver-network-proxy/pkg/agent/agentserver"
 	"sigs.k8s.io/apiserver-network-proxy/konnectivity-client/pkg/client"
 	clientproto "sigs.k8s.io/apiserver-network-proxy/konnectivity-client/proto/client"
-	"sigs.k8s.io/apiserver-network-proxy/proto/agent"
+	"sigs.k8s.io/apiserver-network-proxy/pkg/agent"
+	"sigs.k8s.io/apiserver-network-proxy/pkg/server"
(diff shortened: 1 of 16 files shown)


Jefftree:
Looks like a noop on functionality so LGTM for the code changes.

One small nit: We define the two components as "proxy server" and "proxy agent". Does it make sense to use "proxy" and "agent" instead of something like "server"/"agent" or "proxyserver"/"proxyagent"? I understand this is the structure used in the cmd directory but curious on if we want to rethink it.

caesarxuchao:
Updated to what we agreed on yesterday. PTAL.

tests/proxy_test.go
```diff
@@ -223,12 +223,12 @@ func runGRPCProxyServer() (proxy, func(), error) {
...
 }
 
-func runGRPCProxyServerWithServerCount(serverCount int) (proxy, *agentserver.ProxyServer, func(), error) {
+func runGRPCProxyServerWithServerCount(serverCount int) (proxy, *server.ProxyServer, func(), error) {
 	var proxy proxy
 	var err error
 	var lis, lis2 net.Listener
 
-	server := agentserver.NewProxyServer(uuid.New().String(), serverCount, &agentserver.AgentTokenAuthenticationOptions{})
+	server := server.NewProxyServer(uuid.New().String(), serverCount, &server.AgentTokenAuthenticationOptions{})
```
Jefftree: nit: How are these cases usually handled where the variable and imported package name is the same? Just curious why this line compiles without problem but L268 needed to be renamed

caesarxuchao: I think after the variable server is declared, it takes higher priority than the package server. As long as I don't refer to the package server after this line, it compiles.

Jefftree: Ah I see. If that's the case I'd prefer just renaming all `server` variable names to prevent ambiguity. Not going to block this PR though, LGTM.

Issue #86, opened 2020-03-05
[GRPC Mode] Client -> Proxy connections should be closed when finished


Jefftree:
When running in GRPC mode,  client -> proxy UDS connections are not closed after the (client -> proxy server -> agent -> destination) connection is closed. After a `CLOSE_RSP` It seems that we only remove the frontends, but never call close on the underlying GRPC stream between the client and proxy. This causes resource leaks.

Should be an easy fix though.

After just a couple of operations on the master, the number of opened streams can get quite high. This does not happen in http-connect mode.

```
jying@kubernetes-master ~ $ netstat | grep konnectivity-server.socket | wc -l
204
```

/cc @caesarxuchao

caesarxuchao:
The fix should be in k/k?

Jefftree:
We should be able to close a stream from either end. It'll probably be more convenient to implement this on the server side.

caesarxuchao:
#93 was a temporary solution. I think @Jefftree is working on a more thorough solution. Let's keep this bug open in the meantime.

Jefftree:
After performance testing with reusing connections/streams, I think the final verdict is that we'll keep the current 1 dial = 1 connection mapping.

PR #93, opened 2020-03-16
Close UDS socket connection on completion in GRPC mode

Jefftree:
Close the UDS connection when we call close on the underlying proxy connection. Also, break out of loop and stop receiving data after we get a CLOSE_RSP

Fixes https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/86

### DIFF
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -72,12 +72,14 @@ func CreateGrpcTunnel(address string, opts ...grpc.DialOption) (Tunnel, error) {
 		conns:       make(map[int64]*conn),
 	}
 
-	go tunnel.serve()
+	go tunnel.serve(c)
 
 	return tunnel, nil
 }
 
-func (t *grpcTunnel) serve() {
+func (t *grpcTunnel) serve(c *grpc.ClientConn) {
+	defer c.Close()
+
 	for {
 		pkt, err := t.stream.Recv()
 		if err == io.EOF {
@@ -130,9 +132,9 @@ func (t *grpcTunnel) serve() {
 				t.connsLock.Lock()
 				delete(t.conns, resp.ConnectID)
 				t.connsLock.Unlock()
-			} else {
-				klog.Warningf("connection id %d not recognized", resp.ConnectID)
+				return
 			}
+			klog.Warningf("connection id %d not recognized", resp.ConnectID)
 		}
 	}
 }
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -51,9 +51,11 @@ type grpcTunnel struct {
 	connsLock       sync.RWMutex
 }
 
-// CreateGrpcTunnel creates a Tunnel to dial to a remote server through a
+// CreateSingleUseGrpcTunnel creates a Tunnel to dial to a remote server through a
 // gRPC based proxy service.
-func CreateGrpcTunnel(address string, opts ...grpc.DialOption) (Tunnel, error) {
+// Currently, a single tunnel supports a single connection, and the tunnel is closed when the connection is terminated
+// The Dial() method of the returned tunnel should only be called once
+func CreateSingleUseGrpcTunnel(address string, opts ...grpc.DialOption) (Tunnel, error) {
 	c, err := grpc.Dial(address, opts...)
 	if err != nil {
 		return nil, err
--- a/cmd/client/main.go
+++ b/cmd/client/main.go
@@ -264,7 +264,7 @@ func (c *Client) getUDSDialer(o *GrpcProxyClientOptions) (func(ctx context.Conte
 			}
 			return c, err
 		})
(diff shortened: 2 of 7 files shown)


konnectivity-client/pkg/client/client.go
```diff
@@ -130,9 +132,9 @@ func (t *grpcTunnel) serve() {
 				t.connsLock.Lock()
 				delete(t.conns, resp.ConnectID)
 				t.connsLock.Unlock()
-			} else {
-				klog.Warningf("connection id %d not recognized", resp.ConnectID)
+				return
```
Jefftree: Using return is cleaner than break because we would need to break out of both the `switch` statement and the `for` loop.

konnectivity-client/pkg/client/client.go
```diff
@@ -130,9 +132,9 @@ func (t *grpcTunnel) serve() {
...
 			}
+			klog.Warningf("connection id %d not recognized", resp.ConnectID)
```
dberkov: this warning will be appear each time if connection was found

konnectivity-client/pkg/client/client.go
caesarxuchao: The grpcTunnel is meant to be a persistent "tunnel". Clients will only create the grpcTunnel once, and every time they need to communicate to a remote host, then call the grpcTunnel.Dial(). The tunnel supports multiple communication streams because these streams are uniquely identified by the "connectionID".

In this sense, when one stream is closed, we shouldn't close the tunnel, because there could be other ongoing streams multiplexed on this tunnel.

Kubernetes doesn't use the grpcTunnel this way. When the apiserver needs a network connection to the remote, it always creates a new grpc connection to the proxy server, sends a dial request over the tunnel and gets a connection back, see [1](https://github.com/kubernetes/kubernetes/blob/695cd49c40fb174728245814fd2db25f9b801d42/staging/src/k8s.io/apiserver/pkg/server/egressselector/egress_selector.go#L228-L233).

I think another to fix the too many open UDS connection is to fix Kubernetes. The apiserver should maintain and reuse a single grpcTunnel. This might bring some performance improvements as well. This would be a big change.

Alternatively, as a quick fix, we can create a "singleUseGRPCTunnel" object, and make sure that clients can only use this tunnel to dial once. And docuement that when the connection over this tunnel is closed, the tunnel will also be closed.

Jefftree: That makes sense. I created https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/101 and will follow up. I like the "singleUseGRPCTunnel" idea, but we will probably need more testing for failure conditions (connection broken, health checks, etc). In the short run, is this PR needed to avoid creating too many streams/tunnels?

caesarxuchao: This PR is not mergeable as it is.

> but we will probably need more testing for failure conditions (connection broken, health checks, etc)?

Why do we need more testing? In case we are not on the same page, in my mind, the `singleUseGRPCTunnel` will be very similar to the `grpcTunnel`, but instead of managing a pool of pending dials and a pool of connections, it only manages one pending dial and one connection. And it stops when it receives the CLOSE_RSP.

Jefftree: Created the `singleUseGRPCTunnel` object and renamed the tunnel creation function to `CreateSingleUseGRPCTunnel`. I will do some more investigation and testing with https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/101

konnectivity-client/pkg/client/client.go
```diff
@@ -43,17 +43,18 @@ type dialResult struct {
 }
 
 // grpcTunnel implements Tunnel
-type grpcTunnel struct {
+type singleUseGrpcTunnel struct {
```
caesarxuchao, line 46: Can we keep the grpcTunnel, and make a new struct singleUseGRPCTunnel? Btw, the singleUseGRPCTunnel doesn't need maps for connections and pending Dials.

Jefftree: Renamed this back to grpcTunnel. I think it's fine to just use this struct instead of creating a new one since the goal is to eventually replace this with connection multiplexing.

konnectivity-client/pkg/client/client.go
```diff
@@ -51,9 +51,10 @@ type grpcTunnel struct {
 	connsLock       sync.RWMutex
 }
 
-// CreateGrpcTunnel creates a Tunnel to dial to a remote server through a
+// CreateSingleUseGrpcTunnel creates a Tunnel to dial to a remote server through a
 // gRPC based proxy service.
-func CreateGrpcTunnel(address string, opts ...grpc.DialOption) (Tunnel, error) {
+// Currently, a single tunnel supports a single connection, and the tunnel is closed when the connection is terminated
```
caesarxuchao, line 56: To be extra clear, can you append to the comment that ", thus the Dial() method of the returned tunnel should only be called once."?

caesarxuchao:
Have we created an issue to track the final state we want to be in?

PR #94, opened 2020-03-17
Don't add frontend on proxy server if DIAL_RSP contains error

Jefftree:
/assign @caesarxuchao 
/cc @cheftako

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -55,6 +55,9 @@ func (c *ProxyClientConnection) send(pkt *client.Packet) error {
 			_, err := c.HTTP.Write(pkt.GetData().Data)
 			return err
 		} else if pkt.Type == client.PacketType_DIAL_RSP {
+			if pkt.GetDialResponse().Error != "" {
+				return c.HTTP.Close()
+			}
 			return nil
 		} else {
 			return fmt.Errorf("attempt to send via unrecognized connection type %v", pkt.Type)
@@ -463,15 +466,22 @@ func (s *ProxyServer) serveRecvBackend(stream agent.AgentService_ConnectServer, 
 			if client, ok := s.PendingDial.Get(resp.Random); !ok {
 				klog.Warning("<<< DialResp not recognized; dropped")
 			} else {
+				dialErr := false
+				if resp.Error != "" {
+					klog.Warningf("<<< DIAL_RSP contains error: %v", resp.Error)
+					dialErr = true
+				}
 				err := client.send(pkt)
 				s.PendingDial.Remove(resp.Random)
-				if resp.Error != "" {
-					klog.Warningf("<<< DIAL_RSP received error: %v", resp.Error)
-					break
-				} else if err != nil {
+				if err != nil {
 					klog.Warningf("<<< DIAL_RSP send to client stream error: %v", err)
+					dialErr = true
+				}
+				// Avoid adding the frontend if there was an error dialing the destination
+				if dialErr == true {
 					break
 				}
+
 				client.connectID = resp.ConnectID
 				client.agentID = agentID
 				s.addFrontend(agentID, resp.ConnectID, client)
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -465,14 +465,17 @@ func (s *ProxyServer) serveRecvBackend(stream agent.AgentService_ConnectServer, 
 			} else {
 				err := client.send(pkt)
 				s.PendingDial.Remove(resp.Random)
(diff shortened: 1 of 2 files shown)


pkg/server/server.go
```diff
@@ -465,14 +468,17 @@ func (s *ProxyServer) serveRecvBackend(stream agent.AgentService_ConnectServer,
 			} else {
 				err := client.send(pkt)
 				s.PendingDial.Remove(resp.Random)
-				if err != nil {
+				if resp.Error != "" {
+					klog.Warningf("<<< DIAL_RSP received error: %v", resp.Error)
+					break
```
caesarxuchao, line 473: We shouldn't break here. One dial failed, but there are other streams on this backend, we should keep processing these other streams.

pkg/server/server.go
```diff
@@ -465,14 +468,17 @@ func (s *ProxyServer) serveRecvBackend(stream agent.AgentService_ConnectServer,
...
+				} else if err != nil {
 					klog.Warningf("<<< DIAL_RSP send to client stream error: %v", err)
-				} else {
-					client.connectID = resp.ConnectID
-					client.agentID = agentID
-					s.addFrontend(agentID, resp.ConnectID, client)
-					close(client.connected)
+					break
```
caesarxuchao, line 476: Same here, we shouldn't break.

pkg/server/server.go
```diff
@@ -465,14 +468,17 @@ func (s *ProxyServer) serveRecvBackend(stream agent.AgentService_ConnectServer,
 			} else {
 				err := client.send(pkt)
 				s.PendingDial.Remove(resp.Random)
-				if err != nil {
+				if resp.Error != "" {
+					klog.Warningf("<<< DIAL_RSP received error: %v", resp.Error)
```
caesarxuchao, line 472: How about "DIAL_RSP contains error"?

pkg/server/server.go
```diff
@@ -465,14 +468,17 @@ func (s *ProxyServer) serveRecvBackend(stream agent.AgentService_ConnectServer,
 			} else {
 				err := client.send(pkt)
 				s.PendingDial.Remove(resp.Random)
-				if err != nil {
+				if resp.Error != "" {
```
caesarxuchao, line 471: I would move this condition before the client.send(), i.e.,

```golang
if resp.Error != "" {
  klog.Warningf("<<< DIAL_RSP contains error: %v", resp.Error)
}
err := client.send(pkt)
s.PendingDial.Remove(resp.Random)
if err != nil {
  klog.Warningf("<<< DIAL_RSP send to client stream error: %v", err)
} else {
  ...
}
```

pkg/server/server.go
Jefftree: This breaks out of the case statement, not the for loop (Only intended to skip L478-481). It's a bit ambiguous which is why I considered changing this to use if statements instead of case. Let me know your thoughts.

pkg/server/server.go
Jefftree: See comment above.

pkg/server/server.go
Jefftree: Took a slightly modified approach of your suggestion.

pkg/server/server.go
caesarxuchao: Got it. I think it's ok leaving it as "switch/case".

tests/proxy_test.go
```diff
@@ -92,6 +92,41 @@ func TestBasicProxy_GRPC(t *testing.T) {
...
+
+	c := &http.Client{
+		Transport: &http.Transport{
+			Dial: tunnel.Dial,
+		},
+	}
+
+	_, err = c.Get(invalidServer.URL)
+	if err == nil {
+		t.Error("Expected error when destination is unreachable, did not receive error")
```
caesarxuchao, line 126: Check the error message?

caesarxuchao:
A nit in the test. Also please squash.

tests/proxy_test.go
Jefftree: Just for clarification, do you mean compare the error message with a predefined string instead of just checking for nil?

caesarxuchao: Right.

Jefftree: httptest generates a random port for the server, and the error message is of the form: `Get "http://127.0.0.1:46415": dial tcp 127.0.0.1:46415: connect: connection refused`. I checked for the existence of `connection refused` in the error message instead.

PR #96, opened 2020-03-26
add health server for proxy

### LINKED ISSUE #95: Proxy also need to have a separate health port
We are not running the proxy server in hostNetwork, so we can't have a livenessprobe since the admin port is only listening on localhost

Sh4d1:
Fixes #95 

Signed-off-by: Patrik Cyvoct <patrik@ptrk.io>


cheftako:
Seems like a follow on to #68

Sh4d1:
@cheftako yep, though I did not understand why it's stated that the proxy is always running in host network :sweat_smile:

examples/kubernetes/konnectivity-server.yaml
```diff
@@ -35,7 +36,7 @@ spec:
       httpGet:
         scheme: HTTP
         host: 127.0.0.1
-        port: 8092
+        port: 8093
```
cheftako: We should probably add health in the ports section (below) as well.

cheftako:
> Walter Fender yep, though I did not understand why it's stated that the proxy is always running in host network

Probably an assumption on our part. The view was that it would be a master component like the scheduler or the kube-controller-manager. Further that it would have a kubelet monitoring it locally and acting as its nanny, on the host network. We're so used to that model we didn't take the time to consider others. Not saying that other models are wrong and in fact would love to understand yours better, so we can better support it.

examples/kubernetes/konnectivity-server.yaml
Sh4d1: done

Sh4d1:
We are basically running k8s in k8s, so each master components are pods in another cluster :)

Jefftree:
@Sh4d1: Out of curiosity, in your use case is there a guarantee that all master components (for any k8s cluster) are colocated on the same host? That was one of our assumptions for using hostNetwork on the proxy and we built certain components like UDS and token based authentication revolving around that. Would love to learn more about your use case.

Sh4d1:
@jefftree basically one the master we have 2 control places, ccm, csi, autoscaler and a custom made component. In each control plane there is API, controller, scheduler and VPN/network proxy. Each control plane is a pod so each API/controller... Is guarantee to be on the same host. But there are different control plane on the same host, hence the host network not really possible 😄

Jefftree:
Thanks for the clarification. I would say using hostNetwork is still possible as long as the set of ports do not conflict. I haven't done much testing with hostNetwork disabled, but theoretically it should also work with some minor changes.

Anyway, if you'd like to avoid using hostNetwork, then this change LGTM. Let us know how you'd like to proceed.

Sh4d1:
It works great with host network disabled, been running it like this from the start 😄

cmd/server/main.go
```diff
@@ -256,6 +267,7 @@ func newProxyRunOptions() *ProxyRunOptions {
 		serverPort:             8090,
 		agentPort:              8091,
 		adminPort:              8092,
+		healthPort:             8093,
```
Jefftree: Sorry just a small nit, can we keep the healthPort at 8092 and move the adminPort to 8093 as the default? This reduces the amount of existing readiness/liveness probe code that needs to be changed.

Sh4d1: :heavy_check_mark:

Sh4d1:
@Jefftree  sorry forgot to add a file :sweat_smile: good now

zanetworker:
/lgtm 

just tested it locally from my side after adjusting the livnesspobes, cluster connectivity looks good. 

```
W0403 08:43:28.620971       1 main.go:135] Server port set to 8131.
W0403 08:43:28.620976       1 main.go:136] Agent port set to 8132.
W0403 08:43:28.620981       1 main.go:137] Admin port set to 8133.
W0403 08:43:28.620986       1 main.go:138] Health port set to 8134.
```

Jefftree:
@cheftako Let's merge this?

PR #99, opened 2020-04-06
Log user agent in proxy server

Jefftree:
This logs user agent (caller information) on the proxy server. 

A corresponding PR will be added in k/k to propagate the caller (webhooks, logs, exec, etc)

/assign @caesarxuchao 
/cc @cheftako

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -185,7 +185,12 @@ func NewProxyServer(serverID string, serverCount int, agentAuthenticationOptions
 
 // Proxy handles incoming streams from gRPC frontend.
 func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
-	klog.Info("proxy request from client")
+	md, ok := metadata.FromIncomingContext(stream.Context())
+	if !ok {
+		return fmt.Errorf("failed to get context")
+	}
+	userAgent := md.Get(header.UserAgent)
+	klog.Infof("proxy request from client, userAgent %s", userAgent)
 
 	recvCh := make(chan *client.Packet, 10)
 	stopCh := make(chan error)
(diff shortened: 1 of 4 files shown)


pkg/server/server.go
```diff
@@ -185,7 +185,12 @@ func NewProxyServer(serverID string, serverCount int, agentAuthenticationOptions
 
 // Proxy handles incoming streams from gRPC frontend.
 func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
-	klog.Info("proxy request from client")
+	md, ok := metadata.FromIncomingContext(stream.Context())
+	if !ok {
+		return fmt.Errorf("failed to get context")
+	}
+	userAgent := md.Get("user-agent")
```
zanetworker, line 192: shouldn't this be "User-Agent" instead?

cheftako, line 192: As mentioned by Adel, this should be "User-Agent". RFC 7231 specifies it this was and as a parameter name it would be case sensitive.

Jefftree: GRPC metadata pairs are normalized as lowercase strings. (https://github.com/grpc/grpc-go/blob/master/metadata/metadata.go#L110). The examples I looked at used lowercase representations for "user-agent" and I tried to follow that.

cheftako: Goo point. As long as we're not setting it in the HTTP headers, then this should be correct.

caesarxuchao: Also can you define it in the headers package?

caesarxuchao:
Otherwise lgtm.

Jefftree:
Updated the definition in the headers package. 

@caesarxuchao PTAL

PR #102, opened 2020-04-20
Implement the readiness check of the proxy server

caesarxuchao:
It only returns ready if the proxy server has connections to the proxy agents.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -545,7 +545,14 @@ func (p *Proxy) runHealthServer(o *ProxyRunOptions, server *server.ProxyServer) 
 		fmt.Fprintf(w, "ok")
 	})
 	readinessHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		fmt.Fprintf(w, "ok")
+		ready, msg := server.Readiness.Ready()
+		if ready {
+			w.WriteHeader(200)
+			fmt.Fprintf(w, "ok")
+			return
+		}
+		w.WriteHeader(500)
+		fmt.Fprintf(w, msg)
 	})
 
 	muxHandler := http.NewServeMux()
--- a/pkg/server/readiness_manager.go
+++ b/pkg/server/readiness_manager.go
@@ -1,0 +1,36 @@
+/*
+copyright 2020 the kubernetes authors.
+
+licensed under the apache license, version 2.0 (the "license");
+you may not use this file except in compliance with the license.
+you may obtain a copy of the license at
+
+    http://www.apache.org/licenses/license-2.0
+
+unless required by applicable law or agreed to in writing, software
+distributed under the license is distributed on an "as is" basis,
+without warranties or conditions of any kind, either express or implied.
+see the license for the specific language governing permissions and
+limitations under the license.
+*/
+
+package server
+
+// ReadinessManager supports checking if the proxy server is ready.
+type ReadinessManager interface {
+	// Ready returns if the proxy server is ready. If not, also return an
+	// error message.
+	Ready() (bool, string)
+}
+
+var _ ReadinessManager = &DefaultBackendManager{}
+
+func (s *DefaultBackendManager) Ready() (bool, string) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	if len(s.backends) == 0 {
+		return false, "no connection to any proxy agent"
+	}
+	return true, ""
+
+}
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -102,6 +102,13 @@ type ProxyServer struct {
 	// BackendManager manages the backends.
 	BackendManager BackendManager
 
+	// Readiness reports if the proxy server is ready, i.e., if the proxy
+	// server has connections to proxy agents (backends). Note that the
+	// proxy server does not check the healthiness of the connections,
+	// though the proxy agents do, so this readiness check might report
+	// ready but there is no healthy connection.
+	Readiness ReadinessManager
+
 	// fmu protects frontends.
 	fmu sync.RWMutex
 	// conn = Frontend[agentID][connID]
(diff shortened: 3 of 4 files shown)


pkg/server/server.go
```diff
@@ -172,13 +179,15 @@ func (s *ProxyServer) getFrontend(agentID string, connID int64) (*ProxyClientCon
...
 		frontends:                  make(map[string](map[int64]*ProxyClientConnection)),
 		PendingDial:                NewPendingDialManager(),
 		serverID:                   serverID,
 		serverCount:                serverCount,
-		BackendManager:             NewDefaultBackendManager(),
+		BackendManager:             bm,
 		AgentAuthenticationOptions: agentAuthenticationOptions,
+		Readiness:                  bm,
 	}
 	return p
```
zanetworker, line 192: Maybe simply return without the extra variable? 
```go 
	return  &ProxyServer{
		frontends:                  make(map[string](map[int64]*ProxyClientConnection)),
		PendingDial:                NewPendingDialManager(),
		serverID:                   serverID,
		serverCount:                serverCount,
		BackendManager:             NewDefaultBackendManager(),
		BackendManager:             bm,
		AgentAuthenticationOptions: agentAuthenticationOptions,
		Readiness:                  bm,
	}
```

caesarxuchao: Done.

caesarxuchao:
> A change is also needed in the health server to call this as part of the readiness check.

I disagree. I think we should distinguish the healthz from the readiness check. Healthz just means the process is still running.

> I also feel that we don't really need the extra Readiness variable (could just use something like BackendManager.Ready(), but don't have a strong preference.

I don't want to mix the interfaces, so I added another interface and variable.

caesarxuchao:
cc @anfernee

pkg/server/server.go
```diff
@@ -175,15 +182,16 @@ func (s *ProxyServer) getFrontend(agentID string, connID int64) (*ProxyClientCon
...
+	bm := NewDefaultBackendManager()
+	return &ProxyServer{
 		frontends:                  make(map[string](map[int64]*ProxyClientConnection)),
 		PendingDial:                NewPendingDialManager(),
 		serverID:                   serverID,
 		serverCount:                serverCount,
-		BackendManager:             NewDefaultBackendManager(),
+		BackendManager:             bm,
 		AgentAuthenticationOptions: agentAuthenticationOptions,
+		Readiness:                  bm,
```
cheftako, line 193: Ideally we should have something (ReadinessManager) which aggregates a list of Readiness checks. Any chance we can add the infra now so its easy to additional readiness checks as we find the need?

cmd/server/main.go
```diff
@@ -545,7 +545,7 @@ func (p *Proxy) runHealthServer(o *ProxyRunOptions, server *server.ProxyServer)
 		fmt.Fprintf(w, "ok")
 	})
 	readinessHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		fmt.Fprintf(w, "ok")
```
cheftako, line 548: Isn't this going to return "true" rather than "ok"?

pkg/server/readiness_manager.go
```diff
@@ -0,0 +1,31 @@
...
+
+// ReadinessManager supports checking if the proxy server is ready.
+type ReadinessManager interface {
+	// Ready returns if the proxy server is ready.
+	Ready() bool
+}
+
+var _ ReadinessManager = &DefaultBackendManager{}
+
+func (s *DefaultBackendManager) Ready() bool {
```
cheftako, line 27: Would it make more sense for this to return either a message string ("ok" vs "error") or return (bool, string) where the string is a message string? , or even (int, string) where the int is the http response code?

pkg/server/server.go
caesarxuchao: I did define the ReadinessManager as an interface. Currently I make the backendManager implements the interface. We can compose a dedicated readinessManager with more checks in the future. Is that ok?

cheftako:
> A change is also needed in the health server to call this as part of the readiness check.

+1, It doesn't need to be part of this PR but we re


> > A change is also needed in the health server to call this as part of the readiness check.
> 
> I disagree. I think we should distinguish the healthz from the readiness check. Healthz just means the process is still running.
> 
We should keep liveness (healthz) and readiness checks separate. I think of healthz as the process is more than just running.
It is also in sufficiently healthz state that we don't need Kubelet to restart it. It may be true that if nothing has connected to us in the last ?hour we want to restart, as this may mean something is wrong and that it might be fixed by restarting. In any case we should make sure we have a liveness check as well. (It can be a separate PR)

cmd/server/main.go
cheftako: Also for non "ok" responses we should be returning an error code via w.WriteHeader().
It needs to be called before we call write. We probably want to make it a 5XX error.

Jefftree:
> > A change is also needed in the health server to call this as part of the readiness check.
> 
> +1, It doesn't need to be part of this PR but we re
> 
> > > A change is also needed in the health server to call this as part of the readiness check.
> > 
> > 
> > I disagree. I think we should distinguish the healthz from the readiness check. Healthz just means the process is still running.
> 
> We should keep liveness (healthz) and readiness checks separate. I think of healthz as the process is more than just running.
> It is also in sufficiently healthz state that we don't need Kubelet to restart it. It may be true that if nothing has connected to us in the last ?hour we want to restart, as this may mean something is wrong and that it might be fixed by restarting. In any case we should make sure we have a liveness check as well. (It can be a separate PR)

That makes sense. Originally I was thinking along the lines of "no agent connection = unhealthy", but I agree that having no agents connected shouldn't force kubelet to restart the pod.

caesarxuchao:
Updated. PTAL. Thanks.

Opened https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/113 to track the discussion about healthz. I agree that "whether we want kubelet to restart the pod" is a good criteria for healthiness.

pkg/server/server.go
cheftako: Sure

pkg/server/server.go
```diff
@@ -102,6 +102,13 @@ type ProxyServer struct {
 	// BackendManager manages the backends.
 	BackendManager BackendManager
 
+	// Readiness reports if the proxy server is ready, i.e., if the proxy
+	// server has connections to proxy agents (backends). Note that the
+	// proxy server does not check the healthiness of the connections,
+	// though the proxy agents do, so this readiness check might report
+	// ready but there is no healthy connection.
+	Readiness ReadinessManager
```
cheftako, line 110: Not needed yet but with most K/K liveness/readiness checks we provide a mechanism to disable names checks. I'm guessing we will eventually want this functionality.

caesarxuchao:
Thanks for the reviews!

/hold cancel

PR #104, opened 2020-05-04
Improve validation to when using SA authentication

### LINKED ISSUE #103: Vague errors when running in SA auth mode while the cluster CA  cert is set
**What Happened**
I was testing out service account authentiction with `authentication-audience` and for some reason it was not working (getting connection closed on the clientside). I was using the following flags:

```
          - --cluster-ca-cert=ca.crt
          - --cluster-cert=konnectivity-server.crt
          - --cluster-key=konnectivity-server.key
          - --agent-namespace=namespace
          - --agent-service-account=konnectivity-agent
          - --kubeconfig=kubeconfig
          - --authentication-audience=system:konnectivity-server
          - --mode=http-connect
```
As I was switching between SA auth mode and certificates, the problem was that the `--cluster-ca-cert` flag was set at the same time as the sa auth flags (auth-audience, agent-namespace, ...) which probably led to the konnectivity server running in certificate auth mode. 

**What I expect** 
If I set the agent-service-account, then cluster auth validation should not be used, or at least validation with a clear error to set the right flags.

zanetworker:
**What this PR does** 
This PR adds validation to prevent setting the clusterCA flag while running in SA auth mode thus avoiding vague connectivity issues. 

Fixes #103 

**Notes** 
Validation is the simple solution to prevent the issue, the other way would be to add logic to detect which mode it should run on based on the flags.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -235,6 +235,9 @@ func (o *ProxyRunOptions) Validate() error {
 	// validate agent authentication params
 	// all 4 parametes must be empty or must have value (except kubeconfigPath that might be empty)
 	if o.agentNamespace != "" || o.agentServiceAccount != "" || o.authenticationAudience != "" || o.kubeconfigPath != "" {
+		if o.clusterCaCert != "" {
+			return fmt.Errorf("clusterCaCert can not be used when service account authentication is enabled")
+		}
 		if o.agentNamespace == "" {
 			return fmt.Errorf("agentNamespace cannot be empty when agent authentication is enabled")
 		}


zanetworker:
@cheftako thats exactly the issue, when both are `on` at the same time nothing happens, at the beginning I thought something is off with the networking, but it was that I had cluster-ca set which turned on that certificate authentication. They won't really be running at the same time for the same direction, for agent->server thats going be SA based, for mtls verification of the server -> cluster, that would be using the clusters certs, thats why we need the CA secret on the agent IIUC.

cheftako:
I agree that you probably don't want cert based auth on when using token based auth. But it looks like if you disable cert based auth you disabling ssh altogether, which you also don't want with token based auth.

zanetworker:
I would say this is not really a fix per say but a way to make the SA mode more usable, i.e., give feedback that when SA mode is to be used, that flag should not be set otherwise neither will work. That said we would still need cert auth as you mentioned, but in this case, we don't provide the cluster-ca only the cluster-cert and cluster-key for SSH to work. So the end-result is if this is the config: 

```
- --cluster-ca-cert=ca.crt
- --cluster-cert=konnectivity-server.crt
- --cluster-key=konnectivity-server.key
- --agent-namespace=namespace
- --agent-service-account=konnectivity-agent
- --kubeconfig=kubeconfig
- --authentication-audience=system:konnectivity-server
- --mode=http-connect
``` 
nothing will work and we give the user a warning during validation, what should be set instead is: 

```
- --cluster-cert=konnectivity-server.crt
- --cluster-key=konnectivity-server.key
- --agent-namespace=namespace
- --agent-service-account=konnectivity-agent
- --kubeconfig=kubeconfig
- --authentication-audience=system:konnectivity-server
- --mode=http-connect
```

cheftako:
My bad. Misread the code around tls config and thought that we might not have an encrypted tunnel when ClusterCACert is missing. Confirmed we do have an encrypted tunnel and spotted the error I made reading the code.
/lgtm

PR #105, opened 2020-05-13
recompile client.proto by protoc-gen-go v1.22.0

RainbowMango:
We want to update `github.com/golang/protobuf` to v1.4.0 from k/k, more details please refer to [the discussion](https://github.com/kubernetes/kubernetes/pull/90582#discussion_r423828387).

But the exported type `InternalMessageInfo` has been deprecated by `github.com/golang/protobuf@v1.4.0`:
https://github.com/golang/protobuf/blob/1b794fe86dd6a0c7c52ae69b5c9cb0aeedc52afa/proto/deprecated.go#L86-L92

So, we should recompile the proto, by newer protoc-gen-go, to get rid of `InternalMessageInfo` reference.


RainbowMango:
In favor of https://github.com/kubernetes/kubernetes/pull/90582.

We don't need to re-generate it now.
/close

PR #106, opened 2020-05-15
Turn on race for tests and fix code.

cheftako:
Added the -race option for the package tests.
Fixed a few race conditions found by this option.
Redo of @anfernee's fix
'Add mutex to pass "go test -race" 43'


Jefftree:
Since the generated proto files were updated, looks like we also need to regenerate the mock file used for the tests `make mock_gen`

> race: limit on 8128 simultaneously alive goroutines is exceeded, dying

This also seems unexpected. We shouldn't have that many go routines in the first place.

cheftako:
> > imports sigs.k8s.io/apiserver-network-proxy/proto/agent/mocks: cannot find module providing package sigs.k8s.io/apiserver-network-proxy/proto/agent/mocks
> 
> Since the generated proto files were updated, looks like we also need to regenerate the mock file used for the tests `make mock_gen`

make clean seems to be nuking the mock files but we aren't calling 'make gen' in our instructions.
Fixing the instructions. (May also fix the Makefile as its clearly missing a dependency with this.

> 
> > race: limit on 8128 simultaneously alive goroutines is exceeded, dying
> 
> This also seems unexpected. We shouldn't have that many go routines in the first place.

I think this has to do with me bumping the test wait to 5 seconds.
However it means that something in that test path is spawning goroutines when it shouldn't.
Sadly this isn't happening locally. So I'm debugging 'remotely' (i.e. here).

cheftako:
> > > race: limit on 8128 simultaneously alive goroutines is exceeded, dying
> > 
> > 
> > This also seems unexpected. We shouldn't have that many go routines in the first place.
> 
> I think this has to do with me bumping the test wait to 5 seconds.
> However it means that something in that test path is spawning goroutines when it shouldn't.
> Sadly this isn't happening locally. So I'm debugging 'remotely' (i.e. here).

So it turns out we have several goroutine leaks. The -race option is very useful but it automatically fails if you reach 8K goroutines, which we were doing very quickly because of the leak. At this point, the leaks hits by the testing are fixed, as are several race conditions. I think we should get this committed soon, as the it will clearly help us from accidentally reintroducing some of these type of errors.

pkg/agent/stream.go
```diff
@@ -367,20 +402,20 @@ func (c *RedialableAgentClient) tryConnect() (connectResult, error) {
 	ctx := metadata.AppendToOutgoingContext(context.Background(), header.AgentID, c.agentID)
 	if c.serviceAccountTokenPath != "" {
 		if ctx, err = c.initializeAuthContext(ctx); err != nil {
-			return connectResult{}, err
+			return connectResult{grpcConn: conn}, err
```
caesarxuchao, line 405: Can we do the conn cleanup in tryConnnect instead of requiring the caller to do it? I think a deferred cleanup is possible.

tests/concurrent_test.go
```diff
@@ -13,6 +13,9 @@ import (
 )
 
 func TestProxy_Concurrency(t *testing.T) {
+	t.Log("### Starting TestProxy_Concurrency ###")
+	defer t.Log("### Stopping TestProxy_Concurrency ###")
```
caesarxuchao, line 17: Why do we need these log lines?

pkg/agent/stream.go
```diff
@@ -302,7 +336,8 @@ func (c *RedialableAgentClient) reconnect() {
 	}
 
 	c.cs.RemoveClient(c.serverID)
-	close(c.stopCh)
+	// close(c.stopCh)
+	c.stopCh = nil
```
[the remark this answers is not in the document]
cheftako: RemoveClient already closes the channel and we don't want to close it twice.

pkg/agent/stream.go
```diff
@@ -111,6 +101,42 @@ func NewRedialableAgentClient(address, agentID string, cs *ClientSet, opts ...gr
 	return c, nil
 }
 
+func (c *RedialableAgentClient) copyRedialableAgentClient() *RedialableAgentClient {
+	out := &RedialableAgentClient{
+		cs:                      c.cs,
```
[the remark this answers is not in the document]
cheftako: Copying the object copies fields we don't want to copy. That then triggers the race warning on fields we didn't even want to copy. So its manually copy what we actually want and not copying the things we don't want. Or we start layering the copy in locks for things we don't care about.

pkg/agent/stream.go
cheftako: Yeah we can do that. I'll send out an updated version of the fix.

tests/concurrent_test.go
cheftako: Crud, I thought I had cleared all these up. These are leftovers from me tracing where the leaked goroutines were coming from. I'll go ahead and clear these up.

pkg/agent/stream.go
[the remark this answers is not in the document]
caesarxuchao: Got it. It's good that we don't copy the locks, otherwise this would be recursive.

It seems that the reconnecting code is causing more and more maintenance burden. Maybe we should remove the functionality entirely.

The advantage of the reconnecting is that it will retry the current packet so it surfaces less errors to the caller. Maybe now the maintenance burden finally overgrows the advantage. What do you think? I know @Jefftree and @dberkov advocates to remove this code as well.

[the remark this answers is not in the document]
caesarxuchao: I can do that in a followup PR if we all agree to remove the code.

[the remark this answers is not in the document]
cheftako: I'm fine with removing that if we're finding it too much of a burden. It also occurred to me that the old code is actually more dangerous. It was relying on a shallow copy. If we add more fields (slices, pointers etc) and don't fix this method, we end up with some fairly nasty bugs. The new and old objects look fine but have shared state. At least this code breaks obviously if you add a field and don't fix this method. The field just isn't set.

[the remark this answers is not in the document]
cheftako: I agree that if we want to remove that functionality it should be a separate PR. Honestly I would prefer to break this PR up into pieces. Unfortunately we wouldn't be able to turn on race testing until all the pieces merged.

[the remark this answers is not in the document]
cheftako: I'm generally ok with it. Retrying packets seems like a nice to have. I would say as long as we have a metric to measure how often that case is hit I'm fine with removing it. That way if it goes from nice to have to a something we need to fix, we will have a way to know.

pkg/agent/client.go
```diff
@@ -97,12 +97,12 @@ func newAgentClientWithRedialableAgentClient(rac *RedialableAgentClient) *AgentC
 }
 
 // Close closes the underlying stream.
-func (c *AgentClient) Close() {
-	if c.stream == nil {
+func (a *AgentClient) Close() {
```
Jefftree, line 100: Not something related to this PR, but I think we should fix change the `stream` variable to something more representative to avoid conflicts between grpc stream and network proxy "stream".

- `a.stream` refers to a RedialableAgentClient which contains a grpc stream and conn.
- `a.stream.close()` closes the underlying network proxy stream (the connection), but doesn't affect the grpc stream.

pkg/agent/stream.go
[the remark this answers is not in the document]
Jefftree: After looking at all the fields copied, it looks like they can all be obtained from the clientset with the exception of serverCount. This shouldn't block the PR but perhaps we don't actually need a field by field copy and can just call `NewAgentClient` from the clientset.

tests/ha_proxy_server_test.go
```diff
@@ -136,6 +138,12 @@ func TestBasicHAProxyServer_GRPC(t *testing.T) {
 		}
 	}
 	if !ready {
+		t.Logf("Proxy-0(%s<->%s), Proxy-1(%s<->%s), Proxy-2(%s<->%s)",
```
caesarxuchao, line 141: Are these also leftover of debugging code? There are a few other places adding t.Logf, can you double check if they are needed?

pkg/agent/clientset.go
```diff
@@ -151,6 +159,11 @@ func (cs *ClientSet) sync() {
 			duration = wait.Jitter(backoff.Duration, backoff.Jitter)
 		}
 		time.Sleep(duration)
+		select {
+			case <-cs.stopCh:
+				break
```
Jefftree, line 164: I don't think this does what is intended. It'll break out of the select statement but the for loop will keep running. Probably need to add a flag?

pkg/agent/client.go
caesarxuchao: Good point. This will be solved if we remove the redialiableAgentClient altogether :)

pkg/agent/clientset.go
caesarxuchao: good catch

Makefile
```diff
@@ -44,7 +44,7 @@ mock_gen:
 
 .PHONY: test
 test:
-	GO111MODULE=on go test ./...
+	GO111MODULE=on go test -v -race ./...
```
Jefftree, line 47: This increases the time to run the tests by quite a bit. Thoughts on adding a new CI job/command for checking races and keeping this command at just `go test ./...`?

tests/ha_proxy_server_test.go
cheftako: Fixed

pkg/agent/clientset.go
cheftako: Good catch. Fixed.

Makefile
cheftako: "time make test" with these setting on my laptop (which isn't very powerful) reports :-
real	1m23.557s
user	3m31.509s
sys	0m25.549s

If I remove the verbose flag, this is reduced to :-
real	0m25.310s
user	0m35.678s
sys	0m9.804s

Assuming this is fast enough?

Jefftree: removing the verbose flag is good enough. Thank you!

Jefftree:
Thanks for the fix! 
/lgtm
/hold (for @caesarxuchao)

caesarxuchao:
I'm working on a patch to remove the retry.

PR #108, opened 2020-05-29
Remove the proxy agent retry logic to simplify the code.

caesarxuchao:
Copied from https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/106#discussion_r430680112:

"It seems that the reconnecting code is causing more and more maintenance burden....The advantage of the reconnecting is that it will retry the current packet so it surfaces less errors to the caller." 

It's a nice-to-have, but we think the maintenance burden has overgrown the advantage, so we remove it.

In a follow-up PR, I'll add client-side metrics to record how many send/receive errors we got, so that we will know if the optimization is needed.


anfernee:
lgtm too. not particularly familiar with the host counting logic. everything else looks good.

pkg/agent/client.go
```diff
@@ -76,33 +85,154 @@ type AgentClient struct {
...
 	connManager *connectionManager
 
-	stream *RedialableAgentClient
-	stopCh <-chan struct{}
+	cs *ClientSet // the clientset that includes this AgentClient.
+
+	stream      agent.AgentService_ConnectClient
+	agentID     string
+	serverID    string // the id of the proxy server this client connects to.
+	serverCount int    // the number of the proxy server instances.
```
cheftako, line 93: Not sure we should keep serverCount here. It seems like there should only be one value for serverCount. Not sure what it means if we somehow get different clients with different values for serverCount. Seems better to just have serverCount in ClientSet.

pkg/agent/client.go
```diff
@@ -76,33 +85,154 @@ type AgentClient struct {
...
+		a.cs.RemoveClient(a.serverID)
+	}
+	return err
+}
+
+func (a *AgentClient) Recv() (*client.Packet, error) {
+	a.recvLock.Lock()
+	defer a.recvLock.Unlock()
+
+	pkt, err := a.stream.Recv()
```
cheftako, line 190: Not blocking for  this PR but a little worried that Recv is blocking and may stop us shutting down properly.

pkg/agent/client.go
```diff
@@ -76,33 +85,154 @@ type AgentClient struct {
...
+	return pkt, err
+}
+
+func serverCount(stream agent.AgentService_ConnectClient) (int, error) {
+	md, err := stream.Header()
+	if err != nil {
+		return 0, err
+	}
+	scounts := md.Get(header.ServerCount)
+	if len(scounts) == 0 {
```
cheftako, line 204: When is it ok to get more than 1 ServerCount? If you get more than 1, how do you know which one is the right value?

pkg/agent/clientset.go
```diff
@@ -175,13 +171,13 @@ func (cs *ClientSet) syncOnce() error {
 	if err != nil {
 		return err
 	}
-	cs.serverCount = c.stream.serverCount
-	if err := cs.AddClient(c.stream.serverID, c); err != nil {
+	cs.serverCount = c.serverCount
```
cheftako, line 174: If cs.serverCount > 0 and cs.serverCount != c.serverCount, we should probably at least print a warning.

pkg/agent/client.go
caesarxuchao: Done. Bubbled up the serverCount as a return value.

pkg/agent/client.go
caesarxuchao: Good point. I did some research on this and it seems fine.

When we close the agent client, we close the underlying GRPC connection, which should close the stream and unblocks Recv() with an error, according to https://github.com/grpc/grpc-go/issues/3230#issuecomment-562242727.

pkg/agent/client.go
caesarxuchao: Updated it to check the length is 1.

pkg/agent/clientset.go
caesarxuchao: Removed c.serverCount per your other comment :)

pkg/agent/clientset.go
```diff
@@ -171,17 +167,17 @@ func (cs *ClientSet) syncOnce() error {
...
 		return nil
 	}
-	c, err := cs.newAgentClient()
+	c, serverCount, err := cs.newAgentClient()
 	if err != nil {
 		return err
 	}
-	cs.serverCount = c.stream.serverCount
-	if err := cs.AddClient(c.stream.serverID, c); err != nil {
+	cs.serverCount = serverCount
```
cheftako, line 174: I think we need to log a warning if cs.serverCount !=0 and cs.serverCount != serverCount. Whether we are looking at a serverCount of zero creeping in, two proxy servers/apiservers with differing beliefs in the needed number of servers or a legitimate runtime change in the count, it will be hard to debug these scenarios without this information. In addition this should be a very rare event.

caesarxuchao: Good call. Updated.

PR #111, opened 2020-06-04
Fix retrieving caller info for grpc

Jefftree:


pkg/server/server.go
```diff
@@ -192,8 +192,8 @@ func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
 	if !ok {
 		return fmt.Errorf("failed to get context")
 	}
-	userAgent := md.Get(header.UserAgent)
-	klog.Infof("proxy request from client, userAgent %s", userAgent)
+	caller := md.Get("caller")
```
caesarxuchao, line 195: Why is this change?

Jefftree: GRPC User Agent is tied to a connection, and this was fine before because each proxy request involved creating a new connection. Now that we're using one connection, we need to differentiate each stream with a different caller.

Makefile
```diff
@@ -252,4 +252,4 @@ release-alias-tag: # Adds the tag to the last build tag. BASE_REF comes from the
 .PHONY: clean
 clean:
 	go clean -testcache
-	rm -rf proto/agent/agent.pb.go konnectivity-client/proto/client/client.pb.go easy-rsa.tar.gz easy-rsa-master cfssl cfssljson certs bin proto/agent/mocks
+	rm -rf bin
```
cheftako, line 255: Why are we no longer cleaning up generated files?

cmd/client/main.go
```diff
@@ -264,13 +264,19 @@ func (c *Client) getUDSDialer(o *GrpcProxyClientOptions) (func(ctx context.Conte
 			}
 			return c, err
 		})
-		tunnel, err := client.CreateSingleUseGrpcTunnel(o.proxyUdsName, dialOption, grpc.WithInsecure(), grpc.WithUserAgent(o.userAgent))
+		grpcConn, err := grpc.Dial(o.proxyUdsName, dialOption, grpc.WithInsecure())
+
```
cheftako, line 268: Can we move this blank line to before the Dial?

PR #112, opened 2020-06-09
Add verbosity levels and structured logs

### LINKED ISSUE #28: Use different verbose log level for different log
For example, trace log could be `klog.V(4)`.

mvladev:
Logs now are now use [logging conventions](https://github.com/kubernetes/community/blob/407895f4995d5e311a3fa371d59245c360f78c49/contributors/devel/sig-instrumentation/logging.md) and [structured logging conventions](https://github.com/kubernetes/community/blob/01165b1a4a37c5d38057c4eee0584e2c285d0a08/contributors/devel/sig-instrumentation/migration-to-structured-logging.md)

Fixes #28

### DIFF
--- a/cmd/client/main.go
+++ b/cmd/client/main.go
@@ -97,18 +97,18 @@ func (o *GrpcProxyClientOptions) Flags() *pflag.FlagSet {
 }
 
 func (o *GrpcProxyClientOptions) Print() {
-	klog.Warningf("ClientCert set to %q.\n", o.clientCert)
-	klog.Warningf("ClientKey set to %q.\n", o.clientKey)
-	klog.Warningf("CACert set to %q.\n", o.caCert)
-	klog.Warningf("RequestProto set to %q.\n", o.requestProto)
-	klog.Warningf("RequestPath set to %q.\n", o.requestPath)
-	klog.Warningf("RequestHost set to %q.\n", o.requestHost)
-	klog.Warningf("RequestPort set to %d.\n", o.requestPort)
-	klog.Warningf("ProxyHost set to %q.\n", o.proxyHost)
-	klog.Warningf("ProxyPort set to %d.\n", o.proxyPort)
-	klog.Warningf("ProxyUdsName set to %q.\n", o.proxyUdsName)
-	klog.Warningf("TestRequests set to %q.\n", o.testRequests)
-	klog.Warningf("TestDelaySec set to %d.\n", o.testDelaySec)
+	klog.V(1).Infof("ClientCert set to %q.\n", o.clientCert)
+	klog.V(1).Infof("ClientKey set to %q.\n", o.clientKey)
+	klog.V(1).Infof("CACert set to %q.\n", o.caCert)
+	klog.V(1).Infof("RequestProto set to %q.\n", o.requestProto)
+	klog.V(1).Infof("RequestPath set to %q.\n", o.requestPath)
+	klog.V(1).Infof("RequestHost set to %q.\n", o.requestHost)
+	klog.V(1).Infof("RequestPort set to %d.\n", o.requestPort)
+	klog.V(1).Infof("ProxyHost set to %q.\n", o.proxyHost)
+	klog.V(1).Infof("ProxyPort set to %d.\n", o.proxyPort)
+	klog.V(1).Infof("ProxyUdsName set to %q.\n", o.proxyUdsName)
+	klog.V(1).Infof("TestRequests set to %q.\n", o.testRequests)
+	klog.V(1).Infof("TestDelaySec set to %d.\n", o.testDelaySec)
 }
 
 func (o *GrpcProxyClientOptions) Validate() error {
@@ -244,7 +244,7 @@ func (c *Client) run(o *GrpcProxyClientOptions) error {
 		}
 
 		if i != o.testRequests {
-			klog.Warningf("Waiting for %d seconds.", o.testDelaySec)
+			klog.V(1).InfoS("Wating for next connection test.", "seconds", o.testDelaySec)
 			wait := time.Duration(o.testDelaySec) * time.Second
 			time.Sleep(wait)
 		}
@@ -269,7 +269,7 @@ func (c *Client) makeRequest(o *GrpcProxyClientOptions, client *http.Client) err
 	if err != nil {
 		return fmt.Errorf("failed to read response from client, got %v", err)
 	}
-	klog.Info(string(data))
+	klog.V(4).Infoln(string(data))
 	return nil
 }
 
@@ -292,7 +292,7 @@ func (c *Client) getUDSDialer(o *GrpcProxyClientOptions) (func(ctx context.Conte
 		<-ch
 		if proxyConn != nil {
 			err := proxyConn.Close()
-			klog.Infof("connection closed: %v", err)
+			klog.ErrorS(err, "connection closed")
 		}
 	}()
 
@@ -304,7 +304,7 @@ func (c *Client) getUDSDialer(o *GrpcProxyClientOptions) (func(ctx context.Conte
 			// timeout - is turned off as this is test code and eases debugging.
 			c, err := net.DialTimeout("unix", o.proxyUdsName, 0)
 			if err != nil {
-				klog.Errorf("failed to create connection to uds name %s, error: %v", o.proxyUdsName, err)
+				klog.ErrorS(err, "failed to create connection to uds", "name", o.proxyUdsName)
 			}
 			return c, err
 		})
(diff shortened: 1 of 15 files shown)


pkg/agent/client.go
```diff
@@ -238,13 +238,13 @@ func (a *AgentClient) Serve() {
...
 				if err := a.stream.Send(resp); err != nil {
-					klog.Warningf("close response send error: %v", err)
+					klog.V(2).Infof("close response send error: %v", err)
 					continue
 				}
 			}
 
 		default:
-			klog.Warningf("unrecognized packet type: %+v", pkt)
+			klog.V(2).Infof("unrecognized packet type: %+v", pkt)
```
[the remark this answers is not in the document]
mvladev: It depends in which contexts they are used. For example in the kubernetes main repo there are lots of those: https://github.com/kubernetes/kubernetes/blob/25651408aeadf38c3df7ea8c760e7519fd37d625/pkg/controller/deployment/deployment_controller.go#L246

konnectivity-client/pkg/client/client.go
```diff
@@ -88,11 +88,11 @@ func (t *grpcTunnel) serve(c *grpc.ClientConn) {
 			return
 		}
 		if err != nil || pkt == nil {
-			klog.Warningf("stream read error: %v", err)
+			klog.V(4).Infof("stream read error: %v", err)
 			return
 		}
 
-		klog.V(6).Infof("[tracing] recv packet, type: %s", pkt.Type)
+		klog.V(4).Infof("[tracing] recv packet, type: %s", pkt.Type)
```
cheftako, line 95: Unusual circumstances such as DialResp not recognized should be higher priority log level than something like book keeping that we got a packet. Guessing something like this should probably be a 6.

cheftako:
Thanks so much for this work. I think we should also differentiate expected (Eg we got a packet) from the unexpected (I do not recognize and there for am not sure if I am correctly handling packet X which I just received.)

vpnachev:
Is there any progress on this PR?


I just tried to run the proxy agent with logs to the stdout but it is dumping like crazy - 700 log messages per second that looks like
```
I0914 20:36:14.310336       1 client.go:384] received XXX bytes from remote for connID[YY]
```

The version is v0.0.12 and  flags that I used are:
```
- --logtostderr=true
- --stderrthreshold=1
```

mvladev:
@cheftako I've updated the PR now the following log levels are used:

- `0` - information for flag values.
- `1` - lifecycle warnings or errors when a server (e.g. agent server)
    is started / stopped.
- `2` - lifecycle information when a server (e.g. agent server) is
    started / stopped.
- `3` - agent <-> server registration warnings or errors.
- `4` - agent <-> server registration information.
- `5` - warnings or errors for agent <-> server requests.
- `6` - debugging agent <-> server requests.
- `7` - tracing requests.

cheftako:
> Walter Fender I've updated the PR now the following log levels are used:
> 
> * `0` - information for flag values.
> * `1` - lifecycle warnings or errors when a server (e.g. agent server)
>   is started / stopped.
> * `2` - lifecycle information when a server (e.g. agent server) is
>   started / stopped.
> * `3` - agent <-> server registration warnings or errors.
> * `4` - agent <-> server registration information.
> * `5` - warnings or errors for agent <-> server requests.
> * `6` - debugging agent <-> server requests.
> * `7` - tracing requests.

Still reviewing but I'm a big fan of what I have seen so far. I like the log level above. I actually think we should probably start a dev guide md at the root level and add guidance like this to it. That can be done it  a follow up PR but if you feel up to doing it here that would be great.

konnectivity-client/pkg/client/client.go
```diff
@@ -94,11 +94,11 @@ func (t *grpcTunnel) serve(c clientConn) {
 			return
 		}
 		if err != nil || pkt == nil {
-			klog.Warningf("stream read error: %v", err)
+			klog.V(5).Infof("stream read error: %v", err)
```
cheftako, line 97: Konnectivity-client is our client library, so its running in somewhere like the KAS. As such I think we should probably be using https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md for this package. As such this should be V2 or higher.

mvladev: I think that we should probably also go for https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#change-log-functions-to-structured-equivalent . I've changed some of the logging to use `InfoS`, but in this case as recommendation by this guide `klog.Warning{ln,f,Depth} -> klog.InfoS`

mvladev: What about enforcing the rules from https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md to the entire repo and not only to this specific package? This would simplify things a lot.

Jefftree: > What about enforcing the rules from https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md to the entire repo and not only to this specific package? This would simplify things a lot.

+1 for the simplification.

cheftako: +1 for simplification. For server/agent I think a rule of thumb like abnormal behavior should be logged at a higher level than expected behavior. So handshakes/connections/packets are less important/interesting than shutdown requests which are less important/interesting than unexpectedly broken connections.

trawler:
Hey @mvladev :) my team and I are very interested in this PR to be merged. Let us know if you need a hand!

mvladev:
Hi @trawler I'm almost done with the PR and I'll update it on Monday. Sorry for the delay.

mvladev:
I've updated the PR with structured logs. Logs now are now use [logging conventions](https://github.com/kubernetes/community/blob/407895f4995d5e311a3fa371d59245c360f78c49/contributors/devel/sig-instrumentation/logging.md) and [structured logging conventions](https://github.com/kubernetes/community/blob/01165b1a4a37c5d38057c4eee0584e2c285d0a08/contributors/devel/sig-instrumentation/migration-to-structured-logging.md).

cmd/client/main.go
```diff
@@ -244,7 +244,7 @@ func (c *Client) run(o *GrpcProxyClientOptions) error {
 		}
 
 		if i != o.testRequests {
-			klog.Warningf("Waiting for %d seconds.", o.testDelaySec)
+			klog.V(1).InfoS("Wating for next connection test.", "seconds", o.testDelaySec)
```
cheftako, line 247: typo: waiting

pkg/agent/client.go
```diff
@@ -248,36 +248,36 @@ func (a *AgentClient) initializeAuthContext(ctx context.Context) (context.Contex
...
+			klog.ErrorS(err, "could not read stream")
 			return
 		}
 
-		klog.Infof("[tracing] recv packet, type: %s", pkt.Type)
+		klog.V(5).InfoS("[tracing] recv packet", "type", pkt.Type)
 
 		if pkt == nil {
-			klog.Warningf("empty packet received")
+			klog.V(2).Infoln("empty packet received")
```
cheftako, line 274: Wondering if this should be V(3) or V(4). Not clear this corresponds to a "significant state change in the system". I don't think we're expecting empty packets but I could see us drowning under the log level, given a bug. If we make it at least V(3), then we can turn it off if we find an issue.

pkg/server/server.go
```diff
@@ -548,38 +549,38 @@ func (s *ProxyServer) serveRecvBackend(backend Backend, stream agent.AgentServic
...
+			klog.V(5).InfoS("Received data from agent", "bytes", len(resp.Data), "agentID", agentID, "connectionID", resp.ConnectID)
 			frontend, err := s.getFrontend(agentID, resp.ConnectID)
 			if err != nil {
-				klog.Warning(err)
+				klog.ErrorS(err, "could not get frontent client")
 				break
 			}
 			if err := frontend.send(pkt); err != nil {
-				klog.Warningf("<<< DATA send to client stream error: %v", err)
+				klog.ErrorS(err, "send to client stream failure")
```
cheftako, line 559: At some point we may want to log differently based on broken connection vs other errors.

cheftako:
Thanks for this work :+1: 
Couple of minor tweaks and happy to lgtm.

pkg/server/server.go
mvladev: Any suggestions? The fact that's an error should be taken into account.

mvladev:
Typos and empty packet log level comment has been fixed in 7ce64dc

pkg/server/server.go
cheftako: Its likely to be a nested error and I'm not sure what the root error is. This was more a note for the future than a suggestion for this PR. Sort of placeholder for future work :+1:

PR #115, opened 2020-06-13
add DNS:localhost to master certificate

charleszheng44:
When trying out the `python -m SimpleHTTPServer` example, got following error: 
```bash
Error: failed to get dialer for client, got failed to create tunnel localhost:8090, got rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: x509: certificate is valid for kubernetes, not localhost"
```
Add `DNS:localhost` to the proxy-server's certificate to fix it.


Jefftree:
In the example I think we used 127.0.0.1 which was included in the subject alt name. I guess localhost should be supported as well.

Please note that in kubernetes we use the same certificates generated for kube-apiserver so this code path only affects testing network proxy in isolation. 

/lgtm
/assign @cheftako

PR #118, opened 2020-06-30
Add metrics to the proxy agent

caesarxuchao:
/assign @Jefftree @cheftako @dberkov


pkg/agent/metrics/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+
+// Reset resets the metrics.
+func (a *AgentMetrics) Reset() {
+	a.failures.Reset()
+}
+
+// ObserveFailure records a failure to send to or receive from the proxy
+// server, labeled by the direction.
+func (a *AgentMetrics) ObserveFailure(direction string) {
+	a.failures.WithLabelValues(direction).Inc()
```
cheftako, line 68: Unbounded labels can be a significant issue for many metric systems. Can we type the string to an enumeration of values and use that as the parameter? It would prevent any worry about unbounded labels. It won't stop smeone who passes in a constant but it will at least catch attempts to pass in string variables. (https://play.golang.org/p/7-qnISOh0uS)

caesarxuchao: Updated. PTAL.

PR #119, opened 2020-07-01
Add agent dial latency metrics

caesarxuchao:


pkg/agent/client.go
```diff
@@ -294,6 +295,7 @@ func (a *AgentClient) Serve() {
 				}
 				continue
 			}
+			metrics.Metrics.ObserveDialLatency(time.Since(start))
```
cheftako, line 298: So this looks like it won't create a metric for an error, even a timeout. Is that desired?

pkg/agent/metrics/metrics.go
```diff
@@ -35,17 +37,31 @@ const (
 )
 
 var (
+	// Use buckets ranging from 5 ms to 12.5 seconds.
+	latencyBuckets = []float64{0.005, 0.025, 0.1, 0.5, 2.5, 12.5}
```
cheftako, line 41: I know we've had problems on the KAS with the top end bucket onlye being 8 seconds. A little worried that 12 seconds is still a little too short for the top bucket. What about something like {0.005, 0.025, 0.1, 0.5, 2.5, 10, 30}?

caesarxuchao: Done.

pkg/agent/client.go
caesarxuchao: That's intentional. I want to avoid dial timeout from bloating the average latency. Also this is consistent with the egressselector [metrics](https://github.com/kubernetes/kubernetes/blob/6079cebfae672335e958e03e7b0c258ba264d28f/staging/src/k8s.io/apiserver/pkg/server/egressselector/egress_selector.go#L225-L239) in k/k.

Also note that the metrics in k/k egressslector package covers the end-to-end dial error.

PR #120, opened 2020-07-01
Add proxy server metrics

caesarxuchao:


pkg/server/server.go
```diff
@@ -499,6 +503,7 @@ func (s *ProxyServer) serveRecvBackend(stream agent.AgentService_ConnectServer,
 				client.agentID = agentID
 				s.addFrontend(agentID, resp.ConnectID, client)
 				close(client.connected)
+				metrics.Metrics.ObserveDialLatency(time.Since(client.start))
```
cheftako, line 506: Looks like we fail to record latency on error. Is that desired? Also should be be recording a dial error metric?

pkg/server/server.go
```diff
@@ -24,12 +24,14 @@ import (
...
 	"strings"
 	"sync"
+	"time"
 
 	"google.golang.org/grpc/metadata"
 	authv1 "k8s.io/api/authentication/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/apiserver-network-proxy/konnectivity-client/proto/client"
+	"sigs.k8s.io/apiserver-network-proxy/pkg/agent/metrics"
```
Jefftree, line 34: server instead of agent metrics

pkg/server/server.go
Jefftree: +1. Similar to how dial failures are recorded in the egress_selector in k/k, it might be helpful to also record dial failures here

pkg/server/tunnel.go
```diff
@@ -72,6 +73,7 @@ func (t *Tunnel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		Mode:      "http-connect",
 		HTTP:      conn,
 		connected: connected,
+		start:     time.Now(),
```
Jefftree, line 76: nit: I'm sure the difference is negligible, but for consistency purposes can we arrange the dial flow in the same order between http connect and grpc? (Seems like in http-connect we do the backend selection and package sending after recording the start time, but in grpc those two steps are done before).

pkg/server/server.go
caesarxuchao: Recording dial failures here does not add much to the existing k/k dial error metrics. If we want to know how many dial has failed, we can check the k/k metrics. If we want to know the reason of dial failure, we can check the error message in k/k.

Recording the latency is different, it helps break down the dial latency. Though in my test it show the dial latency isn't a performance bottleneck.

caesarxuchao: Not recording latency on error is intentional. This is to avoid failures like dial timeout from bloating the average latency. Also this is consistent with the egressselector [metrics](https://github.com/kubernetes/kubernetes/blob/6079cebfae672335e958e03e7b0c258ba264d28f/staging/src/k8s.io/apiserver/pkg/server/egressselector/egress_selector.go#L225-L239) in k/k.

pkg/server/tunnel.go
caesarxuchao: Good point. I rearranged the order so that in both cases the latency includes the time of sending the dial_req.

This also fixed a memory leak in the tunnel scenario. Previously even if there is no backend available, a pending dial is added to server's pendingDialManager, but it will never be removed from the manager since the request will not even been sent out.

Issue #69, opened 2020-02-18
Clean up UDS file before listening


caesarxuchao:
When the proxy server restarts, if the UDS file already exists, trying to listen on the socket will fail with "bind: address already in use". The server should delete the file if it exists, and then listen to the socket.

cc @Jefftree

Jefftree:
When does this happen? The socket binding should be automatically removed if the proxy server is shut down. I'm hesitant to just delete to file since it's possible that things were misconfigured, and we're forcefully removing a random file that the user specified.

cheftako:
I believe that at least one of the places (for the http-connect case) which cleans up the UDS file is https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/cmd/proxy/main.go#L395 (defer func() {udsListener.Close()}()). For the grpc case I would guess its the GracefulStop result but calling that method seems less bullet proof.

caesarxuchao:
If the proxy server crashes, then no one is going to cleanup the uds file. This happened to me today.

As a side note, we need to check udsListener is nil before call udsListener.Close().

Jefftree:
I encountered this quite often while simulating error conditions. Will send out a fix

/assign

PR #121, opened 2020-07-02
Add an option to delete existing UDS file before listening on it 

caesarxuchao:
Fix https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/69.

If the server is deployed as a static pod in the Kubernetes master, you can reproduce the error in #69 just by modifying the /etc/kubernetes/manifests/konnectivity-server.yaml to trigger a restart.

This patch deletes the existing UDS file before listening on it. Deleting a file is not safe, especially if the user misconfigured UDS file name, so I made this behavior optional.

If the proxy server is deployed in a managed environment like GKE, it's safe to enable this behavior, because it's unlikely to set a wrong UDS file name. Also, only privileged user can edit /etc/kubernetes/manifests/konnectivity-server.yaml, so enabling this behavior does not increase attacking surface.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -79,6 +79,8 @@ type ProxyRunOptions struct {
 	mode string
 	// Location for use by the "unix" network. Setting enables UDS for server connections.
 	udsName string
+	// If file udsName already exists, delete the file before listen on that UDS file.
+	deleteUDSFile bool
 	// Port we listen for server connections on.
 	serverPort uint
 	// Port we listen for agent connections on.
@@ -117,6 +119,7 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.clusterCaCert, "cluster-ca-cert", o.clusterCaCert, "If non-empty the CA we use to validate Agent clients.")
 	flags.StringVar(&o.mode, "mode", o.mode, "Mode can be either 'grpc' or 'http-connect'.")
 	flags.StringVar(&o.udsName, "uds-name", o.udsName, "uds-name should be empty for TCP traffic. For UDS set to its name.")
+	flags.BoolVar(&o.deleteUDSFile, "delete-existing-uds-file", o.deleteUDSFile, "If true and if file udsName already exists, delete the file before listen on that UDS file")
 	flags.UintVar(&o.serverPort, "server-port", o.serverPort, "Port we listen for server connections on. Set to 0 for UDS.")
 	flags.UintVar(&o.agentPort, "agent-port", o.agentPort, "Port we listen for agent connections on.")
 	flags.UintVar(&o.adminPort, "admin-port", o.adminPort, "Port we listen for admin connections on.")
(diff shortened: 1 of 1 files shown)


cmd/server/main.go
```diff
@@ -267,6 +271,7 @@ func newProxyRunOptions() *ProxyRunOptions {
 		clusterCaCert:          "",
 		mode:                   "grpc",
 		udsName:                "",
+		deleteUDSFile:          false,
```
cheftako, line 274: Do we want to default to true? I realize this is a change in behavior; so we need to consider backward compatibility. I believe the server won't some up if the file already exists and we are still pretty early in our adoption.

caesarxuchao: If this flag is default to true, misconfiguring UDSName become more detrimental.

> I believe the server won't some up if the file already exists
 
That's right. I think it's better to let user find that the server doesn't come up and then manually set this flag to true, than accidentally deleting files and then disable this flag.

cmd/server/main.go
```diff
@@ -391,6 +396,11 @@ func (p *Proxy) runUDSMasterServer(ctx context.Context, o *ProxyRunOptions, s *s
 	if o.mode == "grpc" {
 		grpcServer := grpc.NewServer()
 		client.RegisterProxyServiceServer(grpcServer, s)
+		if o.deleteUDSFile {
```
Jefftree, line 399: I remember this PR was originally created to address an improper uds socket cleanup in grpc mode, but given the description of this flag shouldn't we also perform the cleanup in http-connect mode?

caesarxuchao: Good catch. Updated.

cmd/server/main.go
cheftako: Alright I'll buy that.

cmd/server/main.go
```diff
@@ -387,6 +392,11 @@ func (p *Proxy) runMasterServer(ctx context.Context, o *ProxyRunOptions, server
 }
 
 func (p *Proxy) runUDSMasterServer(ctx context.Context, o *ProxyRunOptions, s *server.ProxyServer) (StopFunc, error) {
+	if o.deleteUDSFile {
+		if err := os.Remove(o.udsName); err != nil {
```
cheftako, line 396: I believe this will error if the file does not exist. (Eg. first time run). It seems like we should only do this if the file exists. Given that maybe check if the fix exists.  If it does and deleteUDSFile is set then delete, if deleteUDSFile is not set give a detailed error message of the problem.

caesarxuchao: I updated the code to ignore the error if it's a "file not exist" error to avoid printing the redundant warning message.

>  if deleteUDSFile is not set give a detailed error message of the problem

Current error message is: "failed to run the master server: failed to listen(unix) name ./<udsName>: listen unix ./<udsName>: bind: address already in use". I think it's clear enough.

Jefftree:
lgtm

cmd/server/main.go
```diff
@@ -110,6 +112,7 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.clusterCaCert, "cluster-ca-cert", o.clusterCaCert, "If non-empty the CA we use to validate Agent clients.")
 	flags.StringVar(&o.mode, "mode", o.mode, "Mode can be either 'grpc' or 'http-connect'.")
 	flags.StringVar(&o.udsName, "uds-name", o.udsName, "uds-name should be empty for TCP traffic. For UDS set to its name.")
+	flags.BoolVar(&o.deleteUDSFile, "delete-existing-uds-file", o.deleteUDSFile, "If true, if file udsName already exists, delete the file before listen on that UDS file")
```
cheftako, line 115: If true and file

caesarxuchao: Updated.

PR #122, opened 2020-07-07
Enable pprof in proxy server

caesarxuchao:
Useful to debug memory leaks, etc.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -25,8 +25,10 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	"net/http/pprof"
 	"os"
 	"os/signal"
+	"runtime"
 	"syscall"
 
 	"github.com/google/uuid"
@@ -85,6 +87,11 @@ type ProxyRunOptions struct {
 	adminPort uint
 	// Port we listen for health connections on.
 	healthPort uint
+	// Enables pprof at host:adminPort/debug/pprof.
+	enableProfiling bool
+	// If enableProfiling is true, this enables the lock contention
+	// profiling at host:adminPort/debug/pprof/block.
+	enableContentionProfiling bool
 
 	// ID of this proxy server.
 	serverID string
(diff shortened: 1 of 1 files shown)


cmd/server/main.go
```diff
@@ -519,6 +527,11 @@ func (p *Proxy) runAgentServer(o *ProxyRunOptions, server *server.ProxyServer) e
 func (p *Proxy) runAdminServer(o *ProxyRunOptions, server *server.ProxyServer) error {
 	muxHandler := http.NewServeMux()
 	muxHandler.Handle("/metrics", promhttp.Handler())
+	muxHandler.HandleFunc("/debug/pprof", pprof.Index)
```
cheftako, line 530: This is enabling a lot more than just contention profiling indicated by the flag.

```
Count | Profile
-- | --
0 | allocs
0 | block
0 | cmdline
12 | goroutine
0 | heap
0 | mutex
0 | profile
13 | threadcreate
0 | trace
```

cmd/server/main.go
```diff
@@ -114,6 +119,7 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.UintVar(&o.agentPort, "agent-port", o.agentPort, "Port we listen for agent connections on.")
 	flags.UintVar(&o.adminPort, "admin-port", o.adminPort, "Port we listen for admin connections on.")
 	flags.UintVar(&o.healthPort, "health-port", o.healthPort, "Port we listen for health connections on.")
+	flags.BoolVar(&o.enableContentionProfiling, "enable-contention-profiling", o.enableContentionProfiling, "enable contention profiling at host:admin-port/debug/pprof/debug")
```
cheftako, line 122: For me the pprof is at http://host:admin-port/debug/pprof

caesarxuchao: Weird. For me it's debug/pprof/block, it's also the address in the [doc](https://golang.org/pkg/net/http/pprof/). debug/block returns 404 for me. Can you double check?

caesarxuchao: Oh I see /debug/pprof/debug is a typo. I fixed it.

cmd/server/main.go
caesarxuchao: Good point. Added --enable-profiling.

caesarxuchao:
@cheftako PTAL. Thanks.

cmd/server/main.go
cheftako: Looks like we both typo'd that one :smile_cat:

cmd/server/main.go
```diff
@@ -519,6 +535,13 @@ func (p *Proxy) runAgentServer(o *ProxyRunOptions, server *server.ProxyServer) e
 func (p *Proxy) runAdminServer(o *ProxyRunOptions, server *server.ProxyServer) error {
 	muxHandler := http.NewServeMux()
 	muxHandler.Handle("/metrics", promhttp.Handler())
+	if o.enableProfiling {
+		muxHandler.HandleFunc("/debug/pprof", pprof.Index)
```
cheftako, line 539: Not a blocker but I noticed that the links generated on the pprof index page when you hit /debug/pprof are for /debug/X not for /debug/pprof/X. So you get 404's if you click any of the links. The links are fine if you hit /debug/pprof/. You might consider using a redirect on "/debug/pprof" to send them to the good url.

func indexRedirect(w http.ResponseWriter, r *http.Request) {
	http.Redirect(w, r, "/debug/pprof/", http.StatusMovedPermanently)
}

caesarxuchao: Done.  PTAL. Thanks.

PR #124, opened 2020-07-13
Added controllable test http server.

cheftako:
Added controllable test http server.
Currently only support /success for baic page returned.
Plan to add various error producing pages as well.
Test server supports connection reuse which the old server did not.
Added option to test client to test connection reuse.
Updated documentation for detail on using UDS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,10 +23,12 @@ STAGING_REGISTRY := gcr.io/k8s-staging-kas-network-proxy
 SERVER_IMAGE_NAME ?= proxy-server
 AGENT_IMAGE_NAME ?= proxy-agent
 TEST_CLIENT_IMAGE_NAME ?= proxy-test-client
+TEST_SERVER_IMAGE_NAME ?= http-test-server
 
 SERVER_FULL_IMAGE ?= $(REGISTRY)/$(SERVER_IMAGE_NAME)
 AGENT_FULL_IMAGE ?= $(REGISTRY)/$(AGENT_IMAGE_NAME)
 TEST_CLIENT_FULL_IMAGE ?= $(REGISTRY)/$(TEST_CLIENT_IMAGE_NAME)
+TEST_SERVER_FULL_IMAGE ?= $(REGISTRY)/$(TEST_SERVER_IMAGE_NAME)
 
 TAG ?= $(shell git rev-parse HEAD)
 
@@ -54,13 +56,16 @@ bin:
 	mkdir -p bin
 
 .PHONY: build
-build: bin/proxy-agent bin/proxy-server bin/proxy-test-client
+build: bin/proxy-agent bin/proxy-server bin/proxy-test-client bin/http-test-server
 
 bin/proxy-agent: proto/agent/agent.pb.go konnectivity-client/proto/client/client.pb.go bin cmd/agent/main.go
 	GO111MODULE=on go build -o bin/proxy-agent cmd/agent/main.go
 
 bin/proxy-test-client: konnectivity-client/proto/client/client.pb.go bin cmd/client/main.go
 	GO111MODULE=on go build -o bin/proxy-test-client cmd/client/main.go
+
+bin/http-test-server: bin cmd/test-server/main.go
+	GO111MODULE=on go build -o bin/http-test-server cmd/test-server/main.go
 
 bin/proxy-server: proto/agent/agent.pb.go konnectivity-client/proto/client/client.pb.go bin cmd/server/main.go
 	GO111MODULE=on go build -o bin/proxy-server cmd/server/main.go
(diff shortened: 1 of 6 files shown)


cmd/client/main.go
```diff
@@ -209,7 +223,25 @@ func (c *Client) run(o *GrpcProxyClientOptions) error {
 	client := &http.Client{
 		Transport: transport,
 	}
-	requestURL := fmt.Sprintf("%s://%s:%d%s", o.requestProto, o.requestHost, o.requestPort, o.requestPath)
+
+	for i := 1; i <= o.testRequests; i++ {
+		err = c.makeRequest(o, client)
```
Jefftree, line 228: We don't support reusing the same dialer for multiple requests. You'll need to obtain the dialer again for each request.

cheftako: I think this is a slightly tricky discussion. We have several separate layers here and we may be confusing them in this discussion. We have the outer tunnel (actually 2 separate tunnels, 1 from Client to Server, another from Server to Agent). We also a TCP connection on top of that tunnel (From  test client to test server) which is essentially a HTTP Session. On top of that we have individual HTTP requests. This is specifically HTTP 1.1 with session reuse. As long as we don't break the TCP connection (which is to say the HTTP session), then additional requests are just obfuscated traffic on top of that connection and the APIServer Network Proxy shouldn't care. That only works as long as we don't close the connection. There is a Close() in this method but its closing the response body (https://golang.org/src/net/http/transfer.go?s=97066:97143#L975) and not the connection, so everything should be fine. Its important that this work as things like exec and logs work on long lived HTTP sessions, so we need to handle it properly. If you try this test you will see that you can make multiple requests and everything works as longs as you don't kill the agent. In that case as you noted we should probably fail gracefully. Right now we just seem to hang.

Jefftree: Thanks for the explanation. I should rephrase my original comment as "we do not support reusing the same dialer for multiple **connections**". With HTTP 1.1 Connection reuse, network proxy should indeed support subsequent requests since the original connection has not been closed. One thing I'd like to note is that golang's default `IdleConnTimeout` is 90 seconds (https://golang.org/src/net/http/transport.go?s=13608:13676) and k/k abides by this default. Thus, when running the test-client with a delay parameter greater than 90s (even when no agent restart is done) `--test-requests=2 --test-delay=100` the second request will fail 100% of the time because currently network proxy explicitly closes the tunnel when a close request is sent to to the inner TCP connection. I'm not sure if this is the desired behavior since I imagine we may have some cases of this kind of reuse in k/k. At the same time, there is no way for us to know whether a tunnel will be reused so to prevent leaking too many tunnel connections, we explicitly close the tunnel on the first CLOSE_RSP we obtain from the inner TCP connection (https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/konnectivity-client/pkg/client/client.go#L137).

As for the issue with the proxy-client hanging on agent restarts, that's a separate issue and I can submit a fix for that.

cheftako: Good point.

Added comment to this effect to the test-client code.

cmd/test-server/main.go
```diff
@@ -0,0 +1,244 @@
...
+		close(stop)
+		<-c
+		os.Exit(1) // second signal. Exit directly.
+	}()
+
+	return stop
+}
+
+func returnSuccess(w http.ResponseWriter, req *http.Request) {
+	fmt.Fprintf(w, "<!DOCTYPE html>\n<html>\n    <head>\n        <title>Success</title>\n    </head>\n    <body>\n        <p>The sucess test page!</p>\n    </body>\n</html>")
```
Jefftree, line 182: nit: s/sucess/success

cmd/test-server/main.go
```diff
@@ -0,0 +1,244 @@
...
+	})
+	if err := command.Execute(); err != nil {
+		klog.Errorf("error: %v\n", err)
+		klog.Flush()
+		os.Exit(1)
+	}
+}
+
+type TestServerRunOptions struct {
+	// Certificate setup for securing communication to the "client" i.e. the Kube API Server.
```
Jefftree, line 56: Are these variables used anywhere? Should we just drop them for now since they're unused?

Makefile
```diff
@@ -178,18 +183,30 @@ docker-push/proxy-test-client: docker-build/proxy-test-client
...
 ## Docker — All ARCH
 ## --------------------------------------
 
 .PHONY: docker-build-all
-docker-build-all: $(addprefix docker-build/proxy-agent-,$(ALL_ARCH)) $(addprefix docker-build/proxy-server-,$(ALL_ARCH)) $(addprefix docker-build/proxy-test-client-,$(ALL_ARCH))
+docker-build-all: $(addprefix docker-build/proxy-agent-,$(ALL_ARCH)) $(addprefix docker-build/proxy-server-,$(ALL_ARCH)) $(addprefix docker-build/proxy-test-client-,$(ALL_ARCH)) $(addprefix docker-build/http-test-server-,$(ALL_ARCH))
 
 .PHONY: docker-push-all
-docker-push-all: $(addprefix docker-push/proxy-agent-,$(ALL_ARCH)) $(addprefix docker-push/proxy-server-,$(ALL_ARCH)) $(addprefix docker-push/proxy-test-client-,$(ALL_ARCH))
+docker-push-all: $(addprefix docker-push/proxy-agent-,$(ALL_ARCH)) $(addprefix docker-push/proxy-server-,$(ALL_ARCH)) $(addprefix docker-push/proxy-test-client-,$(ALL_ARCH)) $(addprefix docker-push/http-test-server-,$(ALL_ARCH))
```
Jefftree, line 205: This was already in place before your PR, but do you know why we push the test client and http test server? k/k only requires the images for proxy server and agent and I believe the other images are not used.

cmd/test-server/main.go
cheftako: No good point. I'll drop them for now. At some point we may want to test things about https inside the tunnel, but that day, is not today.

cmd/test-server/main.go
cheftako: Fixed

Makefile
cheftako: Lets leave that as a follow up. Not sure 100% sure. It *may* be we want them for future test automation but I'm really not sure.

Issue #123, opened 2020-07-10
DATA to Backend failed, after agent restart


Avatat:
Hello,
In the beginning, I would like to thank you, apiserver-network-proxy team, for your amazing work :)

I use Konnectivity in version 0.0.10 and Kubernetes 1.18.5. Most of the time everything works flawlessly, but when I restart agent's pods, konnectivity-server don't want to switch traffic to the new connections:
```
I0710 13:11:42.881321       1 server.go:293] >>> DATA sent to backend
I0710 13:11:42.952402       1 server.go:498] <<< Received 372 bytes of DATA from agentID d3c1f7ef-4730-4300-964e-2e382af9e484, connID 1
I0710 13:11:42.961352       1 server.go:278] >>> Received 42 bytes of DATA(id=1)
I0710 13:11:42.961521       1 server.go:293] >>> DATA sent to backend
I0710 13:11:47.641175       1 server.go:418] Connect request from agent 3adda35e-94ab-46ae-bd6c-96a8197ff651
I0710 13:11:47.641222       1 backend_manager.go:99] register Backend &{0xc00050e180} for agentID 3adda35e-94ab-46ae-bd6c-96a8197ff651
W0710 13:11:47.909837       1 server.go:451] stream read error: rpc error: code = Canceled desc = context canceled
I0710 13:11:47.909895       1 backend_manager.go:119] remove Backend &{0xc000222480} for agentID 1bf0e057-b131-45b9-bbcb-ab752e9dfef4
I0710 13:11:47.909940       1 server.go:531] <<< Close backend &{0xc000222480} of agent 1bf0e057-b131-45b9-bbcb-ab752e9dfef4
W0710 13:11:51.898482       1 server.go:451] stream read error: rpc error: code = Canceled desc = context canceled
I0710 13:11:51.898535       1 backend_manager.go:119] remove Backend &{0xc0002223c0} for agentID d3c1f7ef-4730-4300-964e-2e382af9e484
I0710 13:11:51.898592       1 server.go:531] <<< Close backend &{0xc0002223c0} of agent d3c1f7ef-4730-4300-964e-2e382af9e484
I0710 13:11:55.825758       1 server.go:278] >>> Received 53 bytes of DATA(id=3)
W0710 13:11:55.825884       1 server.go:291] >>> DATA to Backend failed: rpc error: code = Unavailable desc = transport is closing
I0710 13:11:55.825914       1 server.go:293] >>> DATA sent to backend
[...]

PR #125, opened 2020-07-15
Disconnect frontend if backend drops

Jefftree:
Created to address https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/123

The proxy server generates a mapping between a frontend (kube-apiserver client connection) and a backend (particular agentID connection) whenever a connection is established. When the agent connection is broken and a reconnect occurs, this backend still points to the old agent connection which is broken. This is the expected behavior ever since we removed reconnects. However, the frontend has no awareness that the backend is gone and will keep sending packets.

This PR disconnects the frontend connection (more technically sends a close rsp to the frontend to close the connection) if its corresponding agent connection is dropped. This prevents the client from getting stuck (https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/123#issuecomment-657328196) and immediately surfaces an error.

This is similar to how we send a close_req to the backend to terminate the agent -> destination connection when the frontend disconnects (https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/server/server.go#L312-L330)

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -183,6 +183,16 @@ func (s *ProxyServer) getFrontend(agentID string, connID int64) (*ProxyClientCon
 	return conn, nil
 }
 
+func (s *ProxyServer) getFrontends(agentID string) (map[int64]*ProxyClientConnection, error) {
+	s.fmu.RLock()
+	defer s.fmu.RUnlock()
+	conns, ok := s.frontends[agentID]
+	if !ok {
+		return nil, fmt.Errorf("can't find agentID %s in the frontends", agentID)
+	}
+	return conns, nil
+}
+
 // NewProxyServer creates a new ProxyServer instance
 func NewProxyServer(serverID string, serverCount int, agentAuthenticationOptions *AgentTokenAuthenticationOptions) *ProxyServer {
 	bm := NewDefaultBackendManager()
@@ -301,8 +311,9 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer, 
 			if err := backend.Send(pkt); err != nil {
 				// TODO: retry with other backends connecting to this agent.
 				klog.Warningf(">>> DATA to Backend failed: %v", err)
+				continue
 			}
-			klog.Info(">>> DATA sent to backend")
+			klog.Info(">>> DATA sent to Backend")
 
 		default:
 			klog.Infof(">>> Ignore %v packet coming from frontend", pkt.Type)
@@ -480,7 +491,7 @@ func (s *ProxyServer) serveRecvBackend(stream agent.AgentService_ConnectServer, 
 			resp := pkt.GetDialResponse()
 			klog.Infof("<<< Received DIAL_RSP(rand=%d), agentID %s, connID %d)", resp.Random, agentID, resp.ConnectID)
 
-			if client, ok := s.PendingDial.Get(resp.Random); !ok {
+			if frontend, ok := s.PendingDial.Get(resp.Random); !ok {
 				klog.Warning("<<< DialResp not recognized; dropped")
 			} else {
 				dialErr := false
@@ -488,33 +499,32 @@ func (s *ProxyServer) serveRecvBackend(stream agent.AgentService_ConnectServer, 
 					klog.Warningf("<<< DIAL_RSP contains error: %v", resp.Error)
 					dialErr = true
 				}
-				err := client.send(pkt)
+				err := frontend.send(pkt)
 				s.PendingDial.Remove(resp.Random)
 				if err != nil {
-					klog.Warningf("<<< DIAL_RSP send to client stream error: %v", err)
+					klog.Warningf("<<< DIAL_RSP send to frontend stream error: %v", err)
 					dialErr = true
 				}
 				// Avoid adding the frontend if there was an error dialing the destination
 				if dialErr == true {
 					break
 				}
-
-				client.connectID = resp.ConnectID
-				client.agentID = agentID
-				s.addFrontend(agentID, resp.ConnectID, client)
-				close(client.connected)
-				metrics.Metrics.ObserveDialLatency(time.Since(client.start))
+				frontend.connectID = resp.ConnectID
+				frontend.agentID = agentID
+				s.addFrontend(agentID, resp.ConnectID, frontend)
+				close(frontend.connected)
+				metrics.Metrics.ObserveDialLatency(time.Since(frontend.start))
 			}
 
 		case client.PacketType_DATA:
 			resp := pkt.GetData()
 			klog.Infof("<<< Received %d bytes of DATA from agentID %s, connID %d", len(resp.Data), agentID, resp.ConnectID)
-			client, err := s.getFrontend(agentID, resp.ConnectID)
+			frontend, err := s.getFrontend(agentID, resp.ConnectID)
 			if err != nil {
 				klog.Warning(err)
 				break
 			}
-			if err := client.send(pkt); err != nil {
+			if err := frontend.send(pkt); err != nil {
 				klog.Warningf("<<< DATA send to client stream error: %v", err)
 			} else {
 				klog.V(6).Infof("<<< DATA sent to frontend")
@@ -523,12 +533,12 @@ func (s *ProxyServer) serveRecvBackend(stream agent.AgentService_ConnectServer, 
 		case client.PacketType_CLOSE_RSP:
 			resp := pkt.GetCloseResponse()
 			klog.Infof("<<< Received CLOSE_RSP(id=%d)", resp.ConnectID)
-			client, err := s.getFrontend(agentID, resp.ConnectID)
+			frontend, err := s.getFrontend(agentID, resp.ConnectID)
 			if err != nil {
 				klog.Warning(err)
 				break
 			}
-			if err := client.send(pkt); err != nil {
+			if err := frontend.send(pkt); err != nil {
 				// Normal when frontend closes it.
 				klog.Infof("<<< CLOSE_RSP send to client stream error: %v", err)
 			} else {
@@ -542,4 +552,25 @@ func (s *ProxyServer) serveRecvBackend(stream agent.AgentService_ConnectServer, 
 		}
 	}
 	klog.Infof("<<< Close backend %v of agent %s", stream, agentID)
+
+	// Close all connected frontends when the agent connection is closed
+	// TODO(#126): Frontends in PendingDial state that have not been added to the
+	//             list of frontends should also be closed.
+	frontends, _ := s.getFrontends(agentID)
+	klog.Infof("<<< Close %d frontends connected to agent %s", len(frontends), agentID)
+
+	for _, frontend := range frontends {
+		s.removeFrontend(agentID, frontend.connectID)
+		pkt := &client.Packet{
+			Type: client.PacketType_CLOSE_RSP,
+			Payload: &client.Packet_CloseResponse{
+				CloseResponse: &client.CloseResponse{},
+			},
+		}
+		pkt.GetCloseResponse().ConnectID = frontend.connectID
+		if err := frontend.send(pkt); err != nil {
+			klog.Warningf("<<< CLOSE_RSP to frontend failed: %v", err)
+		}
+	}
+
 }
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -485,6 +485,28 @@ func (s *ProxyServer) Connect(stream agent.AgentService_ConnectServer) error {
 
 // route the packet back to the correct client
 func (s *ProxyServer) serveRecvBackend(stream agent.AgentService_ConnectServer, agentID string, recvCh <-chan *client.Packet) {
+	defer func() {
+		// Close all connected frontends when the agent connection is closed
+		// TODO(#126): Frontends in PendingDial state that have not been added to the
+		//             list of frontends should also be closed.
+		frontends, _ := s.getFrontends(agentID)
+		klog.Infof("<<< Close %d frontends connected to agent %s", len(frontends), agentID)
+
+		for _, frontend := range frontends {
+			s.removeFrontend(agentID, frontend.connectID)
+			pkt := &client.Packet{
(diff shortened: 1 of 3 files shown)


Avatat:
@Jefftree, wow, it works!
I took your change, built Docker image from it and tested - everything works as it should :)

pkg/server/backend_manager.go
```diff
@@ -38,8 +39,13 @@ var _Backend = new(agent.AgentService_ConnectServer)
 type backend struct {
 	// TODO: this is a multi-writer single-reader pattern, it's tricky to
 	// write it using channel. Let's worry about performance later.
-	mu   sync.Mutex // mu protects conn
-	conn agent.AgentService_ConnectServer
+	mu      sync.Mutex // mu protects conn
+	conn    agent.AgentService_ConnectServer
+	AgentID string
```
cheftako, line 44: As far as I can tell, thus does not need to be public. Can we switch it to "agentID"?

pkg/server/server.go
```diff
@@ -541,5 +552,21 @@ func (s *ProxyServer) serveRecvBackend(stream agent.AgentService_ConnectServer,
 			klog.Warningf("<<< Unrecognized packet %+v", pkt)
 		}
 	}
+
 	klog.Infof("<<< Close backend %v of agent %s", stream, agentID)
+
+	frontends, _ := s.getFrontends(agentID)
```
cheftako, line 558: This closes a huge hole which is awesome. I believe however there is still a much smaller hole in pending dial. Most of the time I would expect pending dial to be empty. However if there is something in there, there is a chance its request went out via this backend. If so we will never get the response and that also needs to be dealt with. 

The issue is that we do not record in the pending data structure which backend it used, so we cannot tell if anything on the pending list would be effected by a given backend breaking. We also need to work out how to deal with it. One option would be to just fail, which is probably easiest. However as the connection has not yet be established, we should be able to switch to using a different backend.

Jefftree: The worst case scenario for PendingDial is for the client to hang for a bit and hit the 30s dial timeout that we specify in our konnectivity client. I agree this is a much smaller issue since it won't cause any indefinite hanging behavior, and I think we should address this in a separate PR. 

Created https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/126

cheftako: Might be good to a TODO as a placeholder with the bug id.

Jefftree: Good idea, added.

pkg/server/backend_manager.go
Jefftree: This was originally intended to be for debugging and I didn't find an easy way to incorporate this without regenerating the proto. Will remove for now.

caesarxuchao:
Does this cover the http-connect mode?

Jefftree:
The fix in the code covers http-connect mode as well but the added tests currently only cover GRPC.

tests/agent_disconnect_test.go
```diff
@@ -0,0 +1,151 @@
...
+				t.Error("expect second request on broken agent connection to fail")
+			}
+		}
+
+	}
+
+	verify(true)
+	close(stopCh)
+	// It takes up to 5 seconds for the agent to recognize the channel has been closed and shut down all connections.
+	time.Sleep(7 * time.Second)
```
caesarxuchao, line 136: This looks fragile. How about modify this to a waitUntil?

Also why would it take up to 5 seconds? I expect it to be <1s.

tests/agent_disconnect_test.go
```diff
@@ -0,0 +1,151 @@
...
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"google.golang.org/grpc"
+	"sigs.k8s.io/apiserver-network-proxy/konnectivity-client/pkg/client"
+)
+
+func TestProxy_Agent_Disconnect_Same_Stream(t *testing.T) {
```
caesarxuchao, line 14: I think the second test covers the first test. If so, let's remove this test.

tests/agent_disconnect_test.go
```diff
@@ -0,0 +1,151 @@
...
+	stopCh2 := make(chan struct{})
+	runAgent(proxy.agent, stopCh2)
+	defer close(stopCh2)
+
+	// Wait for agent to register on proxy server
+	time.Sleep(time.Second)
+
+	// Proxy requests should work again after agent reconnects
+	verify(true)
+}
```
caesarxuchao, line 151: Can we add a test for http-connect?

pkg/server/server.go
caesarxuchao: Agree that the 30s dial timeout makes pending dial less detrimental.

tests/agent_disconnect_test.go
Jefftree: The second test actually tests the disconnect scenario with 2 requests using 2 different tunnels. This first test uses HTTP 1.1 persistent connections to send 2 requests on the same tunnel. k/k has some commands using the latter scenario so I think it would be beneficial to keep both cases.

I will reword this test case a bit more accurately.

tests/agent_disconnect_test.go
Jefftree: Thanks for the catch, yeah it should be less than 1s. I tested this multiple times with 1s and it passes without any problems. I think it should be fine given that we also do `time.Sleep(time.Second)` in other places of our test code.

pkg/server/server.go
```diff
@@ -542,4 +552,25 @@ func (s *ProxyServer) serveRecvBackend(stream agent.AgentService_ConnectServer,
 		}
 	}
 	klog.Infof("<<< Close backend %v of agent %s", stream, agentID)
+
+	// Close all connected frontends when the agent connection is closed
+	// TODO(#126): Frontends in PendingDial state that have not been added to the
+	//             list of frontends should also be closed.
+	frontends, _ := s.getFrontends(agentID)
```
cheftako, line 559: This clean up block should really be turned into a function and called via a defer at the beginning of serveRecvFrontend.

tests/agent_disconnect_test.go
cheftako: If verify is idempotent we could consider using https://github.com/kubernetes/apimachinery/blob/master/pkg/util/wait/wait.go#L420 or waitUntil. It will allow us to test quickly but be resilient to problems with the test running slowly. We already have a dependency upon apimachinery, so I don't expect that to be a problem.

tests/agent_disconnect_test.go
cheftako: +1

Jefftree: Added

pkg/server/server.go
Jefftree: I wrapped it in an inline function and put it in a defer

tests/agent_disconnect_test.go
Jefftree: verify is idempotent in the sense that the output is the same, but has side effects (eg: golang caching http 1.1 persistent connections) that I'd like to avoid. If we want to poll, I think it should be done on the proxy server to determine the status of its ongoing connections. Currently this kind of function doesn't exist, but I think it would be useful to have for debugging purposes.

cheftako: I'm mostly thinking of it as way to early out in a test when a condition is met. So we can check a condition every second until it passes or some time out (30 seconds). Most of the time the test will run very quickly but when it does slow down it is less likely to become a flaky test.

caesarxuchao: I'm more concerned about the test becoming flaky. I hope one day we can get rid of the 1s sleep as well. If a waitUntil can work here, let's avoid introducing more latency sensitive code.

RE. "golang caching http 1.1 persistent connections", sorry I didn't get why you want to avoid that in this test?

tests/agent_disconnect_test.go
```diff
@@ -0,0 +1,229 @@
...
+
+			if string(data) != "hello" {
+				t.Errorf("expected test, got %v", data)
+			}
+		} else {
+			if err == nil {
+				t.Error("expect second request on broken agent connection to fail")
+			}
+		}
+
```
caesarxuchao, line 70: nit: remove the empty line

tests/agent_disconnect_test.go
caesarxuchao: Got it. Can you also say this in the comments of the tests?

tests/agent_disconnect_test.go
caesarxuchao: You can control the connection cache by creating or not creating new tunnel, right?

tests/agent_disconnect_test.go
caesarxuchao, line 70: nit: remove the empty line

tests/agent_disconnect_test.go
Jefftree: It's not a concern in the current test, but might cause debugging to be harder down the line if it becomes a standard. Logs also get polluted which is undesirable.

I updated this to use an alternate method,  polling for for proxy server readiness instead. I think this is a cleaner way to express what we want, let me know if you have additional suggestions. We can also replace the 1s sleeps with this function as well.

Jefftree: I removed the test for the failure when no backend(agent) is available since this PR isn't targeting dial failures. We handle dial failures by just waiting until the 30s dial timeout, and optimizing this is something that should be done in another PR. (Incorporating this test significantly slows down the test runtime at the moment)

caesarxuchao: Yeah, polling the readiness is much better. That's the right way to replace the 1s sleeps. We can do that in a followup.

caesarxuchao:
Currently the unit test would pass even without the fix. I think the http1 persistent connection is a necessary condition to trigger the bug, i.e., proxy server can't close the broken connection and the http1 client keeps using the broken connection.

I think you can update the clientRequestFunction keep reuse the transport to trigger the bug.

Jefftree:
I refactored the test to capture the http1 persistent connection behavior. PTAL, Thanks!

pkg/server/server.go
```diff
@@ -474,47 +485,68 @@ func (s *ProxyServer) Connect(stream agent.AgentService_ConnectServer) error {
 
 // route the packet back to the correct client
 func (s *ProxyServer) serveRecvBackend(stream agent.AgentService_ConnectServer, agentID string, recvCh <-chan *client.Packet) {
+	defer func() {
+		// Close all connected frontends when the agent connection is closed
+		// TODO(#126): Frontends in PendingDial state that have not been added to the
+		//             list of frontends should also be closed.
+		frontends, _ := s.getFrontends(agentID)
+		klog.Infof("<<< Close %d frontends connected to agent %s", len(frontends), agentID)
```
cheftako, line 493: Not worth holding. This log message seems more important than a data sent message. We should explore coming up with a sane log level strategy, which has something like state transitions (things like this a close connection) at a more significant log level than business as usual messages (sent a packet)

tests/agent_disconnect_test.go
```diff
@@ -0,0 +1,244 @@
...
+func createGrpcTunnelClient(proxyAddr, addr string) (*http.Client, error) {
+	tunnel, err := client.CreateSingleUseGrpcTunnel(proxyAddr, grpc.WithInsecure())
+	if err != nil {
+		return nil, err
+	}
+
+	c := &http.Client{
+		Transport: &http.Transport{
+			Dial: tunnel.Dial,
+		},
```
caesarxuchao, line 200: nit: can you set the Timeout here, and at line 81, checks that the error is not a timeout? Because otherwise, if this bug regresses, the test will just hang there and makes debugging it difficult.

tests/agent_disconnect_test.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+
+	dialer := func(network, addr string) (net.Conn, error) {
+		return conn, nil
+	}
+
+	c := &http.Client{
+		Transport: &http.Transport{
+			Dial: dialer,
+		},
```
caesarxuchao, line 240: Same here, can you set the Timeout?

tests/agent_disconnect_test.go
Jefftree: Added.

pkg/server/server.go
```diff
@@ -474,47 +485,68 @@ func (s *ProxyServer) Connect(stream agent.AgentService_ConnectServer) error {
 
 // route the packet back to the correct client
 func (s *ProxyServer) serveRecvBackend(stream agent.AgentService_ConnectServer, agentID string, recvCh <-chan *client.Packet) {
+	defer func() {
+		// Close all connected frontends when the agent connection is closed
+		// TODO(#126): Frontends in PendingDial state that have not been added to the
+		//             list of frontends should also be closed.
+		frontends, _ := s.getFrontends(agentID)
```
caesarxuchao, line 492: This might be problematic.

In HA setup, the same agent might end up with multiple streams to the proxy server when it tries to establish streams to all server instances. The agent will close the redundant streams. However, this cleanup function will close all frontends that connect via the agent, even if the supporting `stream` is not closed.

I think we should track frontends based on streams, and cleanup frontends based on streams.

zanetworker:
Is there any plan for a release with this fix included?

Avatat:
> Is there any plan for a release with this fix included?

I don't know, but we built own Docker image with fix included and it works like a charm :) The build is very easy process: https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/artifacts/images/server-build.Dockerfile

zanetworker:
yea we could do that but we have some policies in place, and I'd prefer using the k8s.io registry tag. Last resort we would build it ourselves and push to our registry.

Jefftree:
There was a bug introduced by this PR that will be fixed by https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/138. We will tag a release immediately after it's merged.

ETA: today/tomorrow

Avatat:
Thank you for the release, but we (@zanetworker) can't use it: https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/141

PR #130, opened 2020-07-23
Fix concurrent client test to use proper backend struct

Jefftree:
A refactor was done a while ago to use the Backend object in Backend Manager instead of directly manipulating the conn object. This PR fixes the test to follow the new conventions.

Unfortunately this results in a decent amount of duplicate code, but the alternative is to make the (lowercase) backend struct public which is not ideal either.

The PR also adds an AgentID() method to the Backend interface which will be used in a follow up PR.

/assign @caesarxuchao 
/cc @cheftako

### DIFF
--- a/pkg/server/backend_manager.go
+++ b/pkg/server/backend_manager.go
@@ -30,16 +30,17 @@ import (
 type Backend interface {
 	Send(p *client.Packet) error
 	Context() context.Context
+	AgentID() string
 }
 
 var _ Backend = &backend{}
-var _Backend = new(agent.AgentService_ConnectServer)
 
 type backend struct {
 	// TODO: this is a multi-writer single-reader pattern, it's tricky to
 	// write it using channel. Let's worry about performance later.
-	mu   sync.Mutex // mu protects conn
-	conn agent.AgentService_ConnectServer
+	mu      sync.Mutex // mu protects conn
+	conn    agent.AgentService_ConnectServer
+	agentID string
 }
 
 func (b *backend) Send(p *client.Packet) error {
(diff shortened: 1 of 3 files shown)


pkg/server/backend_manager.go
```diff
@@ -53,8 +54,12 @@ func (b *backend) Context() context.Context {
 	return b.conn.Context()
 }
 
-func newBackend(conn agent.AgentService_ConnectServer) *backend {
-	return &backend{conn: conn}
+func (b *backend) AgentID() string {
```
caesarxuchao, line 57: See https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/125/files#r460320297. We might need to expose the `conn`, not the `agentID`

tests/concurrent_client_request_test.go
```diff
@@ -52,39 +55,103 @@ func getTestClient(front string, t *testing.T) *http.Client {
...
+
+func newBackend(agentID string, conn agent.AgentService_ConnectServer) *backend {
+	return &backend{agentID: agentID, conn: conn}
+}
+
 // singleTimeManager makes sure that a backend only serves one request.
 type singleTimeManager struct {
 	mu       sync.Mutex
-	backends map[string]agent.AgentService_ConnectServer
+	backends map[string][]*backend
```
caesarxuchao, line 86: How about `map[string][]Backend`? Then you don't need to duplicate the `backend` code.

You might find it difficult to implement the `RemoveBackend` method if you use `Backend`. If so, we can skip the implementation of `RemoveBackend`, which is not used in this test anyway.

Jefftree: The newBackend function is private, so we'll still need to duplicate the `backend` code I think.

pkg/server/backend_manager.go
Jefftree: ack, I'll revise this with the conn instead.

tests/concurrent_client_request_test.go
caesarxuchao: Can we make newBackend function to public? That sounds a common pattern, a public function that returns a public interface implemented by a private object.

PR #136, opened 2020-07-23
Remove additional docker build before docker push

Jefftree:
Looking at the image push logs (http://storage.googleapis.com/k8s-staging-kas-network-proxy-gcb/logs/log-66bfaf44-b21f-424a-8128-fe6e371d7fcb.txt), it looks like the same image (eg: proxy-agent-arm64:v20200714-v0.0.10-15-g5afb5f2) is built twice before being pushed to the registry. 

This comes from `make release-staging` calling `docker-build-all docker-push-all` while `docker-push-all` calls another `docker-build` for each push. We should skip the redundant build.

/cc @Sh4d1 
/assign @cheftako

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -274,7 +274,7 @@ docker-push-manifest/http-test-server: ## Push the fat manifest docker image.
 
 .PHONY: release-staging
 release-staging: ## Builds and push container images to the staging bucket.
-	REGISTRY=$(STAGING_REGISTRY) $(MAKE) docker-build-all docker-push-all release-alias-tag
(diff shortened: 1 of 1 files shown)


Sh4d1:
Maybe removing the `docker-build-all` in the `release-staging` would be more coherent?

cheftako:
> Maybe removing the `docker-build-all` in the `release-staging` would be more coherent?

+1

Because we are dealing with '.phony' dependencies/targets we cannot rely on timestamps to dedupe. However push needs the containers to have been built first which is what these dependency statements indicate. As it stands, with this change, if you attempt to push a single container it will fail as the container has not been built.

Issue #126, opened 2020-07-15
Handle agent disconnects for PendingDial


Jefftree:
When an agent disconnects https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/125 closes all client side connections that use the corresponding agent. However, PendingDial requests may still be in flight and have not been added to the list of clients yet. We should either fail them or retry with a different agent instead of letting the client hit its dial timeout.

Original context from @cheftako:

> Most of the time I would expect pending dial to be empty. However if there is something in there, there is a chance its request went out via this backend. If so we will never get the response and that also needs to be dealt with.
> 
> The issue is that we do not record in the pending data structure which backend it used, so we cannot tell if anything on the pending list would be effected by a given backend breaking. We also need to work out how to deal with it. One option would be to just fail, which is probably easiest. However as the connection has not yet be established, we should be able to switch to using a different backend.

PR #138, opened 2020-08-05
Fix frontend cleanup for backend disconnect for HA servers

Jefftree:
Fixes https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/125#discussion_r460320297. Check to ensure that we have the right backend (may be multiple with the same agentID) before removing frontend connections. I avoided tracking frontends via streams because we will have one stream per backend almost all the time except when agents are establishing their connections, and we can avoid the additional complexity.

Also fixes a race condition on https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/server/server.go#L495 where the `frontends[agentID]` map could be modified while iterating through the loop.

Adding the backend object to the frontend ProxyClientConnection also allows PendingDial to keep track of the backend if it needs to be closed, providing an easy fix for https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/126 in a follow up.

### DIFF
--- a/pkg/server/backend_manager.go
+++ b/pkg/server/backend_manager.go
@@ -63,7 +63,7 @@ type BackendManager interface {
 	// Backend returns a single backend.
 	Backend() (Backend, error)
 	// AddBackend adds a backend.
-	AddBackend(agentID string, conn agent.AgentService_ConnectServer)
+	AddBackend(agentID string, conn agent.AgentService_ConnectServer) Backend
 	// RemoveBackend removes a backend.
 	RemoveBackend(agentID string, conn agent.AgentService_ConnectServer)
 }
@@ -95,23 +95,25 @@ func NewDefaultBackendManager() *DefaultBackendManager {
 }
 
 // AddBackend adds a backend.
-func (s *DefaultBackendManager) AddBackend(agentID string, conn agent.AgentService_ConnectServer) {
+func (s *DefaultBackendManager) AddBackend(agentID string, conn agent.AgentService_ConnectServer) Backend {
 	klog.Infof("register Backend %v for agentID %s", conn, agentID)
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	_, ok := s.backends[agentID]
+	addedBackend := newBackend(conn)
 	if ok {
 		for _, v := range s.backends[agentID] {
 			if v.conn == conn {
 				klog.Warningf("this should not happen. Adding existing connection %v for agentID %s", conn, agentID)
-				return
+				return v
 			}
 		}
-		s.backends[agentID] = append(s.backends[agentID], newBackend(conn))
-		return
+		s.backends[agentID] = append(s.backends[agentID], addedBackend)
+		return addedBackend
 	}
-	s.backends[agentID] = []*backend{newBackend(conn)}
+	s.backends[agentID] = []*backend{addedBackend}
 	s.agentIDs = append(s.agentIDs, agentID)
+	return addedBackend
 }
 
 // RemoveBackend removes a backend.
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -45,6 +45,7 @@ type ProxyClientConnection struct {
 	connectID int64
 	agentID   string
 	start     time.Time
+	backend   Backend
 }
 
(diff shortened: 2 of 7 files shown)


tests/proxy_test.go
```diff
@@ -45,6 +46,11 @@ func newSizedServer(length, chunks int) *testServer {
 func (s *testServer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	for i := 0; i < s.chunks; i++ {
 		w.Write(s.echo)
+
+		// Wait before sending the last chunk if test requires it
+		if i == (s.chunks-1) && s.wchan != nil {
```
caesarxuchao, line 51: This should be done before w.Write.

tests/concurrent_test.go
```diff
@@ -71,3 +72,73 @@ func TestProxy_Concurrency(t *testing.T) {
...
+			go func() {
+				_, err := clientRequest(c, server.URL)
+				if err != nil {
+					t.Errorf("expected no error on proxy request, got %v", err)
+				}
+				close(fcnStopCh)
+			}()
+
+			// Running an agent with the same ID simulates a second connection from the same agent.
+			// This scenario is often hit when running a HA proxy server
```
caesarxuchao, line 134: How about

"Running an agent with the same ID simulates a second connection from the same agent. This simulates the scenario where a proxy agent established connections with HA proxy server and creates multiple connections with the same proxy server."

tests/concurrent_test.go
```diff
@@ -71,3 +72,73 @@ func TestProxy_Concurrency(t *testing.T) {
 	}
 	wg.Wait()
 }
+
+func TestAgent_MultipleConn(t *testing.T) {
```
caesarxuchao, line 76: Can you add a comment about the purpose of this test, something like
"This test verifies that when one stream between a proxy agent and the proxy server terminates, the proxy server does not terminate frontends supported by the same proxy agent but different streams."

pkg/server/server.go
```diff
@@ -484,12 +496,12 @@ func (s *ProxyServer) Connect(stream agent.AgentService_ConnectServer) error {
 }
 
 // route the packet back to the correct client
-func (s *ProxyServer) serveRecvBackend(stream agent.AgentService_ConnectServer, agentID string, recvCh <-chan *client.Packet) {
+func (s *ProxyServer) serveRecvBackend(backend Backend, stream agent.AgentService_ConnectServer, agentID string, recvCh <-chan *client.Packet) {
```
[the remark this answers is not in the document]
Jefftree: We log the stream at the very end of the function. I think it should be fine to keep since we can't easily access the stream given the Backend interface.

pkg/server/backend_manager.go
```diff
@@ -95,23 +95,25 @@ func NewDefaultBackendManager() *DefaultBackendManager {
...
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	_, ok := s.backends[agentID]
+	addedBackend := newBackend(conn)
 	if ok {
 		for _, v := range s.backends[agentID] {
 			if v.conn == conn {
 				klog.Warningf("this should not happen. Adding existing connection %v for agentID %s", conn, agentID)
-				return
+				return nil
```
cheftako, line 108: This is dangerous and could lead to some bizarre and hard to track down bugs. We will take this nil and use it as the key in frontends to identify the backend. We then won't properly close down the frontend when its actual backend connection break. If we ever call getFrontendsForBackendConn with a backend of nil we could have multiple unrelated frontends all getting closed. Guessing we should just return v.

Jefftree: Good point, fixed.

tests/concurrent_test.go
```diff
@@ -71,3 +72,75 @@ func TestProxy_Concurrency(t *testing.T) {
 	}
 	wg.Wait()
 }
+
+// This test verifies that when one stream between a proxy agent and the proxy server terminates, the proxy server does not terminate other frontends supported by the same proxy agent but on different streams
```
caesarxuchao, line 76: nit: Can you break the line at 80 chars, and add a period at the end?

PR #139, opened 2020-08-12
Move metrics handler from admin to health server

Jefftree:
admin server is binded only to localhost. This will allow metrics to be read by other containers that need to process the metrics.

/assign @cheftako 
/cc @caesarxuchao

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -20,6 +20,7 @@ import (
 	"crypto/tls"
 	"flag"
 	"fmt"
+	"net"
 	"net/http"
 	"os"
 	"time"
@@ -236,6 +237,7 @@ func (a *Agent) runHealthServer(o *GrpcProxyAgentOptions) error {
 	})
 
 	muxHandler := http.NewServeMux()
+	muxHandler.Handle("/metrics", promhttp.Handler())
 	muxHandler.HandleFunc("/healthz", livenessHandler)
 	muxHandler.HandleFunc("/ready", readinessHandler)
 	healthServer := &http.Server{
@@ -257,7 +259,16 @@ func (a *Agent) runHealthServer(o *GrpcProxyAgentOptions) error {
 
 func (a *Agent) runAdminServer(o *GrpcProxyAgentOptions) error {
 	muxHandler := http.NewServeMux()
-	muxHandler.Handle("/metrics", promhttp.Handler())
+	muxHandler.Handle("/metrics", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		host, _, err := net.SplitHostPort(r.Host)
+		// The port number may be omitted if the admin server is running on port
+		// 80, the default port for HTTP
+		if err != nil {
+			host = r.Host
+		}
+		http.Redirect(w, r, fmt.Sprintf("%s:%d%s", host, o.healthServerPort, r.URL.Path), http.StatusMovedPermanently)
+	}))
+
 	adminServer := &http.Server{
 		Addr:           fmt.Sprintf("127.0.0.1:%d", o.adminServerPort),
 		Handler:        muxHandler,


cheftako:
/assign @logicalhan 
Tagging in @logicalhan to make sure he has no issues with exposing metrics on a non localhost server.

cmd/server/main.go
```diff
@@ -592,6 +591,7 @@ func (p *Proxy) runHealthServer(o *ProxyRunOptions, server *server.ProxyServer)
 	})
 
 	muxHandler := http.NewServeMux()
+	muxHandler.Handle("/metrics", promhttp.Handler())
 	muxHandler.HandleFunc("/healthz", livenessHandler)
 	muxHandler.HandleFunc("/ready", readinessHandler)
```
logicalhan, line 596: shouldn't this be `/readyz`?

cmd/server/main.go
```diff
@@ -551,7 +551,6 @@ func redirectTo(to string) func(http.ResponseWriter, *http.Request) {
 
 func (p *Proxy) runAdminServer(o *ProxyRunOptions, server *server.ProxyServer) error {
 	muxHandler := http.NewServeMux()
-	muxHandler.Handle("/metrics", promhttp.Handler())
```
cheftako, line 554: Can we add a redirect (status code 301) handler to the new port? We can remove it in a bit (and I know we have not gone to 1.0 yet) but its good to start ingraining backward compatibility concerns.

logicalhan: Why not just leave this here for backwards compatibility?

cheftako: I'd be ok with that solution as well. I just don't want the endpoint to disappear without explanation or any attempt to maint compatibility.

cmd/server/main.go
cheftako: shrug, Kubelet doesn't care. Happy to have it changed if we believe that is more standard.

cmd/server/main.go
cheftako: If we are ever plan to stop serving /metrics on the admin port, then we should 301 to the new destination. If we're ok keeping it in perpetuity then go ahead and leave it. If we think we would rather just maintain it in one place then its probably better to put in the 301 now, before usage goes up.

caesarxuchao:
Is it safe to expose metrics on non-localhost? Is it necessary?

Jefftree:
Hmm, it seems that this has more implications than I originally expected. Will do some more investigation on alternatives before proceeding with this route.

/hold

cheftako:
> Is it safe to expose metrics on non-localhost? Is it necessary?

Would love to get @logicalhan's (SIG Instrumentation) take on that.

mikedanese:
Talked to Jeff and this isn't a publicly exposed endpoint (at least by default).

/lgtm

cmd/server/main.go
Jefftree: Added the 301 redirect

Jefftree:
/unhold

@cheftako @caesarxuchao PTAL, thanks!

cmd/agent/main.go
```diff
@@ -257,7 +259,15 @@ func (a *Agent) runHealthServer(o *GrpcProxyAgentOptions) error {
...
 func (a *Agent) runAdminServer(o *GrpcProxyAgentOptions) error {
 	muxHandler := http.NewServeMux()
-	muxHandler.Handle("/metrics", promhttp.Handler())
+	muxHandler.Handle("/metrics", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		host, _, err := net.SplitHostPort(r.Host)
+		// The port number may be omitted if the admin server is running on port 80
+		if err != nil {
+			host = r.Host
+		}
+		http.Redirect(w, r, fmt.Sprintf("%s:%d%s", host, o.healthServerPort, r.URL.Path), 301)
```
caesarxuchao, line 268: nit: s/301/http.StatusMovedPermanently

cmd/agent/main.go
```diff
@@ -257,7 +259,15 @@ func (a *Agent) runHealthServer(o *GrpcProxyAgentOptions) error {
 
 func (a *Agent) runAdminServer(o *GrpcProxyAgentOptions) error {
 	muxHandler := http.NewServeMux()
-	muxHandler.Handle("/metrics", promhttp.Handler())
+	muxHandler.Handle("/metrics", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		host, _, err := net.SplitHostPort(r.Host)
+		// The port number may be omitted if the admin server is running on port 80
```
caesarxuchao, line 264: how about "The port number may be omitted if the admin server is running on port 80, the default port for HTTP"?

PR #143, opened 2020-08-27
Decouple the Backend method from the other methods of the BackendManager

### LINKED ISSUE #142: Allow backend manager to return backend based on agentID
Currently, `BackendManager.Backend()` does not take any input, and can only return a random agent. Sometimes, the server may want to pick up a specific backend, consider the following: 

Agents are located in different network regions, and the IP address is used as the agentID. when the server receives a request, it can choose the right agent based on the IP address.

charleszheng44:
1. Using the BackendStorage interface to decouple backend CRUD operations from the BackendManager
2. Implement the DesignatingBackendManager, which retrieves the backend associating to the given agentID 
resolve #142 
3. Change test cases

### DIFF
--- a/pkg/server/backend_manager.go
+++ b/pkg/server/backend_manager.go
@@ -57,21 +57,42 @@ func newBackend(conn agent.AgentService_ConnectServer) *backend {
 	return &backend{conn: conn}
 }
 
-// BackendManager is an interface to manage backend connections, i.e.,
-// connection to the proxy agents.
-type BackendManager interface {
-	// Backend returns a single backend.
-	Backend() (Backend, error)
+// BackendStorage is an interface to manage the storage of the backend
+// connections, i.e., get, add and remove
+type BackendStorage interface {
 	// AddBackend adds a backend.
 	AddBackend(agentID string, conn agent.AgentService_ConnectServer) Backend
 	// RemoveBackend removes a backend.
 	RemoveBackend(agentID string, conn agent.AgentService_ConnectServer)
+	// NumBackends returns the number of backends.
+	NumBackends() int
+}
+
+// BackendManager is an interface to manage backend connections, i.e.,
+// connection to the proxy agents.
+type BackendManager interface {
+	// Backend returns a single backend.
+	// WARNING: the context passed to the function should be a session-scoped
+	// context instead of a request-scoped context, as the backend manager will
+	// pick a backend for every tunnel session and each tunnel session may
+	// contains multiple requests.
+	Backend(ctx context.Context) (Backend, error)
+	BackendStorage
 }
 
 var _ BackendManager = &DefaultBackendManager{}
 
 // DefaultBackendManager is the default backend manager.
 type DefaultBackendManager struct {
+	*DefaultBackendStorage
+}
+
+func (dbm *DefaultBackendManager) Backend(_ context.Context) (Backend, error) {
+	return dbm.DefaultBackendStorage.GetRandomBackend()
+}
+
+// DefaultBackendStorage is the default backend storage.
+type DefaultBackendStorage struct {
 	mu sync.RWMutex //protects the following
 	// A map between agentID and its grpc connections.
 	// For a given agent, ProxyServer prefers backends[agentID][0] to send
@@ -88,14 +109,19 @@ type DefaultBackendManager struct {
 
 // NewDefaultBackendManager returns a DefaultBackendManager.
 func NewDefaultBackendManager() *DefaultBackendManager {
-	return &DefaultBackendManager{
+	return &DefaultBackendManager{DefaultBackendStorage: NewDefaultBackendStorage()}
+}
+
+// NewDefaultBackendStorage returns a DefaultBackendStorage
+func NewDefaultBackendStorage() *DefaultBackendStorage {
+	return &DefaultBackendStorage{
 		backends: make(map[string][]*backend),
 		random:   rand.New(rand.NewSource(time.Now().UnixNano())),
 	}
 }
 
 // AddBackend adds a backend.
-func (s *DefaultBackendManager) AddBackend(agentID string, conn agent.AgentService_ConnectServer) Backend {
+func (s *DefaultBackendStorage) AddBackend(agentID string, conn agent.AgentService_ConnectServer) Backend {
 	klog.Infof("register Backend %v for agentID %s", conn, agentID)
 	s.mu.Lock()
 	defer s.mu.Unlock()
@@ -117,7 +143,7 @@ func (s *DefaultBackendManager) AddBackend(agentID string, conn agent.AgentServi
 }
 
 // RemoveBackend removes a backend.
-func (s *DefaultBackendManager) RemoveBackend(agentID string, conn agent.AgentService_ConnectServer) {
+func (s *DefaultBackendStorage) RemoveBackend(agentID string, conn agent.AgentService_ConnectServer) {
 	klog.Infof("remove Backend %v for agentID %s", conn, agentID)
 	s.mu.Lock()
 	defer s.mu.Unlock()
@@ -151,6 +177,13 @@ func (s *DefaultBackendManager) RemoveBackend(agentID string, conn agent.AgentSe
 	}
 }
 
+// NumBackends resturns the number of available backends
+func (s *DefaultBackendStorage) NumBackends() int {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return len(s.backends)
+}
+
 // ErrNotFound indicates that no backend can be found.
 type ErrNotFound struct{}
 
@@ -159,8 +192,8 @@ func (e *ErrNotFound) Error() string {
 	return "No backend available"
 }
 
-// Backend returns a random backend.
-func (s *DefaultBackendManager) Backend() (Backend, error) {
+// GetRandomBackend returns a random backend.
+func (s *DefaultBackendStorage) GetRandomBackend() (Backend, error) {
 	s.mu.RLock()
 	defer s.mu.RUnlock()
 	if len(s.backends) == 0 {
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -270,7 +270,7 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer, 
 			// the address, then we can send the Dial_REQ to the
 			// same agent. That way we save the agent from creating
 			// a new connection to the address.
-			backend, err = s.BackendManager.Backend()
+			backend, err = s.BackendManager.Backend(context.TODO())
 			if err != nil {
 				klog.Errorf(">>> failed to get a backend: %v", err)
 				continue
--- a/pkg/server/tunnel.go
+++ b/pkg/server/tunnel.go
@@ -17,6 +17,7 @@ limitations under the License.
 package server
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"math/rand"
@@ -68,7 +69,7 @@ func (t *Tunnel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		},
 	}
 	klog.Infof("Set pending(rand=%d) to %v", random, w)
-	backend, err := t.Server.BackendManager.Backend()
+	backend, err := t.Server.BackendManager.Backend(context.TODO())
 	if err != nil {
 		http.Error(w, fmt.Sprintf("currently no tunnels available: %v", err), http.StatusInternalServerError)
 		return
--- a/tests/concurrent_client_request_test.go
+++ b/tests/concurrent_client_request_test.go
@@ -2,6 +2,7 @@ package tests
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -79,7 +80,7 @@ func (s *singleTimeManager) RemoveBackend(agentID string, conn agent.AgentServic
 	delete(s.backends, agentID)
 }
 
-func (s *singleTimeManager) Backend() (server.Backend, error) {
+func (s *singleTimeManager) Backend(_ context.Context) (server.Backend, error) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	for k, v := range s.backends {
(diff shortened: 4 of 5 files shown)


Jefftree:
Hello, thank you for the PR. I have one question about the usage of the DesignatingBackendManager. Is the invocation of this backend manager going to be added in a separate PR? AFAIK the proxy server currently doesn't have any information about which agentID a request would route to (And I see in your PR that the Backend() invocation does not pass in the agentID parameter (https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/server/server.go#L273)). Would this routing be something computed by the proxy server or passed in from a request?

Also, I noticed in https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/142 you mentioned that the server would choose a corresponding backend based on the agent IP address/region. I'm wondering if using agentID is a bit too granular and passing in something like a network region (so the proxy server can appropriately load balance requests between multiple agents in the same region) would be more appropriate.

/cc @cheftako

charleszheng44:
@Jefftree Thanks for your reply. As we discussed in issue #142 , in the future, we may want to support different proxy strategies, which means the `Backend()` method may need to take in different numbers/types of variables. To this end, I refactored the BackendManager interface as following: 

```go
type BackendManager interface {
	// Backend returns a single backend.
	Backend(i ...interface{}) (Backend, error)
	BackendStorage
}
```

Then, users can implement their own BackendManager that support different strategies, for example
```
// choose agent by network region
RegionBackendManager.Backend(regionName)
// choose agent by datacenter
DatacenterBackendManager.Backend(dcName)
// or something advanced
LoadBalanceBackendManager.Backend(lbStrategy)
```

And when initializing the proxy server, users can specify which BackendManager they wanna use (https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/edd6f7f63a7229e3e2dcf7e31983d06cf06c46c2/cmd/server/main.go#L137)

at the line ( https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/server/server.go#L273), there is no argument passed into the Backend method, because I was not sure if we want to change the behavior of the DefaultBackendManager, which chooses a random agent for the incoming request.

In our case, we set up the agent on every node and use the nodeIP as the agentID, therefore, in the implementation of the DesignatedBackendManager, I uses the agentID to query the desired node.

Jefftree (commented):
@charleszheng44 thanks for the explanation. A couple of overall thoughts:

- +1 for decoupling BackendManager and BackendStorage.
- Using an interface array and type casting for the Backend() function seems very prone to error. I'd recommend either using a context or struct object and embedding desired parameters as fields within the object.
- It should be fine to pass in a parameter to https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/server/server.go#L273 since the DefaultBackendManager would ignore the parameter anyway.

pkg/server/backend_manager.go
```diff
@@ -57,21 +59,66 @@ func newBackend(conn agent.AgentService_ConnectServer) *backend {
...
+	return be, nil
+}
+
+var _BackendManager = &DesignatingBackendManager{}
+
+type DesignatingBackendManager struct {
+	BackendStorage
+}
+
+func (dbm *DesignatingBackendManager) Backend(i ...interface{}) (Backend, error) {
```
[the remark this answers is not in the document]
charleszheng44: change the parameter to `context.Context`

pkg/server/backend_manager.go
```diff
@@ -159,16 +233,16 @@ func (e *ErrNotFound) Error() string {
 	return "No backend available"
 }
 
-// Backend returns a random backend.
-func (s *DefaultBackendManager) Backend() (Backend, error) {
+// GetRandomBackend returns a random backend.
+func (s *DefaultBackendStorage) GetRandomBackend() Backend {
```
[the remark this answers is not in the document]
charleszheng44: I forgot to commit the change when submitting the PR. The latest commit contains the implementation that invokes different Backend method for different proxy strategy

pkg/server/backend_manager.go
```diff
@@ -87,15 +134,35 @@ type DefaultBackendManager struct {
...
+
+// NewDefaultBackendStorage returns a DefaultBackendStorage
+func NewDefaultBackendStorage() *DefaultBackendStorage {
+	return &DefaultBackendStorage{
 		backends: make(map[string][]*backend),
 		random:   rand.New(rand.NewSource(time.Now().UnixNano())),
 	}
 }
 
+// GetBackend gets a backend associating to the agentID
```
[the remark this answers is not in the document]
charleszheng44: `s.backends[agentID][0]` should be the first connection of the agent?

charleszheng44:
@Jefftree I revised the code based on your suggestions. Also, I change the `DesignatingBackendManager` to `DestIPBackendManager` as it will return the backend that associates to the request's destination IP. 

However, there is one more issue, the current implementation requires the agent to use the IP address as the `agentID` when connecting to the server, something like:

```yaml
hostNetwork: true
containers:
  command:
  - proxy-agent
  args:
  - --agent-id=$(POD_IP)
  env:
  - name: POD_IP
     valueFrom:
       fieldRef:
         fieldPath: status.podIP
```

I am not sure if we should keep it this way, or maybe we can let the agent passes the `nodeName` to the proxy server, then the BackendManager can get the IP, regionID by accessing the node object, and groups agents based on the strategy, e.g., `DestIPBackendManager` uses `map[IP]conn` and `RegionIDBackendManager` uses `map[regionID]conn`.

cmd/server/main.go
```diff
@@ -272,6 +275,13 @@ func (o *ProxyRunOptions) Validate() error {
 		}
 	}
 
+	switch o.proxyStrategy {
+	case "destIP":
+	case "default":
```
cheftako, line 280: It would be good if we started making our flags backward compatible. Specifically older scripts/containers should get the origianl/default behavior id the flag is missing. Can we make empty string ("") equivalent to default?

Jefftree:
+1 for the agent passing additional metadata to the proxy server. We already embed auth information and the agentID in the agent -> server connection header (https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/agent/client.go#L135), adding additional fields should be fairly straightforward. That's probably outside the scope of this decoupling PR though.

Also, would you be able to split out the DestIPBackendManager code into a separate PR? I have some specific comments around that, but it doesn't need to hold up the rest of this PR. This PR can just address the decoupling of the backend methods.

pkg/server/backend_manager.go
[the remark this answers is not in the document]
Jefftree: Yep, the nit was just referring to the comment above the function.

pkg/server/server.go
```diff
@@ -270,7 +281,15 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
 			// the address, then we can send the Dial_REQ to the
 			// same agent. That way we save the agent from creating
 			// a new connection to the address.
-			backend, err = s.BackendManager.Backend()
+			ctx := context.Background()
+			switch s.proxyStrategy {
+			case "destIP":
+				dialReq := pkt.GetDialRequest()
+				ip := strings.Split(dialReq.Address, ":")[0]
```
cheftako, line 288: This is very IPv4 specific and will do the wrong thing with IPv6.

pkg/server/tunnel.go
```diff
@@ -68,7 +70,14 @@ func (t *Tunnel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		},
 	}
 	klog.Infof("Set pending(rand=%d) to %v", random, w)
-	backend, err := t.Server.BackendManager.Backend()
+	ctx := context.Background()
+	switch t.Server.proxyStrategy {
+	case "destIP":
+		ip := strings.Split(r.Host, ":")[0]
```
cheftako, line 76: Again IPv4 specific.

cheftako: Can we use a utility?

pkg/server/backend_manager.go
```diff
@@ -57,21 +60,66 @@ func newBackend(conn agent.AgentService_ConnectServer) *backend {
...
+var _BackendManager = &DestIPBackendManager{}
+
+type DestIPBackendManager struct {
+	BackendStorage
+}
+
+func (dbm *DestIPBackendManager) Backend(ctx context.Context) (Backend, error) {
+	destIPInf := ctx.Value(DestIPContextKey)
+	if destIPInf == nil {
+		return nil, errors.New("no agentID found in the context")
```
cheftako, line 108: Not sure there is a legitimate way for the context to be missing this value, but if one crops up, I'm not sure that refusing to tunnel the traffic is the best available behavior.

pkg/server/backend_manager.go
```diff
@@ -57,21 +60,66 @@ func newBackend(conn agent.AgentService_ConnectServer) *backend {
...
+	}
+	agentID, ok := destIPInf.(string)
+	if !ok {
+		return nil, errors.New("type assertion failed")
+	}
+	be := dbm.BackendStorage.GetBackend(agentID)
+	if be == nil {
+		return nil, &ErrNotFound{}
```
cheftako, line 116: This does not seem desirable. If the agent is down on the relevant node, we are saying that we will refuse to attempt to send traffic to that node. Its possible the node is up and that we are exec a request to fix the agent. The existing kube-dns code should route the traffic. I think we want this as a best effort to route through the best tunnel rather than only routing through that tunnel.

cheftako: If you do have a use case for *only* sending to the requested IP, I would suggest we have an addition parameter which is a ?strictness or ?exactDestination flag. That way the admin can control which way it goes.

charleszheng44:
> +1 for the agent passing additional metadata to the proxy server. We already embed auth information and the agentID in the agent -> server connection header (https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/agent/client.go#L135), adding additional fields should be fairly straightforward. That's probably outside the scope of this decoupling PR though.
> 
> Also, would you be able to split out the DestIPBackendManager code into a separate PR? I have some specific comments around that, but it doesn't need to hold up the rest of this PR. This PR can just address the decoupling of the backend methods.

Yes. I will create an independent PR for the DestIPBackendManager

charleszheng44:
@Jefftree @cheftako Just removed the DestIPBackendManager code from this PR as @Jefftree suggested. Now, this PR just addresses the decoupling problem.

Jefftree (commented):
A couple small nits, otherwise LGTM.

@caesarxuchao or @cheftako can give you the final approve.

pkg/server/backend_manager.go
```diff
@@ -87,15 +110,30 @@ type DefaultBackendManager struct {
...
+// NewDefaultBackendStorage returns a DefaultBackendStorage
+func NewDefaultBackendStorage() *DefaultBackendStorage {
+	return &DefaultBackendStorage{
 		backends: make(map[string][]*backend),
 		random:   rand.New(rand.NewSource(time.Now().UnixNano())),
 	}
 }
 
+// GetBackend gets a backend associating to the agentID
+func (s *DefaultBackendStorage) GetBackend(agentID string) Backend {
```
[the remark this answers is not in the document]
charleszheng44: Yes. the function has been removed.

pkg/server/backend_manager_test.go
```diff
@@ -34,7 +34,7 @@ func TestAddRemoveBackends(t *testing.T) {
 	conn22 := new(fakeAgentService_ConnectServer)
 	conn3 := new(fakeAgentService_ConnectServer)
 
-	p := NewDefaultBackendManager()
+	p := NewDefaultBackendManager(NewDefaultBackendStorage())
```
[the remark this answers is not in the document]
charleszheng44: Yes. the initialization of the DefaultBackendStorage has been included in the function NewDefaultBackendManager.

pkg/server/backend_manager.go
```diff
@@ -159,16 +204,16 @@ func (e *ErrNotFound) Error() string {
 	return "No backend available"
 }
 
-// Backend returns a random backend.
-func (s *DefaultBackendManager) Backend() (Backend, error) {
+// GetRandomBackend returns a random backend.
```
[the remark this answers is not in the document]
charleszheng44: Restored back to the original function signature.

tests/concurrent_client_request_test.go
```diff
@@ -91,6 +92,14 @@ func (s *singleTimeManager) Backend() (server.Backend, error) {
 	return nil, fmt.Errorf("cannot find backend to a new agent")
 }
 
+func (s *singleTimeManager) GetBackend(agentID string) server.Backend {
+	return nil
+}
+
+func (s *singleTimeManager) NumBackends() int {
```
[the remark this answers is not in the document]
charleszheng44: I agree. let's keep it like this just for now. We can refactor the `_test` in another PR.

Jefftree:
@charleszheng44 Thanks for the refactor! I'm sure the interface will change as we add additional backend selection strategies but this is a good step towards it.

/lgtm
/assign @cheftako

pkg/server/backend_manager.go
```diff
@@ -57,21 +57,38 @@ func newBackend(conn agent.AgentService_ConnectServer) *backend {
...
 	RemoveBackend(agentID string, conn agent.AgentService_ConnectServer)
+	// NumBackends returns the number of backends.
+	NumBackends() int
+}
+
+// BackendManager is an interface to manage backend connections, i.e.,
+// connection to the proxy agents.
+type BackendManager interface {
+	// Backend returns a single backend.
+	Backend(ctx context.Context) (Backend, error)
```
cheftako, line 75: Can we add a warning that we should not be passing individual request context objects here? Finishing handling a request on the tunnel signifies that request is complete and *not* that the tunnel is complete.

cheftako:
Did a review. Basically goods. A little concerned about the context change in isolation. I would at least like a warning comment put in, not to confuse a request scoped context with a tunnel/session scoped context. I do not want to debug a problem where we keep ending tunnels early because we crossed the two scopes of context. Especially as I assume most of the contexts will be of the wrong scope (i.e. they are request scoped)

pkg/server/backend_manager.go
charleszheng44: Yes. that is a very good point. I have added a warning comment.

PR #144, opened 2020-09-18
Feature: implement the BackendManager list

charleszheng44:
This PR implements the BackendManager list mechanism, which allows the proxy server to have multiple backend manager(BM). Each BM can have a corresponding BackendStorage(BS) that will group agents by the corresponding header.

For example, we have a list of BMs: `[DestHostBM, DestCIDRBM, DefaultBM]` 

All of them will use the `DefaultBackendStorage`, but the `DestHostBM`(implemented in this PR)will group agents by AgentHost, i.e., `map[AgentHost]conn`, the `DestCIDRBM`(will be included in an independent PR) will group agents by AgentCIDR, i.e. `map[AgentCIDR]conn`, and the `DefaultBM` will group agents by AgentID, using `map[AgentID]conn`.

When an agent joins the server, it can provide a list of addresses that are reachable through it and will be added into different BMs (but will always be added into the DefaultBM), say, an agent providing only AgentCIDR will not be added into the DestHostBM, but if it provides AgentHost as well as AgentCIDR, it will be added into all the three BMs. 

During the runtime, proxy server will call `Backend()` method in the order of the listed BM. For example, given the list `[DestHostBM, DestCIDRBM, DefaultBM]`, the server will first try to find a backend with the same host address, if not found, it will use the backend whose CIDR cover the request's host, if still not found, then the server will just choose a random backend.

Take the [GRPC+mTLS](https://github.com/kubernetes-sigs/apiserver-network-proxy#grpc-client-using-mtls-proxy-with-dial-back-agent) as an example, to enable the BackendManager list, one can run the `proxy-server` with the option `--proxy-strategies "destHost,default"` and the `proxy-agent` with the option `--agent-addr ""host=localhost,cidr=127.0.0.1/16,ipv6=:::::,failure-zone=us-central1-b""`.


pkg/agent/clientset.go
```diff
@@ -118,6 +121,7 @@ func (cc *ClientSetConfig) NewAgentClientSet(stopCh <-chan struct{}) *ClientSet
 	return &ClientSet{
 		clients:                 make(map[string]*AgentClient),
 		agentID:                 cc.AgentID,
+		agentHost:                 cc.AgentHost,
```
Jefftree, line 124: nit: same as above

cmd/server/main.go
```diff
@@ -132,6 +136,7 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.agentServiceAccount, "agent-service-account", o.agentServiceAccount, "Expected agent's service account during agent authentication (used with agent-namespace, authentication-audience, kubeconfig).")
 	flags.StringVar(&o.kubeconfigPath, "kubeconfig", o.kubeconfigPath, "absolute path to the kubeconfig file (used with agent-namespace, agent-service-account, authentication-audience).")
 	flags.StringVar(&o.authenticationAudience, "authentication-audience", o.authenticationAudience, "Expected agent's token authentication audience (used with agent-namespace, agent-service-account, kubeconfig).")
+	flags.StringVar(&o.proxyStrategy, "proxy-strategy", o.proxyStrategy, "TODO ...")
```
Jefftree, line 139: Missed a todo?

pkg/server/server.go
```diff
@@ -446,6 +477,18 @@ func (s *ProxyServer) Connect(stream agent.AgentService_ConnectServer) error {
...
 	}
+	if s.proxyStrategy == "destHost" {
+		// if the proxyStrategy is "destHost", use the agentHost as the ID
+		agentHost, err := agentHost(stream)
+		if err != nil {
+			return err
+		}
+		if agentHost != "" {
+			// if the agentHost is set, assign it to the agentID
+			agentID = agentHost
```
Jefftree, line 488: I'm not a fan of changing the agentID, especially if one was provided already. This seems to be only used for speeding up the backend selection, and it should be possible to optimize around there rather than the renaming.

@cheftako do you have any thoughts?

pkg/server/server.go
```diff
@@ -446,6 +477,18 @@ func (s *ProxyServer) Connect(stream agent.AgentService_ConnectServer) error {
 	if err != nil {
 		return err
 	}
+	if s.proxyStrategy == "destHost" {
```
Jefftree, line 480: Should the proxy server with this proxy strategy reject agents without the corresponding header? I'd imagine cases where a subset of agents are missing the header would be very difficult to detect.

proto/header/header.go
```diff
@@ -20,6 +20,7 @@ const (
 	ServerCount = "serverCount"
 	ServerID    = "serverID"
 	AgentID     = "agentID"
+	AgentHost     = "agentHost"
```
Jefftree, line 23: nit: formatting

pkg/server/server.go
```diff
@@ -270,7 +281,12 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
 			// the address, then we can send the Dial_REQ to the
 			// same agent. That way we save the agent from creating
 			// a new connection to the address.
-			backend, err = s.BackendManager.Backend(context.TODO())
+			ctx := context.Background()
+			if s.proxyStrategy == "destHost" {
+				addr := util.RemovePortFromHost(pkt.GetDialRequest().Address)
```
Jefftree, line 286: Does this mean it's only possible to send requests to nodes? Pod IPs within the cluster are off limits?

Jefftree (commented):
Thanks for the PR @charleszheng44!

One concern I have is that this the way this is currently written, all kube-apiserver -> cluster requests must use some node IP as a destination address...this seems to be a big implicit assumption and may break certain functionality. 

Limiting a request to only one routing path also seems a bit fragile. If the connection to the underlying agent is broken/overloaded/etc, is there no alternative routing for the corresponding request?

Overall the PR is good, I'm just thinking in a more generalized way to see if this can support a broader set of use cases.

cmd/server/main.go
```diff
@@ -107,6 +107,10 @@ type ProxyRunOptions struct {
 	authenticationAudience string
 	// Path to kubeconfig (used by kubernetes client)
 	kubeconfigPath string
+
+	// Proxy strategy used by the server, can be either "destHost" or "default".
+	// If not set, the "default" strategy will be used.
+	proxyStrategy string
```
cheftako, line 113: We haven't been great about this up to this point. However limited value variables should really be enums. Should consider making this an enum.

cmd/server/main.go
```diff
@@ -157,6 +162,7 @@ func (o *ProxyRunOptions) Print() {
 	klog.Warningf("AgentServiceAccount set to %q.\n", o.agentServiceAccount)
 	klog.Warningf("AuthenticationAudience set to %q.\n", o.authenticationAudience)
 	klog.Warningf("KubeconfigPath set to %q.\n", o.kubeconfigPath)
+	klog.Warningf("ProxyStrategy set to %s.\n", o.proxyStrategy)
```
cheftako, line 165: Would prefer %q to %s.

pkg/server/desthost_backend_manager.go
```diff
@@ -0,0 +1,39 @@
+package server
```
cheftako, line 1: Needs a 2020 Kubernetes Authors copyright block

pkg/server/desthost_backend_manager.go
```diff
@@ -0,0 +1,39 @@
...
+	if len(dibm.backends) == 0 {
+		return nil, &ErrNotFound{}
+	}
+	destHost := ctx.Value("destHost").(string)
+	if destHost != "" {
+		bes, exist := dibm.backends[destHost]
+		if exist && len(bes) > 0 {
+			return dibm.backends[destHost][0], nil
+		}
+		return nil, &ErrNotFound{}
```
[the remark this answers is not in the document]
cheftako: This behavior seems to differ from the intro comment ("If the desired backend is not found, just return a default one")

pkg/server/desthost_backend_manager.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	// fallback to the default behavior, i.e., get a random backend
+	agentID := dibm.agentIDs[dibm.random.Intn(len(dibm.agentIDs))]
+	klog.Infof("pick agentID=%s as backend", agentID)
+	return dibm.backends[agentID][0], nil
```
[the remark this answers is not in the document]
cheftako: Can we compose backend managers?

Have a composable list of backend managers. If you want host match with the default/random behavior as a fallback you have [desthost, default]. If want exact host match you have [desthost]. Then there is a driver which calls the list in order till it gets a result. Returns an error with no match.

pkg/server/server.go
```diff
@@ -270,7 +281,12 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
 			// the address, then we can send the Dial_REQ to the
 			// same agent. That way we save the agent from creating
 			// a new connection to the address.
-			backend, err = s.BackendManager.Backend(context.TODO())
+			ctx := context.Background()
+			if s.proxyStrategy == "destHost" {
```
cheftako, line 285: Can we make destHost a constant?

pkg/server/server.go
cheftako: +1, If need to make requests of the KAS about the agent, this is the identifier for those requests. Having the requests work initially (prior to this line) and then stop working (once this is executed is going to be very confusing. I realize we have no queries like that today but other similar backend managers are likely to need to make sure queries. (Eg. to lookup the related failure zone)

pkg/server/server_test.go
```diff
@@ -159,7 +159,7 @@ func TestAgentTokenAuthenticationErrorsToken(t *testing.T) {
 				conn.EXPECT().Recv().Return(nil, io.EOF)
 			}
 
-			p := NewProxyServer("", 1, &AgentTokenAuthenticationOptions{
```
cheftako, line 162: Also can we make default a constant?

pkg/util/net.go
```diff
@@ -0,0 +1,50 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
cheftako, line 2: Should be 2020.

cheftako: If the header check fails, let me know.

pkg/server/desthost_backend_manager.go
cheftako: This should have caused a header check failure. If not please let me know.

pkg/util/net_test.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+func TestRemovePortFromHost(t *testing.T) {
+	tests := []struct {
+		name     string
+		origHost string
+		expect   string
+	}{
+		{"Domain&Port", "localhost:8080", "localhost"},
+		{"Domain", "localhost", "localhost"},
```
cheftako, line 35: Can we add an addition full domain name test? Eg. www.kubernetes.test. As both domain and Domain&Port?

pkg/server/desthost_backend_manager.go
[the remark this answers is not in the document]
charleszheng44: The comment is wrong. The correct one is "If the desired backend is not found, fall back to the default behavior and just return a random one"

pkg/server/desthost_backend_manager.go
```diff
@@ -0,0 +1,39 @@
...
+// Backend tries to get a backend by IP. If the desired backend is not found,
+// just return a default one
+func (dibm *DestHostBackendManager) Backend(ctx context.Context) (Backend, error) {
+	dibm.mu.RLock()
+	defer dibm.mu.RUnlock()
+	if len(dibm.backends) == 0 {
+		return nil, &ErrNotFound{}
+	}
+	destHost := ctx.Value("destHost").(string)
+	if destHost != "" {
```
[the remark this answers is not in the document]
charleszheng44: I try to achieve this inside the `DestHostBackendManager.Backend(ctx context.Context)`, i.e. if we can't find an agent that has the desired host, then we will choose a random agent.

pkg/server/desthost_backend_manager.go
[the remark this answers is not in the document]
charleszheng44: That's a good idea. Yes, I will make a composable list of backend managers. The backendmanagers inside the list will be called in order. Let's say we have a list `[desthost, default]`, then we will first call  `DestHostBackendManager.Backend(ctx)`, if it return none, the `DefaultBackendManager.Backend(ctx)` will be called in stead.

pkg/server/server.go
charleszheng44: I was not sure if users forget to run agents with the corresponding proxy strategy (i.e., not providing corresponding headers), should the server still take the agent.   
Yes, I agree, this behavior can make it difficult to bug. Let's assume that if the server is running with a certain proxy strategy, agents not providing the required header will be rejected.

pkg/server/server.go
charleszheng44: Will KAS send requests to POD IPs directly? I thought all requests will be just from KAS to Kubelet.

cmd/server/main.go
charleszheng44: will make it enum

cmd/server/main.go
charleszheng44: will change to %q

pkg/server/desthost_backend_manager.go
charleszheng44: Will a wrong copyright year cause a header check failure?

pkg/server/server.go
charleszheng44: will make it constant

pkg/server/server_test.go
charleszheng44: will make it a constant

pkg/util/net_test.go
charleszheng44: Will add test cases for FQDN

charleszheng44:
@Jefftree @cheftako Thanks for your comments. I think both of you agree on that the current implementation is not generic enough. Once we specify a proxy strategy on the server side, the agents have to provide required headers, e.g. `destHost`  require `AgentIP` header, this requires all kube-apiserver -> cluster requests must use some node IP as the destination, while some request may use pod IP as dest host.  

I think @cheftako 's suggestion can be a good solution to this problem. We can compose a BackendManager(BM) list containing multiple BMs, and run all of them with the proxy server. Each BM can have a corresponding BackendStorage(BS) that will group agents by corresponding header.

For example, we have a list of BMs: [DestHostBM, DestCIDRBM, DefaultBM]

DestHostBM uses DestHostBS, which groups agents by AgentHost, i.e., `map[AgentHost]conn`

DestCIDRBM uses DestIPRangeBS, which groups agent by AgentCIDR, i.e. `map[AgentCIDR]conn`

DefaultBM uses DefaultBS, which groups agetns by AgentID, using `map[AgentID]conn`

When a agent join the server, based on the headers it provided, it will be added into different BSs (but will always be added into the DefaultBS), say, an agent providing only `AgentCIDR`  will not be added in to the `DestHostBS`, but if it provides `AgentHost` as well as `AgentCIDR`, it will be added in to all the three BSs.  

During the runtime, proxy server will call `Backend()` method in the order of the listed BM. For example, given list  [DestHostBM, DestCIDRBM, DefaultBM], the server will first try to find backend with same host address, if not found, it will use the backend whose CIDR cover the request host, if still not found, then the server will just choose a random backend. 

I think this mechanism can cover most of the cases. Please let me know your thoughts. If you agree on this plan, I will implement the DestHostBM in this PR, and submit an independent PR for the DestCIDRBM.

cheftako:
Hi @charleszheng44 hows the work coming. Very much like the direction your heading. Please let us know if we can help.

charleszheng44:
> Hi @charleszheng44 how's the work coming? Very much like the direction your heading. Please let us know if we can help.

Hi @cheftako, If you agree that the "BackendManager(BM) list containing multiple BM" might be a good solution, I will try to wrap up this PR by the end of this week.

cheftako:
> > Hi Chao Zheng how's the work coming? Very much like the direction your heading. Please let us know if we can help.
> 
> Hi Walter Fender, If you agree that the "BackendManager(BM) list containing multiple BM" might be a good solution, I will try to wrap up this PR by the end of this week.

Sounds like a great solution to me. Were you thinking an ordered, comma seperated list of the BMs? Probably defaulting to just the current BM?

pkg/server/desthost_backend_manager.go
cheftako: +1

cmd/agent/main.go
```diff
@@ -70,6 +70,7 @@ type GrpcProxyAgentOptions struct {
 	adminServerPort  int
 
 	agentID       string
+	agentHost     string
```
cheftako, line 73: I generally like  this but wonder if we're short changing on the agentHost. I'm assuming this is just a single IP address and that we are running 1 agent per node. However we frequently give nodes IP ranges to assign to each pod running on that node its own individual IP address. (See https://cloud.google.com/kubernetes-engine/docs/concepts/network-overview). In addition we may have both IPv4 and IPv6 addresses for both nodes and pods. I wonder if it would be better to have a map of agent properties? Then we can easily extend this for things like exact IP match, range IP match, rang IPv6 match, failure zone match, etc. We could encode this as something like "host=foo,cidr=127.0.0.1/16,ipv6=:::::,failure-zone=us-central1-b". If we wanted to be really paranoid we could url encode it (https://en.wikipedia.org/wiki/Percent-encoding).

charleszheng44: I like the idea of encoding all potential addresses into a single string. What about something like the following,

1. when starting an agent, users can specify all potential addresses by passing a string, "host=foo,cidr=127.0.0.1/16,ipv6=:::::,failure-zone=us-central1-b" through the command line,

2. the agent will validate if the the addresses' string is valid and add it to the gRPC header,

3. then, the server will deserialize the addresses from the header, and add the agent to corresponding backendmanager.

As we can put the addresses' string into the request header(we are currently storing the `agentHost` and `agentID` in the gRPC stream header), do we really need to do URL encoding? I am not sure if the gRPC header can hold reserved characters of the Percent-encoding, but I will try it out.

Jefftree: Question for both of you: How much of the information in `host=foo,cidr=127.0.0.1/16,ipv6=:::::,failure-zone=us-central1-b` etc is directly obtainable when starting the konnectivity agent pod? Are these parameters directly obtainable through something like environment variables or would additional logic be needed? Theoretically kube-apiserver would have this kind of information about the nodes, but adding additional logic there seems quite complicated. I'm wondering how we can support this in the simplest way.

charleszheng44: IMO, there may be two kinds of "agent address",

1. addresses that can be got directly when starting the konnectivity agent, like podIP(s), hostIP, etc,

2. addresses that added by user through cmd option, like, `--agent-addresses="host=foo,cidr=127.0.0.1/16,ipv6=:::::,failure-zone=us-central1-b"`, if some addresses are stored in env, we can pass them through the pod's template yaml. 

And it will be the users' (or whoever starts the agent ) responsiblity to ensure the provided addresses are reachable.

Jefftree:
Hi @charleszheng44, I took a quick skim and the overall logic looks good. I think my only concern is around where to obtain the parameters. The two most common ways to deploy the agent is via DaemonSet and Deployment. When creating the [`konnectivity-agent.yaml`](https://kubernetes.io/docs/tasks/extend-kubernetes/setup-konnectivity/#configure-the-konnectivity-service) kubernetes object, it's not clear to me how the agent-address parameter is determined. Since nodes come and go, I'm going to assume that this information isn't hard coded and we're using something like the [Downward API](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/) to get things like pod IP, node name, etc from env variables. I'm not exactly sure if the variables needed for agentAddress are all available?

Is there some additional capability of a cloud provider that we are relying on? Also, is hostNetwork for all agent pods an assumption when using destHost routing? (Else the agent pod and the node will use different IP addresses)

Would it be possible to get a sample of how a `konnectivity-agent.yaml` would look assuming this change is merged?

charleszheng44:
@Jefftree thanks for the comments.

charleszheng44:
@Jefftree Thanks for the comments. 
```
it's not clear to me how the agent-address parameter is determined. 
```
IMO, it depends on how does the konnectivity-agent.yaml is generated, if it is generated using tools like kustomized or helm, we can put a place holder `--agent-address={.agentAddr}` in the konnectivity-agent.yaml template. Or we can use env refer to the podIP and the hostIP, for example
```yaml
...
--agent-addr="host=$(NODE_NAME),ipv4=$(POD_IP)"
...
env
- name: NODE_NAME
     valueFrom:
       fieldRef:
         fieldPath: spec.nodeName
- name: POD_IP
     valueFrom:
       fieldRef:
         fieldPath: status.podIP
```
```
Is there some additional capability of a cloud provider that we are relying on? 
```
No, for now. But if we plan to support additional address types, like CIDR or failure-zone, we may require to use some cloud provider specific env to pass the information to the agent

```
is hostNetwork for all agent pods an assumption when using destHost routing?
``` 
I don't think the hostNetwork is necessary. We can get the node address through `status.hostIP`(the node's IP, available since v1.7.0-alpha.1), and add it to the `--agent-Addr`

```
Would it be possible to get a sample of how a konnectivity-agent.yaml?
```
Yes, I can make one. Should I include a konnectivity-agent.yaml sample in this PR?

Jefftree:
> IMO, it depends on how does the konnectivity-agent.yaml is generated, if it is generated using tools like kustomized or helm, we can put a place holder `--agent-address={.agentAddr}` in the konnectivity-agent.yaml template. Or we can use env refer to the podIP and the hostIP, for example
> 
> ```yaml
> ...
> --agent-addr="host=$(NODE_NAME),ipv4=$(POD_IP)"
> ...
> env
> - name: NODE_NAME
>      valueFrom:
>        fieldRef:
>          fieldPath: spec.nodeName
> - name: POD_IP
>      valueFrom:
>        fieldRef:
>          fieldPath: status.podIP
> ```
Ahh I totally forgot about that. Thanks for the clarification.

> ```
> is hostNetwork for all agent pods an assumption when using destHost routing?
> ```
> 
> I don't think the hostNetwork is necessary. We can get the node address through `status.hostIP`(the node's IP, available since v1.7.0-alpha.1), and add it to the `--agent-addr`
> 

I see, so we're providing the node IP as the agentAddress and not the agent pod IP. Sorry just a follow up question, are any changes needed in the kubernetes repo? I'm not sure if any control plane -> cluster requests route to the node IP by default.

> ```
> Would it be possible to get a sample of how a konnectivity-agent.yaml?
> ```
> 
> Yes, I can make one. Should I include a konnectivity-agent.yaml sample in this PR?

 I think it'd be nice to include a snippet in our [examples](https://github.com/kubernetes-sigs/apiserver-network-proxy/tree/master/examples/kubernetes) in the future though. This is a very helpful feature and it would be nice to provide more visibility. Feel free to do that in a separate PR though.

charleszheng44:
> I see, so we're providing the node IP as the agentAddress and not the agent pod IP. 

We can provide both `--agent-addr="ipv4=podIP,ipv4=nodeIP"`, and on the serverside, we will have a destHost map
map[string]conn{"<podIP>": agentConn, "<nodeIP>": agentConn}. If the server receive a request that uses podIP as the destination host, then the "<podIP>:agentConn" entry will be chosen. 

> Sorry just a follow up question, are any changes needed in the kubernetes repo? 

No change need to be made in the kube repo, supporting extra addresses, like node IP, is just a plus not a mandatory. 

> I'm not sure if any control plane -> cluster requests route to the node IP by default.

Yes, for my knowledge,  there is no `control plane -> cluster` requests that use on the node IP. What I am thinking is that in the future, may be other components that uses node IP as the request dest, like Prometheus or metric servers, would leverage ANP as well. :smile: 

> This is a very helpful feature and it would be nice to provide more visibility. 

I will submit an independent PR for the example and doc. This PR is a little bit large :grin:

Jefftree:
> We can provide both `--agent-addr="ipv4=podIP,ipv4=nodeIP"`, and on the serverside, we will have a destHost map
> `map[string]conn{"<podIP>": agentConn, "<nodeIP>": agentConn}`. If the server receive a request that uses podIP as the destination host, then the `"<podIP>:agentConn"` entry will be chosen

Up to you :). I think it's also fine to just put one IP address as well as long as it's clearly documented what address to use. 

An interesting note, kubernetes SSH tunnels have a hack when routing requests to check the kubelet IP and try to optimize the routing to hit the same destination node (https://github.com/kubernetes/kubernetes/blob/master/pkg/ssh/ssh.go#L399-L404). Obviously, the implementation here in ANP is much more extensible. However, since SSH tunnels already perform this optimization, I'm thinking it might not be a bad idea to set destHost as the default method and fallback to selecting a random backend if the IP doesn't match. This can help avoid fragmentation in the configs, and without the cidr and failure zone filled in, this is more of a routing optimization rather than feature.
 
cc @cheftako

pkg/agent/client.go
```diff
@@ -79,6 +80,58 @@ func newConnectionManager() *connectionManager {
...
+// GenAgentAddress generates an AgentAddress based on the input string, the
+// input string should be a comma-seprated list with each item in the format
+// of <addressType>=<address>
+func GenAgentAddress(addrs string) AgentAddress {
+	var agentAddr AgentAddress
+	entries := strings.Split(addrs, ",")
+	for _, entry := range entries {
+		kv := strings.Split(entry, "=")
+		if len(kv) != 2 {
+			continue
```
Jefftree, line 111: We should at least print a warning/error if the configuration is incorrect

charleszheng44: warning message added

cmd/agent/main.go
```diff
@@ -70,6 +72,7 @@ type GrpcProxyAgentOptions struct {
 	adminServerPort  int
 
 	agentID       string
+	agentAddress  string
```
cheftako, line 75: Quibble: I would change this to agentIdentifiers as its more than just the address at this point and it can encode multiple identifiers.

cheftako: Or maybe characteristics or labels.

cmd/server/main.go
```diff
@@ -135,6 +143,7 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.agentServiceAccount, "agent-service-account", o.agentServiceAccount, "Expected agent's service account during agent authentication (used with agent-namespace, authentication-audience, kubeconfig).")
 	flags.StringVar(&o.kubeconfigPath, "kubeconfig", o.kubeconfigPath, "absolute path to the kubeconfig file (used with agent-namespace, agent-service-account, authentication-audience).")
 	flags.StringVar(&o.authenticationAudience, "authentication-audience", o.authenticationAudience, "Expected agent's token authentication audience (used with agent-namespace, agent-service-account, kubeconfig).")
+	flags.StringVar(&o.proxyStrategies, "proxy-strategies", o.proxyStrategies, "The list of proxy strategies used by the server. e.g., \"destHost,default\"")
```
cheftako, line 146: The list of proxy strategies used by the server to *pick a backend/tunnel.* e.g., \"destHost,default\"

cmd/server/main.go
```diff
@@ -160,6 +169,7 @@ func (o *ProxyRunOptions) Print() {
 	klog.V(1).Infof("AgentServiceAccount set to %q.\n", o.agentServiceAccount)
 	klog.V(1).Infof("AuthenticationAudience set to %q.\n", o.authenticationAudience)
 	klog.V(1).Infof("KubeconfigPath set to %q.\n", o.kubeconfigPath)
+	klog.V(1).Infof("ProxyStrategies set to %s.\n", o.proxyStrategies)
```
cheftako, line 172: Please use %q

cmd/agent/main.go
```diff
@@ -101,6 +105,7 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.DurationVar(&o.syncInterval, "sync-interval", o.syncInterval, "The initial interval by which the agent periodically checks if it has connections to all instances of the proxy server.")
 	flags.DurationVar(&o.probeInterval, "probe-interval", o.probeInterval, "The interval by which the agent periodically checks if its connections to the proxy server are ready.")
 	flags.StringVar(&o.serviceAccountTokenPath, "service-account-token-path", o.serviceAccountTokenPath, "If non-empty proxy agent uses this token to prove its identity to the proxy server.")
+	flags.StringVar(&o.agentAddress, "agent-addr", o.agentAddress, "Addresses that are reachable through this agent. e.g.,host=localhost,cidr=127.0.0.1/16,ipv6=:::::,failure-zone=us-central1-b.")
```
cheftako, line 108: Might be worth documenting you support multiple values by repeating keys. Eg host=localhost,host=node1.mydomain.com

pkg/server/backend_manager.go
```diff
@@ -27,6 +28,29 @@ import (
...
 
+type ProxyStrategy string
+
+const (
+	ProxyStrategyDestHost ProxyStrategy = "destHost"
+	ProxyStrategyDefault  ProxyStrategy = "default"
+)
+
+// GenProxyStrategiesFromStr generates the list of proxy strategies from the
+// comma-seperated string, i.e., destHost,defualt.
```
cheftako, line 39: default

pkg/server/backend_manager.go
```diff
@@ -27,6 +28,29 @@ import (
...
+	strs := strings.Split(proxyStrategies, ",")
+	for _, s := range strs {
+		switch s {
+		case string(ProxyStrategyDestHost):
+			ps = append(ps, ProxyStrategyDestHost)
+		default:
+		}
+	}
+	// always append the default backendmanager as the last one
+	return append(ps, ProxyStrategyDefault)
```
cheftako, line 51: We document above that you can explicitly place default in the strategies config. To me that means if I don't include it then it should be absent (which seems correct). However that means it should be in the switch statement. It is then possible to not have a default strategy if that is what you want.

cmd/server/main.go
```diff
@@ -301,6 +311,7 @@ func newProxyRunOptions() *ProxyRunOptions {
 		agentServiceAccount:       "",
 		kubeconfigPath:            "",
 		authenticationAudience:    "",
+		proxyStrategies:           "default",
```
cheftako, line 314: We should validate proxyStrategies (eg make it hasn't been configured to empty and the strategies specified are all recognized.)

pkg/server/server.go
```diff
@@ -141,6 +145,90 @@ var _ agent.AgentServiceServer = &ProxyServer{}
...
+				klog.V(5).InfoS("Add the agent to DestHostBackendManager", "agent address", ipv6)
+				s.BackendManagers[i].AddBackend(ipv6, conn)
+			}
+			for _, host := range agentAddress.Host {
+				klog.V(5).InfoS("Add the agent to DestHostBackendManager", "agent address", host)
+				s.BackendManagers[i].AddBackend(host, conn)
+			}
+		default:
+			klog.V(5).InfoS("Add the agent to DefaultBackendManager", "agentID", agentID)
+			backend = s.BackendManagers[i].AddBackend(agentID, conn)
```
cheftako, line 199: agentID is the universal ID for the agent and shouldn't be conflated with the other IDs. As an example of default backend manager should probably just ignore anything which is not the universal ID. I would also specify and track the "type" of the id into the backend manager. That way its easy to write a backend manager which only care about something like exact IP match and will not confuse an IP with anything else.

cmd/agent/main.go
charleszheng44: change the `agentAddress` to `agentIdentifiers`

pkg/server/backend_manager.go
charleszheng44: That makes sense. I have removed the default strategy. And the `DefaultBackendManager` will always be added to the end of `BM` list when creating a new proxy server

pkg/server/server.go
charleszheng44: To track the identifier type in BackendManager, I added a new field `idTypes` to the `DefaultBackendStorage`, which contains a list of valid identifier types. When calling `AddBackend`, the Identifiertype will also need to be passed in to the function, if the identifier type is not included in the `idTypes`, the BackendStorage will not add it to the backend list.

pkg/agent/client.go
```diff
@@ -79,6 +80,62 @@ func newConnectionManager() *connectionManager {
 	}
 }
 
+// AgentIdentifiers stores identifiers that will be used by the server when
+// choosing agents
+type AgentIdentifiers struct {
+	IPv4        []string
+	IPv6        []string
+	Host        []string
+	CIDR        []string
```
Jefftree, line 89: @cheftako I understand eventually we want to support CIDR and FailureZone options, but this isn't currently used in the PR. Can we remove it for now to avoid misuse?

Supporting these options should be done on the proxy server/kube-apiserver side. I'm not sure if an agent would even be able to provide this information without querying KAS.

Jefftree (commented):
Hi @charleszheng44. Would it be possible to create an integration test with the desthost backend manager to demonstrate the routing? I think you should be able to create two agents that route to two different test servers, and check that requests to a specific server only go through a specific agent. https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/tests/concurrent_test.go might be a good reference.

charleszheng44:
@Jefftree Yes. I can make an integration test, but will that make this PR too large?  What about having an independent PR just for test and konnectivity example?

Jefftree:
Sure, separate PR sounds good.

pkg/agent/client.go
cheftako: Sure. Though I would guess @charleszheng44 may want CIDR sooner rather than later. Pods usually have their IP assigned from the CIDR range for a given node. So if you want to ensure your requests are sent to exactly the right node and you have requests for pods then you will need to do a CIDR match.

Jefftree: Oh, so CIDR range is something that can be directly inferred from a pod IP? If that's the case this seems like a pretty big performance improvement since we can almost guarantee that every request routes to the exact node that contains a pod?

charleszheng44: @Jefftree Yes. CIDR can be quite useful. For example, the proxy server receives a request with dest IP 192.168.0.1, and there exists an agent with CIDR 192.168.0.0/16 as one of its identifiers, then the server will forward the request to the agent. 

Let's add the FailureZone in a seperate PR. I am still wondering how to pass the FailureZone information through a request?  Does pod/node in different FailureZone can have same IP? We can open an issue to discusses this.

pkg/server/backend_manager.go
```diff
@@ -18,15 +18,39 @@ package server
...
+// GenProxyStrategiesFromStr generates the list of proxy strategies from the
+// comma-seperated string, i.e., destHost.
+func GenProxyStrategiesFromStr(proxyStrategies string) []ProxyStrategy {
+	var ps []ProxyStrategy
+	strs := strings.Split(proxyStrategies, ",")
+	for _, s := range strs {
+		switch s {
+		case string(ProxyStrategyDestHost):
+			ps = append(ps, ProxyStrategyDestHost)
+		default:
```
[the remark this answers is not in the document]
charleszheng44: warning added

pkg/agent/client.go
charleszheng44: Remove the FailureZone IdentifierType for now.

pkg/server/backend_manager.go
charleszheng44: The default strategy will always be added. And we won't let the user specify default anymore.

pkg/server/backend_manager.go
```diff
@@ -61,7 +86,7 @@ func newBackend(conn agent.AgentService_ConnectServer) *backend {
 // connections, i.e., get, add and remove
 type BackendStorage interface {
 	// AddBackend adds a backend.
-	AddBackend(agentID string, conn agent.AgentService_ConnectServer) Backend
+	AddBackend(identifier string, idType pkgagent.IdentifierType, conn agent.AgentService_ConnectServer) Backend
 	// RemoveBackend removes a backend.
 	RemoveBackend(agentID string, conn agent.AgentService_ConnectServer)
```
Jefftree, line 91: The RemoveBackend function now accepts more than just agentIDs right? Should the function signature mirror the AddBackend function?

Jefftree (commented):
Hi @charleszheng44. Thank you for your work on the BackendManager list. I have one small comment otherwise LGTM.

The new file `pkg/server/desthost_backend_manager.go` also needs a 2020 Kubernetes Authors copyright block.

pkg/server/backend_manager.go
charleszheng44: Change the functin signature of `RemoveBackend` to mirro the `AddBackend`

charleszheng44:
@Jefftree The copyright block is added and I have changed the function signature of RemoveBackend based on your comment.

pkg/agent/client.go
```diff
@@ -79,6 +80,58 @@ func newConnectionManager() *connectionManager {
...
+	IPv6 IdentifierType = "ipv6"
+	Host IdentifierType = "host"
+	CIDR IdentifierType = "cidr"
+	UID  IdentifierType = "uid"
+)
+
+// GenAgentIdenfiers generates an AgentIdentifiers based on the input string, the
+// input string should be a comma-seprated list with each item in the format
+// of <IdentifierType>=<address>
+func GenAgentIdenfiers(addrs string) AgentIdentifiers {
```
Jefftree, line 105: s/Idenfiers/Identifiers

Jefftree:
Thanks! lgtm from me. I feel like some of the backendstorage code may need to be changed as we add additional managers (especially https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/155), but it's in a good state at the moment.

I'll defer to @caesarxuchao and @cheftako for approval.

cmd/agent/main.go
```diff
@@ -101,6 +105,7 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.DurationVar(&o.syncInterval, "sync-interval", o.syncInterval, "The initial interval by which the agent periodically checks if it has connections to all instances of the proxy server.")
 	flags.DurationVar(&o.probeInterval, "probe-interval", o.probeInterval, "The interval by which the agent periodically checks if its connections to the proxy server are ready.")
 	flags.StringVar(&o.serviceAccountTokenPath, "service-account-token-path", o.serviceAccountTokenPath, "If non-empty proxy agent uses this token to prove its identity to the proxy server.")
+	flags.StringVar(&o.agentIdentifiers, "agent-identifiers", o.agentIdentifiers, "Identifiers of the agent that will be used by the server when choosing agent. e.g.,host=localhost,host=node1.mydomain.com,cidr=127.0.0.1/16,ipv4=1.2.3.4,ipv4=5.6.7.8,ipv6=:::::")
```
cheftako, line 108: As is we have ',' and '=' as reserved characters which cannot be present in the identifiers. If we encode the value/identifier, then it is a bit more complicated to set a value but that restrictions are removed. We would need to define the encoder. 

URLEncode (https://play.golang.org/p/Zx0wMq7oacr) makes things fairly easy though it would switch us to using '&' and '='. Your example would then be "cidr=127.0.0.1%2F16&host=localhost&host=node1.mydomain.com&ipv4=1.2.3.4&ipv4=5.6.7.8&ipv6=%3A%3A%3A%3A%3A"

cmd/agent/main.go
```diff
@@ -116,6 +121,7 @@ func (o *GrpcProxyAgentOptions) Print() {
 	klog.V(1).Infof("SyncInterval set to %v.\n", o.syncInterval)
 	klog.V(1).Infof("ProbeInterval set to %v.\n", o.probeInterval)
 	klog.V(1).Infof("ServiceAccountTokenPath set to %q.\n", o.serviceAccountTokenPath)
+	klog.V(1).Infof("AgentIdentifiers set to %s.\n", o.agentIdentifiers)
```
cheftako, line 124: Especially if we decide to encode the data I would suggest we print each key=value pair separately (& decoded).

pkg/server/backend_manager.go
```diff
@@ -105,67 +132,93 @@ type DefaultBackendStorage struct {
...
+		}
 	}
+	return false
 }
 
 // AddBackend adds a backend.
-func (s *DefaultBackendStorage) AddBackend(agentID string, conn agent.AgentService_ConnectServer) Backend {
+func (s *DefaultBackendStorage) AddBackend(identifier string, idType pkgagent.IdentifierType, conn agent.AgentService_ConnectServer) Backend {
+	if !containIdType(s.idTypes, idType) {
+		klog.ErrorS(&ErrWrongIDType{idType, s.idTypes}, "fail to add backend")
```
cheftako, line 171: This seems an alarming and misleading error to log. As I understand it, it is expected that we would call AddBackend on every IdentifierType an agent presents for each BackendStorage. This means we expect some of those identifiers to not match all BackendStorage and these error message are expected and do not (normally) indicate an actual error. We should either advertise the supported backends and filter this call to prevent this from being expected. Or we should fix the level V(4)? and probably adjust the message to be less alarming "backend not using unsupported type"

pkg/server/backend_manager.go
```diff
@@ -105,67 +132,93 @@ type DefaultBackendStorage struct {
...
+		return nil
+	}
 	klog.V(2).InfoS("Register backend for agent", "connection", conn, "agentID", agentID)
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	_, ok := s.backends[agentID]
+	_, ok := s.backends[identifier]
```
cheftako, line 177: At this point we have lost the id type, which means there is at least theoretically the possibility of a collision across types. Good news is we can always fix this later.

charleszheng44: will try to fix this issue in the next PR.

cmd/agent/main.go
charleszheng44: I agree that it would be better if we don't need to reserve any character. But does that mean users need to pass the argument in the format of "cidr=127.0.0.1%2F16&ipv6=%3A%3A%3A%3A%3A"? Will that be a little bit user-unfriendly.

cmd/agent/main.go
charleszheng44: Will print out each identifier separately.

pkg/server/backend_manager.go
charleszheng44: Will switch to `klog.V(4).Infof`

cmd/agent/main.go
cheftako: There is also information which is probably category 1) but is not directly tied to the agent. So for instance things like failure zone are characteristics of the Node that the agent is running in. It can be looked up when the agent is started but its not clear to me that is when/how we will obtain that data. Its also possible it may be looked up by the (appropriate) backend manager when the agent is registered.

cmd/agent/main.go
cheftako: https://play.golang.org/p/3E93xZ-i11n shows that the library will pass through unencoded parameters so your "cidr=127.0.0.1/16&host=localhost&host=node1.mydomain.com&ipv4=1.2.3.4&ipv4=5.6.7.8&ipv6=:::::" will also work. I would still document the "correct" form but you can mention that simple form also works. If you mention it though I would add a unit test to make sure that a later change to the library does not break our customers.

pkg/server/backend_manager.go
```diff
@@ -168,18 +168,18 @@ func containIdType(idTypes []pkgagent.IdentifierType, idType pkgagent.Identifier
 // AddBackend adds a backend.
 func (s *DefaultBackendStorage) AddBackend(identifier string, idType pkgagent.IdentifierType, conn agent.AgentService_ConnectServer) Backend {
 	if !containIdType(s.idTypes, idType) {
-		klog.ErrorS(&ErrWrongIDType{idType, s.idTypes}, "fail to add backend")
+		klog.V(4).InfoS("fial to add backend", "backend", identifier, "error", &ErrWrongIDType{idType, s.idTypes})
```
Jefftree, line 171: s/fial/fail

go.mod
```diff
@@ -18,6 +18,7 @@ require (
 	k8s.io/apimachinery v0.17.1
 	k8s.io/client-go v0.17.1
 	k8s.io/klog/v2 v2.0.0
+	rsc.io/quote/v3 v3.1.0 // indirect
```
cheftako, line 21: I believe this means we need to ensure we are including the https://pkg.go.dev/rsc.io/quote/v3?tab=licenses.

pkg/server/server.go
```diff
@@ -202,16 +290,28 @@ func (s *ProxyServer) getFrontendsForBackendConn(agentID string, backend Backend
...
+func NewProxyServer(serverID string, proxyStrategies []ProxyStrategy, serverCount int, agentAuthenticationOptions *AgentTokenAuthenticationOptions) *ProxyServer {
+	var bms []BackendManager
+	for _, ps := range proxyStrategies {
+		switch ps {
+		case ProxyStrategyDestHost:
+			bms = append(bms, NewDestHostBackendManager())
+		default:
+		}
+	}
+	bms = append(bms, NewDefaultBackendManager())
```
cheftako, line 302: Seems like the default behavior (i.e. no proxy strategies provided) should be to just have the default. However if you *do* list strategies I think it is legitimate to leave out the default manager. Essentially saying its not ok to attempt to use a random backend.

Jefftree: I actually like the original cascading behavior where a fallback (the default manager) is provided if the user selected manager fails in some way. I don't think including this should cause any problems, unless if in the case that the selected manager fails, we want to fail the routing instead of finding a different path?

cheftako: I would agree that if no strategy is provided that using the default is correct. I would also agree that most users would want the default as a fallback. However that could be accomplished by setting (for example) "--proxy-strategies=destHost,default". This issue is that is nodes do not have inter-communication or worse yet have it but its not desirable, then you cannot avoid it. If we allow default to be disabled (eg. by setting "--proxy-strategies=destHost") then whoever is provisioning the cluster can choose the right behavior for them.

go.mod
charleszheng44: where should be put the license?

pkg/server/server.go
charleszheng44: I agree that we should allow users to disable the default proxy strategy if they know exactly what proxy strategy they wanna use. 

What about if no strategies are provided, i.e., "--proxy-strategies" is not specified, the default strategy will be adopted. But if the user specifies any proxy strategies without including the default strategy, e.g. "--proxy-strategies=destHost", then we won't fall back to default, if no backends are available.

Jefftree: Ahh thanks for the explanation. +1 from me. This is an elegant way for users to decide if they want to incorporate fallback strategies or not.

go.mod
cheftako: Ideally we should have a top level directory LICENSES and a script which populated the directory for things like this. For now lets file a bug and I'll see if we can add that easily as a follow on.

pkg/server/server.go
cheftako: So I think we are all in agreement about the desired behavior. However I think to implement that we would need to add a strategy name for the default backend manager. (I'd be fine with "default"). Then this addition would occur if that was specified not in all cases as this code shows.

pkg/server/backend_manager.go
charleszheng44: fixed

go.mod
charleszheng44: will file an issue.

cmd/server/main.go
```diff
@@ -141,6 +150,7 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.agentServiceAccount, "agent-service-account", o.agentServiceAccount, "Expected agent's service account during agent authentication (used with agent-namespace, authentication-audience, kubeconfig).")
 	flags.StringVar(&o.kubeconfigPath, "kubeconfig", o.kubeconfigPath, "absolute path to the kubeconfig file (used with agent-namespace, agent-service-account, authentication-audience).")
 	flags.StringVar(&o.authenticationAudience, "authentication-audience", o.authenticationAudience, "Expected agent's token authentication audience (used with agent-namespace, agent-service-account, kubeconfig).")
+	flags.StringVar(&o.proxyStrategies, "proxy-strategies", o.proxyStrategies, "The list of proxy strategies used by the server to pick a backend/tunnel. e.g., \"destHost,\"")
```
cheftako, line 153: Not blocking but would be nice if the tool tip listed the valid strategy values.

cmd/server/main.go
```diff
@@ -282,6 +293,19 @@ func (o *ProxyRunOptions) Validate() error {
...
 
+	// validate the proxy strategies
+	if o.proxyStrategies != "" {
+		pss := strings.Split(o.proxyStrategies, ",")
+		for _, ps := range pss {
+			switch ps {
+			case string(server.ProxyStrategyDestHost):
+			case string(server.ProxyStrategyDefault):
+			default:
+				return fmt.Errorf("unknown proxy strategy: %s, available strategy is: destHost", ps)
```
cheftako, line 304: Again not blocking but would be nice if the error message listed all the valid strategy values.

pkg/server/backend_manager.go
```diff
@@ -18,15 +18,43 @@ package server
...
 	client "sigs.k8s.io/apiserver-network-proxy/konnectivity-client/proto/client"
+	pkgagent "sigs.k8s.io/apiserver-network-proxy/pkg/agent"
 	"sigs.k8s.io/apiserver-network-proxy/proto/agent"
 )
 
+type ProxyStrategy string
+
+const (
+	ProxyStrategyDefault  ProxyStrategy = "default"
+	ProxyStrategyDestHost ProxyStrategy = "destHost"
```
cheftako, line 37: Not blocking but we should really document what each proxy strategy means. Eg.
"default": With this strategy the Proxy Server will randomly pick a backend from the current healthy backends to establish the tunnel overwhich to forward requests.

cheftako:
Seems pretty good to me. Waiting on @Jefftree review :+1:

cmd/server/main.go
charleszheng44: list all supported strategies in the error message

pkg/server/backend_manager.go
charleszheng44: add a description for each proxy strategy

cmd/server/main.go
charleszheng44: list all the valid strategy values

Jefftree:
Please don't forget to add a follow up to test the desthost manager routing.

LGTM from me :)

cheftako:
/approve
@charleszheng44 thanks for your hard work on this. I realize the review on this was arduous but I think you created a great feature!

charleszheng44:
@cheftako @Jefftree will this PR be merged automatically?

Jefftree:
@charleszheng44 Reiterating Walter's point, thank you so much for adding this PR! I really look forward to using it and seeing how it evolves in the future!

/lgtm
/unhold

PR #148, opened 2020-10-09
konnectivity-server should make the socket group readable and writeab…

vinayakankugoyal:
…le when it listens on the socket. Most systems have a default umask of 022, meaning when konnectivity-server creates the socket only the user that it is running as will have read-write permissions to the socket. This means that any other process that needs to read-write to the socket needs to be run as the same user as konnectivity-server.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -29,6 +29,7 @@ import (
 	"os"
 	"os/signal"
 	"runtime"
+	"sync"
 	"syscall"
 
 	"github.com/google/uuid"
@@ -45,6 +46,8 @@ import (
 	"sigs.k8s.io/apiserver-network-proxy/pkg/util"
 	"sigs.k8s.io/apiserver-network-proxy/proto/agent"
 )
+
+var udsListenerLock sync.Mutex
 
 func main() {
 	// flag.CommandLine.Parse(os.Args[1:])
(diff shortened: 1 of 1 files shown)


cmd/server/main.go
```diff
@@ -415,10 +415,12 @@ func (p *Proxy) runUDSMasterServer(ctx context.Context, o *ProxyRunOptions, s *s
 		grpcServer := grpc.NewServer()
 		client.RegisterProxyServiceServer(grpcServer, s)
 		var lc net.ListenConfig
+		oldUmask := syscall.Umask(0007)
 		lis, err := lc.Listen(ctx, "unix", o.udsName)
 		if err != nil {
 			return nil, fmt.Errorf("failed to listen(unix) name %s: %v", o.udsName, err)
 		}
+		syscall.Umask(oldUmask)
```
caesarxuchao, line 423: This isn't going to be reached if the Listen fails.

caesarxuchao (commented):
Thanks, @vinayakankugoyal.

For future reference, this is the recommended way of setting UDS file descriptor's file permission, according to https://github.com/golang/go/issues/11822

cmd/server/main.go
vinayakankugoyal: :man_facepalming: Fixed. Thanks for pointing this out.

caesarxuchao:
lgtm. Please squash the commits.

vinayakankugoyal:
> lgtm. Please squash the commits.

Done. Thanks!

caesarxuchao:
LGTM.

I'll defer it to Walter to apply the label.

cmd/server/main.go
```diff
@@ -431,7 +433,9 @@ func (p *Proxy) runUDSMasterServer(ctx context.Context, o *ProxyRunOptions, s *s
 		stop = func() { server.Shutdown(ctx) }
 		go func() {
 			var lc net.ListenConfig
+			oldUmask := syscall.Umask(0007)
```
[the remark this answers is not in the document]
vinayakankugoyal: Strongly agree we should do what you suggested. Updated the code PTAL.

cmd/server/main.go
```diff
@@ -397,6 +400,18 @@ func SetupSignalHandler() (stopCh <-chan struct{}) {
 	return stop
 }
 
+func getUDSListener(ctx context.Context, udsName string) (net.Listener, error) {
+	udsListenerLock.Lock()
+	defer udsListenerLock.Unlock()
+	defer syscall.Umask(syscall.Umask(0007))
```
cheftako, line 406: Seems a bit sneaky to rely on parameter evaluation of a deferred function to get the first Umask call to happen before the rest of the function. Going to lgtm but we should consider either comments or breaking it out to make it a little more obvious to anyone reading the code.

vinayakankugoyal: Broke it out to make the code more obvious. Thanks!

Issue #149, opened 2020-10-13
The proxy server can't read from frontend


charleszheng44:
Tried to run the examples in the readme with the latest master, and kept getting the following error message:
```bash
E1012 18:09:16.558653   63575 server.go:245] "Stream read from frontend failure" err="rpc error: code = Canceled desc
```
I have checked out some previous commits, and I think the issue is introduced by PR #112.

@cheftako @mvladev @Jefftree

charleszheng44:
@cheftako @mvladev @Jefftree

cheftako:
#112 Rephrased the message a little and bumped it from a warning to a full error. As far as I can tell #112 doesn't make any changes to the programmatic logic. Will attempt to pull the code and try it out. Was there a particular sample config in the readmes you were seeing this on?

charleszheng44:
The error message is harmless as the tunnel being shut down. To show the html text, please run the proxy-test-client with "-v 4".

Jefftree:
Does it make sense to add a warning message that this is may be normal in certain cases? Generally when the frontend or backend closes the connection, the other side will always receive some kind of error that the stream was abruptly terminated, but technically it shouldn't be an error?

cheftako:
1) The new log levels are "hiding" the client output. If you add a "--v=4" flag to the ./bin/proxy-test-client command you will see things appear to be working correctly. I'm seeing the http request from the client getting to the server and the html page being successfully returned. Might be good to set a default log level on the client to avoid having to specify it on the command line.

2) We seem to have both agent and server listening on port 8093 by default. That may be fine in Production but is a bit of an issue when running both on the same box. I had to add a "--health-server-port=8095" flag to the proxy-agent to work around the issue. At a minimum we should fix the README to get around this. Might be nice to work out if we can fix one the default ports if its not too late to avoid the conflict.

3) I'm actually seeing 2 unhappy looking error messages. The aforementioned "E1012 23:26:45.651803  140242 server.go:245] "Stream read from frontend failure" err="rpc error: code = Canceled desc = context canceled" on the proxy-server and "1012 23:26:45.652099  141738 client.go:391] "connection read failure" err="read tcp 127.0.0.1:33550->127.0.0.1:8000: use of closed network connection" on the proxy agent. I believe these are harmless "expected" errors based on a tunnel being shut down. I agree we should probably lower both of them from error to info level 2 and reword them to indicate a tunnel is being shutdown.

PR #150, opened 2020-10-13
Defaulted test client log level to 4.

cheftako:
Defaulted test client log level to 4.

Helping to address concerns raised in #149 .

### DIFF
--- a/cmd/client/main.go
+++ b/cmd/client/main.go
@@ -47,6 +47,7 @@ func main() {
 	flags.AddFlagSet(o.Flags())
 	local := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	klog.InitFlags(local)
+	local.Set("v", "4")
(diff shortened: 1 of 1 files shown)


Jefftree:
Since klog sets a verbosity of 0 by default, does it make sense to only update our sample code that invokes `./proxy-test-client` to include `--v=4` instead?

cheftako:
> Since klog sets a verbosity of 0 by default, does it make sense to only update our sample code that invokes `./proxy-test-client` to include `--v=4` instead?

The primary use case I was trying to fix was how it behaves when run from the command line. We could fix all our documentation to explicitly tell people to add that flag. However I think it makes more sense for the defaults to minimize the number of things you need to type in. In addition as a test tool we don't have to worry as much about backward compatibility.

Jefftree:
Sounds good.
/lgtm

Issue #152, opened 2020-10-15
Server tries to route traffic through "hard" disconnected agents


Avatat:
Hello,
I have Kubernetes 1.18 cluster where Konnectivity v0.0.12 is running.
Most of the time, everything works perfectly, but after ~20 days of konnectivity-server uptime, some DIAL_REQ are getting nowhere.
For example, `kubectl logs -n kube-system konnectivity-agent-h6bt9 -f` only worked after the third execution.

Server log:
```
I1015 11:35:55.013608       1 server.go:224] proxy request from client, userAgent [grpc-go/1.26.0]
I1015 11:35:55.013887       1 server.go:257] start serving frontend stream
I1015 11:35:55.013926       1 server.go:268] >>> Received DIAL_REQ
I1015 11:35:55.013940       1 backend_manager.go:170] pick agentID=f42e628f-f3a5-42cc-bb75-8065768b1be1 as backend
I1015 11:35:55.014079       1 server.go:290] >>> DIAL_REQ sent to backend

I1015 11:36:10.378211       1 server.go:224] proxy request from client, userAgent [grpc-go/1.26.0]
I1015 11:36:10.378345       1 server.go:257] start serving frontend stream
I1015 11:36:10.378357       1 server.go:268] >>> Received DIAL_REQ
I1015 11:36:10.378363       1 backend_manager.go:170] pick agentID=f42e628f-f3a5-42cc-bb75-8065768b1be1 as backend
I1015 11:36:10.378418       1 server.go:290] >>> DIAL_REQ sent to backend

I1015 11:36:18.243790       1 server.go:224] proxy request from client, userAgent [grpc-go/1.26.0]
I1015 11:36:18.243873       1 server.go:257] start serving frontend stream
I1015 11:36:18.244049       1 server.go:268] >>> Received DIAL_REQ
I1015 11:36:18.244062       1 backend_manager.go:170] pick agentID=15fb1f4a-55a5-454d-9257-a717e70dc5dd as backend
I1015 11:36:18.244138       1 server.go:290] >>> DIAL_REQ sent to backend
I1015 11:36:18.247179       1 server.go:522] <<< Received DIAL_RSP(rand=2352227733928774978), agentID 15fb1f4a-55a5-454d-9257-a717e70dc5dd, connID 84)
[...]

PR #153, opened 2020-10-16
Support multiple grpc dial options for agent -> server connection

Jefftree:
grpc supports multiple dial options but currently agent -> server dials (cmd/agent/main.go) only supports 1 dial option as a parameter. Refactor the code to support multiple.

This might be helpful for https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/152

/assign @cheftako

### DIFF
--- a/pkg/agent/clientset.go
+++ b/pkg/agent/clientset.go
@@ -43,7 +43,7 @@ type ClientSet struct {
 	// proxy server.
 	probeInterval time.Duration // The interval by which the agent
 	// periodically checks if its connections to the proxy server is ready.
-	dialOption grpc.DialOption
+	dialOptions []grpc.DialOption
 	// file path contains service account token
 	serviceAccountTokenPath string
 	// channel to signal shutting down the client set. Primarily for test.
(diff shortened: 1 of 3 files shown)


Avatat:
I sit to implementing keepalive for the agent, and now I understand your change ;)
I will wait for a merge and then start implementing the keepalive option in the agent.

pkg/agent/clientset.go
```diff
@@ -43,7 +43,7 @@ type ClientSet struct {
 	// proxy server.
 	probeInterval time.Duration // The interval by which the agent
 	// periodically checks if its connections to the proxy server is ready.
-	dialOption grpc.DialOption
+	dialOption []grpc.DialOption
```
cheftako, line 46: Slightly more work but can we name it dialOtions to hint that its a slice?

pkg/agent/clientset.go
```diff
@@ -110,7 +110,7 @@ type ClientSetConfig struct {
 	AgentID                 string
 	SyncInterval            time.Duration
 	ProbeInterval           time.Duration
-	DialOption              grpc.DialOption
+	DialOption              []grpc.DialOption
```
cheftako, line 113: And here? Public so requires a bit more thought.

pkg/agent/clientset.go
Jefftree: Updated.

pkg/agent/clientset.go
Jefftree: Should be no problem renaming this since the agent is a standalone binary and is not imported anywhere.

PR #154, opened 2020-10-20
Add keepalive option to agent server

Avatat:
This is a fix to issue #152.

Sometimes, agent can disconnect without closing connection, and server don't know about it. Disconnected agent is still on the available agents list and server tries to route traffic through it.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -31,6 +31,7 @@ import (
 	"runtime"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/google/uuid"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -38,6 +39,7 @@ import (
 	"github.com/spf13/pflag"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/keepalive"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
@@ -92,6 +94,9 @@ type ProxyRunOptions struct {
 	adminPort uint
 	// Port we listen for health connections on.
 	healthPort uint
+	// After a duration of this time if the server doesn't see any activity it
+	// pings the client to see if the transport is still alive.
+	keepaliveTime time.Duration
 	// Enables pprof at host:adminPort/debug/pprof.
 	enableProfiling bool
 	// If enableProfiling is true, this enables the lock contention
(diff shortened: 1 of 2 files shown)


Avatat:
I signed CLA.

Avatat:
@Jefftree, should I write a test, which will cover a bug from #152?
To be honest, I have no idea how to do this, where to begin.

cmd/server/main.go
```diff
@@ -541,8 +543,11 @@ func (p *Proxy) runAgentServer(o *ProxyRunOptions, server *server.ProxyServer) e
 	}
 
 	addr := fmt.Sprintf(":%d", o.agentPort)
-	serverOption := grpc.Creds(credentials.NewTLS(tlsConfig))
-	grpcServer := grpc.NewServer(serverOption)
+	serverOptions := []grpc.ServerOption {
+		grpc.Creds(credentials.NewTLS(tlsConfig)),
+		grpc.KeepaliveParams(keepalive.ServerParameters{Time: 30 * time.Second}),
```
cheftako, line 548: This is way too short to be the default. Currently the max K8s cluster size is 15K nodes. For such a cluster you would be sending 500 keep-alive packets per second to each server. I would suggest making the default 1 hour (the default TCP keepalive on the server is 2 hours (so 1/2 that is good) and the default server enforced minimum is 5 minutes).  I'm fine with this being a optional command line flag so you can tune this to you needs. I think we are still likely to also want to mark problematic tunnels as unhealthy but having both seems a good idea.

cheftako: I wonder if we also want to set a keep alive on the agent side.

Avatat: Adding an optional command line flag is a good idea - I will figure it out.

As I understand, we don't need a keepalive mechanism on the agent side, because we currently [check health there](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/3cf58829c4f429fd038b7a5797d112979878060b/pkg/agent/client.go#L436), using [grpc.ClientConn.GetState](https://godoc.org/google.golang.org/grpc#ClientConn.GetState).

cheftako: I *believe* (I could certainly be wrong) that that GetState is mostly tracking if closed has been called. It is not tracking if we have either a network outage such that traffic doesn't route or that the server has gone away without notifying us. These are the conditions which keepalive is meant to help. https://godoc.org/google.golang.org/grpc#WithKeepaliveParams client side keep alive is almost identical to server side keepalive and they should work well together.

Jefftree: Just going to add that SSH tunnels have a healthCheckPoll time of 1 minute (https://github.com/kubernetes/kubernetes/blob/master/pkg/ssh/ssh.go#L316), so it's possible that we could go lower than 1 hour for the keepalive time.

We can probably address client side keep alive in a separate PR.

Avatat: I think so, too. ~1 minute keepalive time isn't something bad for performance - remember, that we send keepalive packets only when the channel is inactive for longer than specified time. In a 15K-node cluster, administrators probably will not deploy konnectivity as daemonset, on every node.
I'm currently working on adding a command line option to allow the administrator to tune this parameter.

After I complete this PR, I would like to try to add keepalive to the agent and then create a new PR :)

Avatat: @cheftako, @Jefftree, I've just added a `--keepalive-time` command-line option to the server code and Kubernetes example.

cmd/server/main.go
```diff
@@ -92,6 +94,9 @@ type ProxyRunOptions struct {
 	adminPort uint
 	// Port we listen for health connections on.
 	healthPort uint
+	// After a duration of this time if the server doesn't see any activity it
+	// pings the client to see if the transport is still alive.
+	keepaliveTime uint
```
Jefftree, line 99: In this repo we've been using the `time.Duration` type for time related variables. (See https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/cmd/agent/main.go#L74). Can we change this to follow the same format?

Avatat: Thank you for pointing this out! It's a more clever way and simplify the other changes. I used `time.Duration` and updated the PR.

PR #155, opened 2020-10-24
Added code to track tainting of a tunnel.

cheftako:
Taint lasts for the tcp keepalive period.
Taint code only works if client is remembering to properly close its
connection.
Tested using ifconfig down to break the connection and prevent tcp close
from the OS. (kill -9 and similar do not work).
This does NOT close tunnels but instead relies on the already
implemented TCP keepalive for that functionality.


pkg/server/server.go
```diff
@@ -238,11 +285,24 @@ func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
...
 				close(stopCh)
 				return
 			}
 			if err != nil {
-				klog.ErrorS(err, "Stream read from frontend failure")
+				if agentID, lookupErr := frontStream.getAgentID(); lookupErr == nil {
+					if frontStream.getStatus() != Closing {
+						klog.ErrorS(err, "Stream read from frontend failure",
+							"agentID", agentID)
+						s.BackendManager.TaintBackend(agentID, err)
```
caesarxuchao, line 297: Why do we taint the backend when the frontend stream read is broken?

caesarxuchao (commented):
Walter, can you remind me what problem this tainting mechanism is trying to solve? IIUC, it tries to avoid picking a backend if it has recently failed. If so, the benefit is saving end users of the network proxy from waiting for the server to detect a GRPC connection to the backend is broken, which could take up to the keepAlive seconds.

The tainting mechanism is quite complex. I don't know if the benefit is worth the complexity.

pkg/server/backend_manager.go
```diff
@@ -99,24 +110,36 @@ type DefaultBackendStorage struct {
...
+	// agentIDs is a slice of agentIDs to enable randomly picking one
+	// in the Backend() method. There is no reliable way to randomly
+	// pick a key from a map (in this case, the backends) in Golang.
 	agentIDs []string
+	// untaintedIDs is a slice of agentIDs which are not currently
+	// marked as tainted to enable randomly picking a known good ID
+	// in the Backend() method. There is no reliable way to randomly
+	// pick a key from a map (in this case, the backends) in Golang.
+	// Consider switching this to a map[string]bool to get set behavior
+	// A little nasty as length where true is yuck.
```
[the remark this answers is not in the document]
cheftako: Golang has no set function, but having one would make things a little easier (prevent issues with readding existing values and throwing of the count). The standard workaround it to use a map[string]bool to get that functionality. We a set length function which would be equivalent to counting the number of keys where the bool is true. Very doable but I decided it was a nice to have right now.

pkg/server/server.go
cheftako: Sadly if the lose the backend to something like a machine crash or network outage we get no error on the backend. Essentially we send TCP packets which disappear into the ether. TCP timeout or even keepalive is generally in minutes and the frontend request timeout is in seconds. So the error shows up as a front ends error (really a timeout but to us it looks like a context closed error). We can't be sure what is wrong, so its better to "taint" the connection than close it. We need to rely on the keepalive options others have added to take care of that. However keepalive can be slow. This helps us to not send requests to a dead connection while we are determining if the connection is actually dead.

cmd/client/main.go
```diff
@@ -250,6 +250,7 @@ func (c *Client) run(o *GrpcProxyClientOptions) error {
 			time.Sleep(wait)
 		}
 	}
+	client.CloseIdleConnections()
```
Jefftree, line 253: What happens if client idle connections are not closed?

pkg/server/server.go
Jefftree: Does this mean hitting any client side timeout results in a backend taint? I've encountered timeouts on the k/k client side and sometimes it's just the destination service being overloaded. This happens a decent amount in the k/k e2e tests and the retry mechanism for most clients takes care of the problem.

I'm worried a problematic destination (eg: webhook) could cause all backends to be tainted for a short period of time.

cmd/client/main.go
cheftako: In my tests it causes us to incorrectly taint tunnels. We then systematically work our way through the list of tunnels/backends until they are all tainted. At that point the code works the way it does today. So its not worse than today but it does mean your not getting any benefit out of tainting tunnels/backends.

Jefftree: The default IdleConnTimeout for golang's http library is 90 seconds, and almost all clients in k/k keep that parameter untouched.

I'm surprised that we get incorrectly tainted tunnels rather than just a really delayed CLOSE_REQ.

cheftako: 90 seconds is a really long time. With our test client (and no CloseIdleConnections call) I was seeing a context closed error on the front end connection after <10 seconds. That may just be the OS cleaning up the connection once the process went away.

Jefftree: > That may just be the OS cleaning up the connection once the process went away.

Ahh I think that's exactly it. I remember adding a sleep interval at the end of of the client previously and it gracefully closed the connection after 90s was hit.

This line addition makes sense but we shouldn't make this assumption for all our clients. It's unlikely for the kube-apiserver process to be terminated so we probably wouldn't run into these context closed errors in k/k too often. However with the 90s idle time, that means the earliest time we'd be able to detect a problem on a connection is also 90s, which is already in the magnitude of minutes.

cheftako: I'd be interested in seeing that happens with webhook calls which have <10 second http timeouts.

pkg/server/server.go
cheftako: All backends getting tainted just reverts us to the current behavior. We lose the benefits of tainting but the system continues to work.

cmd/client/main.go
cheftako: https://github.com/kubernetes/kubernetes/pull/95981 would reduce that in general to 45s. (Still not great). I am also curious to see the behavior with webhook calls as they generally had a sub 10s http timeout.

pkg/server/server.go
caesarxuchao: We can detect broken agent by observing if a dial response is returned by the agent on time. That's less prone to false-positive.

Issue #156, opened 2020-11-10
Default logging level should be higher


Jefftree:
https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/112 helped us reduce the amount of noise within our logs, however the default log level of 0 only logs errors without any other context and isn't very helpful. If the proxy is working correctly without any problems, we will have 0 log lines. We should set the default to something higher like 3 or 4, and users can override it if necessary.

For context, a full run of k/k e2e tests passed all the test cases and yielded this in the logs. https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/96389/pull-kubernetes-e2e-gce-network-proxy-grpc/1325948514512932864/artifacts/e2e-0e724b8666-019b4-master/konnectivity-server.log. Seeing the error messages in isolation is not very helpful.

Something similar to https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/150/files would be ideal.

/cc @cheftako

Jefftree:
Fixed.
/close

PR #157, opened 2020-11-10
Set default logging to v4

Jefftree:
Refer to https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/156

I'm not sure if v4 is the right number, but it should definitely be higher than what it is currently (v0)

V(5) logs the individual packets which is a bit excessive, but V(4) seems to have all the important bits without being too verbose.

/assign @cheftako


cheftako:
/lgtm
/approve
Reviewed the messages at 4 which seem fairly valuable and unlikely to generate insane amounts of logs.
Reviewed the messages at 5 which seem less valuable in most situations and incredibly verbose/common.

PR #159, opened 2020-12-03
Update README.md

daixiang0:
Add missing `kubectl` cmd.

Signed-off-by: Xiang Dai <long0dai@foxmail.com>


Jefftree:
/ok-to-test

Hello, thank you for the PR! From the [static pod docs](https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/), pods running as static pods in the control plane instance are managed directly by the kubelet on the VM. They're usually started by placing the yaml file in a specific directory rather than via a `kubectl apply`.

Obviously, we could also just directly apply this manifest to the API server but I don't think we should make it the default selection. Make you can add a section below showing the steps without using a static pod?

daixiang0:
Done.

PR #162, opened 2020-12-08
send PacketType_CLOSE_REQ to agent when tunnel serve exits 

rambohe-ch:
when the tunnel serve exits, it need to send PacketType_CLOSE_REQ to agent for releasing conn between agent and remote, in order to prevent conn  leakage at the agent.

### DIFF
--- a/pkg/server/tunnel.go
+++ b/pkg/server/tunnel.go
@@ -102,7 +102,21 @@ func (t *Tunnel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	case <-connection.connected: // Waiting for response before we begin full communication.
 	}
 
-	defer conn.Close()
+	defer func() {
+		packet := &client.Packet{
+			Type: client.PacketType_CLOSE_REQ,
+			Payload: &client.Packet_CloseRequest{
+				CloseRequest: &client.CloseRequest{
+					ConnectID: connection.connectID,
+				},
+			},
+		}
+
+		if err = backend.Send(packet); err != nil {
+			klog.V(2).InfoS("failed to send close request packet", "host", r.Host, "agentID", connection.agentID, "connID", connection.connectID)
+		}
+		conn.Close()
+	}()
 
 	klog.V(3).InfoS("Starting proxy to host", "host", r.Host)
 	pkt := make([]byte, 1<<12)


pkg/server/tunnel.go
```diff
@@ -144,5 +144,18 @@ func (t *Tunnel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			"connectionID", connection.connectID)
 	}
 
+	packet := &client.Packet{
```
cheftako, line 147: If we're going to do this it would make sense to have the added code as a separate (anonymous) deferred function after the deferred close call on the defer stack.

pkg/server/tunnel.go
```diff
@@ -144,5 +144,18 @@ func (t *Tunnel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		Type: client.PacketType_CLOSE_REQ,
+		Payload: &client.Packet_CloseRequest{
+			CloseRequest: &client.CloseRequest{
+				ConnectID: connID,
+			},
+		},
+	}
+
+	if err = backend.Send(packet); err != nil {
+		klog.ErrorS(err, "error sending close request packet", "host", r.Host, "agentID", agentID, "connID", connID)
```
cheftako, line 157: Most of the time if you get here it's because the relevant backend is already closed. As such this should not be an ErrorS. V(4).InfoS at most.

pkg/server/tunnel.go
Jefftree: +1

pkg/server/tunnel.go
Jefftree: The destination may have its connection closed but sending a CLOSE_REQ to a backend should never fail unless the agent goes away (or the network connection is broken). If that's the case, then I think the ErrorS log makes sense?

cheftako: For us to get here, a break had to happen earlier, that suggests we either got a EOF or an earlier Send received an error and was logged with a ErrorS. I would rather just see that message as it pertains to the actual problem. This seems more likely to mislead than help. In the EOF case it suggests we may have a worse problem than we do. In the failed on data request, seeing a failure on an attempt to close may trick us into thinking it was going down anyway.

Jefftree: Maybe I'm misunderstanding this but this ServeHTTP faces the apiserver so an EOF implies that the client (apiserver) chose to close the connection. This seems fairly common so we should get into this code path quite often. I guess if this code path was reached via failing to send data to the backend, an error logged here would be misleading...but that shouldn't happen too often right?

cheftako: Sounds right. However if the client closed the connection, then we can expect this send to fail. I'd rather not have alarming errors in the log for what is the normal case. Even in the case where there was an actual error, we are still the second failure and the first legitimate failure will have been logged as an error.

Jefftree: Discussed offline. V(2) might be more preferable but I agree we probably don't need this to be `ErrorS`.

rambohe-ch: @cheftako @Jefftree thank you for your kind comments, i will modify it to kog.V(2).InfoS

pkg/server/tunnel.go
rambohe-ch: @cheftako ok, add a anonymous deferred function to include the added code and the deferred conn close.

rambohe-ch:
@Jefftree How about approve to merge this PR?

pkg/server/tunnel.go
```diff
@@ -102,7 +102,21 @@ func (t *Tunnel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			Type: client.PacketType_CLOSE_REQ,
+			Payload: &client.Packet_CloseRequest{
+				CloseRequest: &client.CloseRequest{
+					ConnectID: connection.connectID,
+				},
+			},
+		}
+
+		if err = backend.Send(packet); err != nil {
+			klog.V(2).InfoS("failed to send close request packet", "host", r.Host, "agentID", connection.agentID, "connID", connection.connectID)
```
cheftako, line 116: Please move this to V(4). Failing to send a close request on tunnel shutting down isn't a surprise nor a significant state change. (Stopping the tunnel may be but the failure to send shutdown, less so)

cheftako: Talked to @Jefftree, and they convinced me we already downgraded the priority of this message once and to see how this priority plays out.

PR #163, opened 2020-12-08
cleanup all of conn contexts when agent client exits to prevent conn contexts leakage

rambohe-ch:
when connection between anp server and anp agent broken, the agent client will exits. and it need to clean up all of conn contexts to avoid connection between agent and remote leakage.

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -74,6 +74,16 @@ func (cm *connectionManager) Delete(connID int64) {
 	delete(cm.connections, connID)
 }
 
+func (cm *connectionManager) List() []*connContext {
+	cm.mu.RLock()
+	defer cm.mu.RUnlock()
+	connContexts := make([]*connContext, 0, len(cm.connections))
+	for _, connCtx := range cm.connections {
+		connContexts = append(connContexts, connCtx)
+	}
+	return connContexts
+}
+
 func newConnectionManager() *connectionManager {
 	return &connectionManager{
 		connections: make(map[int64]*connContext),
@@ -304,6 +314,14 @@ func (a *Client) initializeAuthContext(ctx context.Context) (context.Context, er
 // The requests include things like opening a connection to a server,
 // streaming data and close the connection.
 func (a *Client) Serve() {
+	defer func() {
+		// close all of conns with remote when Client exits
+		for _, connCtx := range a.connManager.List() {
+			connCtx.cleanup()
+		}
+		klog.V(2).InfoS("cleanup all of conn contexts when client exits", "agentID", a.agentID)
+	}()
+
 	klog.V(2).InfoS("Start serving", "serverID", a.serverID)
 	go a.probe()
 	for {


Jefftree:
Hello thank you for the PR! This looks like a pretty important fix to prevent leaking connections :)

Have you tested this already? I'm not sure if we can lock connManager and run the cleanup function since the cleanup removes the connection from the connManager and requires the lock (https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/b03bdaf5497312066131fedc6e8c270a4b19d735/pkg/agent/client.go#L333)

pkg/agent/client.go
```diff
@@ -248,6 +248,16 @@ func (a *AgentClient) initializeAuthContext(ctx context.Context) (context.Contex
 // The requests include things like opening a connection to a server,
 // streaming data and close the connection.
 func (a *AgentClient) Serve() {
+	defer func() {
+		// close all of conns with remote when AgentClient closed
+		a.connManager.mu.Lock()
+		klog.V(2).InfoS("cleanup all of conn contexts when agent client exits", "agentID", a.agentID)
+		for _, ctx := range a.connManager.connections {
```
cheftako, line 255: To avoid confusion can we change ctx to something else (maybe conn or connCtx)? We generally reserve ctx for variables of type context.Context which this isn't.

pkg/agent/client.go
```diff
@@ -248,6 +248,16 @@ func (a *AgentClient) initializeAuthContext(ctx context.Context) (context.Contex
...
+			ctx.cleanup()
+		}
+		a.connManager.mu.Unlock()
```
cheftako, line 258: The Unlock method call should be defer call right after the Lock call.

cheftako:
> I'm not sure if we can hold the lock to connManager and run the cleanup function since the cleanup removes the connection from the connManager and requires the lock

+1, Golang mutexes are not reentrant.

rambohe-ch:
> Hello thank you for the PR! This looks like a pretty important fix to prevent leaking connections :)
> 
> Have you tested this already? I'm not sure if we can hold the lock to connManager and run the cleanup function since the cleanup removes the connection from the connManager and requires the lock

@Jefftree thank you for your comments, this pr does raise the issue of interlocking.
and i think maybe we don't need lock the connManager in the defer routine when agent client exits, because new connection from server can not added into the agent client.

pkg/agent/client.go
rambohe-ch: yeah,  changed ctx to connCtx

pkg/agent/client.go
rambohe-ch: according to @Jefftree comments, removed the connManager lock here

rambohe-ch:
> > Hello thank you for the PR! This looks like a pretty important fix to prevent leaking connections :)
> > Have you tested this already? I'm not sure if we can hold the lock to connManager and run the cleanup function since the cleanup removes the connection from the connManager and requires the lock
> 
> @Jefftree thank you for your comments, this pr does raise the issue of interlocking.
> and i think maybe we don't need lock the connManager in the defer routine when agent client exits, because new connection from server can not added into the agent client.

 if not to lock connManager, a data race problem raised(https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_apiserver-network-proxy/163/pull-apiserver-network-proxy-test/1339416242276012032). so it need to add a List() func for getting all of connContexts, and then cleanup connContext in a loop.

rambohe-ch:
@Jefftree Based on review comments,  I had refactored the code for clean context when client exits, Would you take a review, and merge the pr if no other problems? Thank you very much.

Jefftree:
@rambohe-ch Thanks for addressing all the comments!

/lgtm
/assign @cheftako

pkg/agent/client.go
```diff
@@ -73,6 +73,16 @@ func (cm *connectionManager) Delete(connID int64) {
 	delete(cm.connections, connID)
 }
 
+func (cm *connectionManager) List() []*connContext {
+	cm.mu.Lock()
```
cheftako, line 77: Should be safe for this to be RLock()

pkg/agent/client.go
```diff
@@ -73,6 +73,16 @@ func (cm *connectionManager) Delete(connID int64) {
...
+	defer cm.mu.Unlock()
+	connContexts := make([]*connContext, 0, len(cm.connections))
+	for _, connCtx := range cm.connections {
```
cheftako, line 80: Might be better to return pairs of ConnectionID and ConnContexts.

pkg/agent/client.go
```diff
@@ -248,6 +258,14 @@ func (a *AgentClient) initializeAuthContext(ctx context.Context) (context.Contex
 // The requests include things like opening a connection to a server,
 // streaming data and close the connection.
 func (a *AgentClient) Serve() {
+	defer func() {
+		// close all of conns with remote when AgentClient closed
+		for _, connCtx := range a.connManager.List() {
+			connCtx.cleanup()
```
cheftako, line 264: Can we delete the connCtx from the connManager. Otherwise we end up with a broken connection dangling in the connection list. While we are on the way out, it still seems prone to race conditions.

Jefftree: The cleanup function automatically deletes the connection from the list. (https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/ff2a951482c1cdc615e00107fac9a42bf279f042/pkg/agent/client.go#L341)

pkg/agent/client.go
cheftako: Given https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/163#discussion_r573280986 please ignore this comment.

pkg/agent/client.go
cheftako: Good catch. I withdraw the need for the delete. I think the RLock still applies though.

pkg/agent/client.go
rambohe-ch: ok

rambohe-ch:
@cheftako I had revised client.go according to your review comments.  Would you take a review once more.

rambohe-ch:
@Jefftree I think this PR is ready for merging. would you make a merging for it?

PR #164, opened 2020-12-08
fix request hangs up when agent failed to return PacketType_DIAL_RSP

rambohe-ch:
- background:
at present if agent does not return response for some reason, the request will be hang up in tunnel serve. so we need to add timeout process for pending client connection. 

- solution:
 when anp agent doesn't return response after 15seconds, tunnel server will take timeout process for proxy client connection to release client connection.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -41,6 +41,10 @@ import (
 
 type key int
 
+var (
+	ConnectionSetupTimeout = 15 * time.Second
+)
+
 type ProxyClientConnection struct {
 	Mode      string
 	Grpc      client.ProxyService_ProxyServer
@@ -80,9 +84,33 @@ func (c *ProxyClientConnection) send(pkt *client.Packet) error {
 }
 
 func NewPendingDialManager() *PendingDialManager {
-	return &PendingDialManager{
+	pm := &PendingDialManager{
 		pendingDial: make(map[int64]*ProxyClientConnection),
 	}
+
+	// spawn a goroutine to verify connection waiting is timeout or not.
+	go func() {
+		ticker := time.NewTicker(3 * time.Second)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-ticker.C:
+				pm.mu.Lock()
+				for random, proxyClientConn := range pm.pendingDial {
+					// client connection pending over 15seconds
+					if time.Since(proxyClientConn.start) > ConnectionSetupTimeout {
+						delete(pm.pendingDial, random)
+						close(proxyClientConn.connected)
+						klog.V(2).InfoS("connection setup timeout", "random", random)
+					}
+				}
+				pm.mu.Unlock()
+			}
+		}
+	}()
+
+	return pm
 }
 
 type PendingDialManager struct {
--- a/pkg/server/tunnel.go
+++ b/pkg/server/tunnel.go
@@ -103,6 +103,11 @@ func (t *Tunnel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	case <-connection.connected: // Waiting for response before we begin full communication.
 	}
 
+	if connection.connectID == 0 {
+		klog.Errorf("wait for dial response timeout with random %d", random)
+		return
+	}
+
 	defer func() {
 		packet := &client.Packet{
 			Type: client.PacketType_CLOSE_REQ,


Jefftree:
Thank you for the PR! AFAIK this is handled by the GRPC client with a dial timeout (https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/konnectivity-client/pkg/client/client.go#L200), so this problem only exists for http connect. It might be a good idea to first add a client side timeout there (https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/egressselector/egress_selector.go#L173)

/ok-to-test

rambohe-ch:
> Thank you for the PR! AFAIK this is handled by the GRPC client with a dial timeout (https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/konnectivity-client/pkg/client/client.go#L200), so this problem only exists for http connect. It might be a good idea to first add a client side timeout there (https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/egressselector/egress_selector.go#L173)

@Jefftree Thank you for your detailed response.
the timeout in the pr is added for a PacketType_DIAL_REQ request(https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/server/tunnel.go#L61) in tunnel serve, not for dial connection between anp server and anp client. 
so add in client side((https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/egressselector/egress_selector.go#L173)) can not solve tunnel hang up(https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/server/tunnel.go#L102)

pkg/server/tunnel.go
```diff
@@ -71,7 +71,7 @@ func (t *Tunnel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	klog.V(4).InfoS("Set pending", "random", random, "value", w)
 	backend, err := t.Server.BackendManager.Backend(context.TODO())
 	if err != nil {
-		http.Error(w, fmt.Sprintf("currently no tunnels available: %v", err), http.StatusInternalServerError)
+		klog.ErrorS(err, "currently no tunnels available")
```
Jefftree, line 74: Why the change here? I think it still makes sense to send the error back to the client if we have no tunnels.

Jefftree:
Ahh got it, I originally thought this PR was addressing timing out the client but it looks like even timing out the client doesn't fix the problem of the server hanging.

I'm not a huge fan of locking and iterating through all the PendingDials every 3 seconds, but don't really have alternative suggestions. The http-connect side looks great, but we'll probably need to send an additional DIAL_RSP to the frontend to indicate that the dial timed out in GRPC mode.

Maybe @cheftako can provide some more insight.

pkg/server/tunnel.go
cheftako: I believe this is switching to the new structured logging system (https://godoc.org/github.com/kubernetes/klog#ErrorS) and will automatically attach the error in a standardized manner.

pkg/server/server.go
```diff
@@ -73,9 +77,33 @@ func (c *ProxyClientConnection) send(pkt *client.Packet) error {
...
+
+	// spawn a goroutine to verify connection waiting is timeout or not.
+	go func() {
+		ticker := time.NewTicker(3 * time.Second)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-ticker.C:
+				pm.mu.Lock()
```
cheftako, line 92: Please break this block into a separate block so we can defer the unlock.

pkg/server/server.go
```diff
@@ -73,9 +77,33 @@ func (c *ProxyClientConnection) send(pkt *client.Packet) error {
...
 
 func NewPendingDialManager() *PendingDialManager {
-	return &PendingDialManager{
+	pm := &PendingDialManager{
 		pendingDial: make(map[int64]*ProxyClientConnection),
 	}
+
+	// spawn a goroutine to verify connection waiting is timeout or not.
+	go func() {
+		ticker := time.NewTicker(3 * time.Second)
```
cheftako, line 86: I would guess that most of the time connections resolve quickly. The would suggest that if we walk the pendingDials (we are anyway) we could determine minimum interval before the next connection request would expire and it would be close to the ConnectionSetupTimeout. That would suggest we would be better waiting the max of 3  seconds or ConnectionSetupTimeout. This reduces how often we have to grab the lock and walk the list.

pkg/server/server.go
```diff
@@ -73,9 +77,33 @@ func (c *ProxyClientConnection) send(pkt *client.Packet) error {
...
+		defer ticker.Stop()
```
cheftako, line 87: We should probably log V(2) something when this goroutine exits as it suggests that we will no longer be doing clean up.

pkg/server/server.go
```diff
@@ -73,9 +77,33 @@ func (c *ProxyClientConnection) send(pkt *client.Packet) error {
...
+		for {
+			select {
+			case <-ticker.C:
+				pm.mu.Lock()
+				for random, proxyClientConn := range pm.pendingDial {
+					// client connection pending over 15seconds
+					if time.Since(proxyClientConn.start) > ConnectionSetupTimeout {
+						delete(pm.pendingDial, random)
+						close(proxyClientConn.connected)
+						klog.V(2).InfoS("connection setup timeout", "random", random)
```
cheftako, line 98: metrics.Metrics.ObserveDialLatency(time.Since(pm.pendingDial.start)) is especially important in this use case. Would also be good if we had a DialError metric.

pkg/server/server.go
```diff
@@ -37,6 +37,10 @@ import (
 	"sigs.k8s.io/apiserver-network-proxy/proto/header"
 )
 
+var (
+	ConnectionSetupTimeout = 15 * time.Second
```
cheftako, line 41: This seems a fine default. It should be configurable though.

pkg/server/tunnel.go
Jefftree: Right, this makes sense if we we were changing from another form of klog. But going from sending an error on the tunnel back to the client to only logging the error locally seems like a big jump?

cheftako: Doh!
Your completely correct! We probably want both but we definitely should not remove the http.Error.

charleszheng44: Can we keep both of them? The `klog` can be helpful when debugging on the server-side.

pkg/server/server.go
Jefftree: Food for thought: I'm wondering if it'll make sense to add another data structure that could sort the PendingDials by their start time. Then we'd only need to iterate through the oldest n entries every time.

pkg/server/tunnel.go
cheftako: Yes

pkg/server/server.go
cheftako: Depends on the length of the list. If we find it gets long then yes it would make sense. My guess (and its only a guess) is that usually requests spend very little time in the pendingDials list. If that's true then the length should be short and it may be mostly wasted effort. If it turns out to be false, then I think it would be a good change.

pkg/server/tunnel.go
rambohe-ch: > Why the change here? I think it still makes sense to send the error back to the client if we have no tunnels.

ResponseWriter is already hijacked, and i think the object w can not be used as an error responder. so modified to klog.Errors() and log the error info.

PR #168, opened 2021-02-02
Fix a memory leak that can happen when the initial dial fails.

jkh52:


konnectivity-client/pkg/client/client.go
```diff
@@ -115,6 +115,12 @@ func (t *grpcTunnel) serve(c clientConn) {
 					connid: resp.ConnectID,
 				}
 			}
+
+			if resp.Error != "" {
+				// On dial error, avoid leaking serve goroutine.
```
caesarxuchao, line 120: Can we log the error?

caesarxuchao (commented):
LGTM. Just a nit.

konnectivity-client/pkg/client/client.go
cheftako: Not against this. Worth noting that I believe the resp.Error will be passed back to Dial() through the channel and then returned up to the caller of Dial. I'm assuming that caller will log it with more context.

caesarxuchao: Good point. I missed line 114. Then the log is less important.

konnectivity-client/pkg/client/client.go
```diff
@@ -115,6 +115,12 @@ func (t *grpcTunnel) serve(c clientConn) {
 					connid: resp.ConnectID,
 				}
 			}
+
+			if resp.Error != "" {
```
liggitt, line 119: is it possible to add a test for this scenario that would have caught the leak?

Keramblock:
Just for those guys, that experiencing same problem, as I with GKE 1.19.8 in regular channel and "dialing backend: dial timeout" error in logs. And cannot find anything related to it in google. That PR fixed it, you just need to update k8s master to 1.19.9. 

P.S. Thx for the Gitlab support for helping me find the root cause of the problem!

PR #170, opened 2021-02-04
introduce lint target

Tamas-Biro1:
- this PR introduces lint Makefile target
- contains some minor fixes and changes for lint, gosec and gofmt

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -45,7 +45,10 @@ func main() {
 	flags.AddFlagSet(o.Flags())
 	local := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	klog.InitFlags(local)
-	local.Set("v", "4")
+	err := local.Set("v", "4")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error setting klog flags: %v", err)
+	}
 	local.VisitAll(func(fl *flag.Flag) {
 		fl.Name = util.Normalize(fl.Name)
 		flags.AddGoFlag(fl)
--- a/cmd/client/main.go
+++ b/cmd/client/main.go
@@ -47,7 +47,10 @@ func main() {
 	flags.AddFlagSet(o.Flags())
 	local := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	klog.InitFlags(local)
-	local.Set("v", "4")
+	err := local.Set("v", "4")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error setting klog flags: %v", err)
+	}
 	local.VisitAll(func(fl *flag.Flag) {
 		fl.Name = util.Normalize(fl.Name)
 		flags.AddGoFlag(fl)
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -28,6 +28,7 @@ import (
 	"net/http/pprof"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"sync"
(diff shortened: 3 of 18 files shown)


Tamas-Biro1:
I signed it.

cmd/agent/main.go
```diff
@@ -45,7 +45,10 @@ func main() {
 	flags.AddFlagSet(o.Flags())
 	local := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	klog.InitFlags(local)
-	local.Set("v", "4")
+	err := local.Set("v", "4")
+	if err != nil {
+		klog.Errorf("error: %v\n", err)
```
cheftako, line 50: There is a certain optimism inherent in attempting to use klog when you just received an error when configuring klog. Not saying I have a better idea but I wonder if it might be worth also sending the error message to stderr.

cmd/client/main.go
```diff
@@ -47,7 +47,10 @@ func main() {
 	flags.AddFlagSet(o.Flags())
 	local := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	klog.InitFlags(local)
-	local.Set("v", "4")
+	err := local.Set("v", "4")
+	if err != nil {
+		klog.Errorf("error: %v\n", err)
```
cheftako, line 52: Ditto

cmd/server/main.go
```diff
@@ -61,7 +62,10 @@ func main() {
 	flags.AddFlagSet(o.Flags())
 	local := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	klog.InitFlags(local)
-	local.Set("v", "4")
+	err := local.Set("v", "4")
+	if err != nil {
+		klog.Errorf("error: %v\n", err)
```
cheftako, line 67: Ditto

cmd/agent/main.go
cheftako: Also please use structured logging in you PR. 
Eg. 
klog.ErrorS(err, "Failed to set log level")

cmd/server/main.go
```diff
@@ -482,14 +486,18 @@ func (p *Proxy) runUDSMasterServer(ctx context.Context, o *ProxyRunOptions, s *s
...
+		}
 		go func() {
 			udsListener, err := getUDSListener(ctx, o.udsName)
 			if err != nil {
 				klog.ErrorS(err, "failed to get uds listener")
 			}
 			defer func() {
-				udsListener.Close()
+				err := udsListener.Close()
+				klog.Errorf("error: %v\n", err)
```
cheftako, line 500: ErrorS(err, "Failed to close UDS listener")

pkg/agent/client.go
```diff
@@ -80,15 +80,16 @@ func newConnectionManager() *connectionManager {
...
 // choosing agents
-type AgentIdentifiers struct {
+type Identifiers struct {
 	IPv4 []string
 	IPv6 []string
 	Host []string
 	CIDR []string
 }
 
+// IdentifierType ...
```
cheftako, line 92: Can we use a TODO instead of ... if you want someone to follow up adding verbiage?

cmd/agent/main.go
Tamas-Biro1: thanks, let's try fmt.Fprint in those cases when klog initialization happens

pkg/agent/client.go
Tamas-Biro1: finally I removed these comments that I added, because this PR is not intended to solve this. Maybe someone (or I) might create another PR for only fixing these verbiage

cmd/test-server/main.go
```diff
@@ -204,6 +207,11 @@ func (p *TestServer) runMasterServer(ctx context.Context, o *TestServerRunOption
 		klog.Warningf("HTTP test server stopped listening\n")
 	}()
 
-	stop := func() { server.Shutdown(ctx) }
+	stop := func() {
+		err := server.Shutdown(ctx)
+		if err != nil {
+			klog.Errorf("error: %v\n", err)
```
cheftako, line 213: Missed one.

Tamas-Biro1: sry, missed that one. fixed

cheftako:
Can we look into added an automated pre-submit lint check once this merges?

Tamas-Biro1:
> Can we look into added an automated pre-submit lint check once this merges?

Sure, good idea. I will do it soon.

PR #172, opened 2021-02-23
Fix bug around low steps for exponential backoff

Jefftree:
With Steps=3, the backoff interval will change from 1s to 1.5s to 2.25s and cap at that number. This PR allows the backoff interval to reach the Cap of 60s.

/assign @cheftako 
/assign @caesarxuchao 
/cc @jkh52

### DIFF
--- a/pkg/agent/clientset.go
+++ b/pkg/agent/clientset.go
@@ -18,6 +18,7 @@ package agent
 
 import (
 	"fmt"
+	"math"
 	"sync"
 	"time"
 
@@ -43,6 +44,9 @@ type ClientSet struct {
 	// proxy server.
 	probeInterval time.Duration // The interval by which the agent
 	// periodically checks if its connections to the proxy server is ready.
+	syncIntervalCap time.Duration // The maximum interval
+	// for the syncInterval to back off to when unable to connect to the proxy server
+
 	dialOptions []grpc.DialOption
 	// file path contains service account token
 	serviceAccountTokenPath string
@@ -114,6 +118,7 @@ type ClientSetConfig struct {
 	AgentIdentifiers        string
 	SyncInterval            time.Duration
 	ProbeInterval           time.Duration
+	SyncIntervalCap         time.Duration
 	DialOptions             []grpc.DialOption
 	ServiceAccountTokenPath string
 }
@@ -126,6 +131,7 @@ func (cc *ClientSetConfig) NewAgentClientSet(stopCh <-chan struct{}) *ClientSet 
 		address:                 cc.Address,
 		syncInterval:            cc.SyncInterval,
 		probeInterval:           cc.ProbeInterval,
+		syncIntervalCap:         cc.SyncIntervalCap,
 		dialOptions:             cc.DialOptions,
 		serviceAccountTokenPath: cc.ServiceAccountTokenPath,
 		stopCh:                  stopCh,
@@ -138,11 +144,11 @@ func (cs *ClientSet) newAgentClient() (*Client, int, error) {
 
 func (cs *ClientSet) resetBackoff() *wait.Backoff {
 	return &wait.Backoff{
-		Steps:    3,
+		Steps:    math.MaxInt32,
 		Jitter:   0.1,
 		Factor:   1.5,
 		Duration: cs.syncInterval,
-		Cap:      60 * time.Second,
+		Cap:      cs.syncIntervalCap,
 	}
 }
 
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -78,6 +78,7 @@ type GrpcProxyAgentOptions struct {
 	agentIdentifiers string
 	syncInterval     time.Duration
 	probeInterval    time.Duration
+	syncIntervalCap  time.Duration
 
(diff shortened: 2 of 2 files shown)


cheftako:
So I agree with the analysis of what the code is doing. I even agree that the actual cap vs the duration limit is confusing. However do we really want the backoff interval to scale all the way to 60 seconds? This is determining how long it can take to reestablish tunnels after whatever breakage has been fixed. Given our retry on serverID/connection collision, it might even just be a fraction of the time. The cost of making an attempt seems small enough that it's not clear that retrying every few seconds is a problem. I could even see stretching that to 10.

Jefftree:
I'm fine with changing the limit to 10s.

The cost of making an attempt is small on the agent side, but on the server side these attempts add up. In an HA environment, when a control plane instance is down all the agents will try to reconnect immediately. It's possible that the other control plane instances could be overwhelmed at that time. Our code has a drawback where a connection must be authenticated before we can detect it's a duplicate and terminate it. Every connection attempt is an extra request to the apiserver to validate the agent token so the retries indirectly contribute to apiserver's QPS. Given a cluster with ~1k nodes, an additional 1k QPS is probably not negligible.

jkh52:
All of this hardcoded tuning is relative to --sync-interval. But what if --sync-interval is 20s? Then this cap of 10s does not make sense.

> Our code has a drawback where a connection must be authenticated before we can detect it's a duplicate and terminate it. Every connection attempt is an extra request to the apiserver to validate the agent token so the retries indirectly contribute to apiserver's QPS. For example, adding an extra 1k QPS to apiserver is probably not negligible in a 1k node cluster.

Yes - as an operator I want to reason about the agent grpc connect rate for degraded HA cluster. This happens during a rolling restart, or if one control plane instance is simply down (we must not allow the agent to take down the others.)

Nit: "1k agent cluster". Various examples (https://kubernetes.io/docs/tasks/extend-kubernetes/setup-konnectivity/) use DaemonSet but that doesn't seem ideal for large clusters. Does this project document a scalability guideline around max agent count?

cheftako:
> All of this hardcoded tuning is relative to --sync-interval. But what if --sync-interval is 20s? Then this cap of 10s does not make sense.
> 
> > Our code has a drawback where a connection must be authenticated before we can detect it's a duplicate and terminate it. Every connection attempt is an extra request to the apiserver to validate the agent token so the retries indirectly contribute to apiserver's QPS. For example, adding an extra 1k QPS to apiserver is probably not negligible in a 1k node cluster.
> 
> Yes - as an operator I want to reason about the agent grpc connect rate for degraded HA cluster. This happens during a rolling restart, or if one control plane instance is simply down (we must not allow the agent to take down the others.)
> 
> Nit: "1k agent cluster". Various examples (https://kubernetes.io/docs/tasks/extend-kubernetes/setup-konnectivity/) use DaemonSet but that doesn't seem ideal for large clusters. Does this project document a scalability guideline around max agent count?

We have certainly discussed in various of the related designs the idea that we might not need/want DaemonSet on a large cluster but we have no concrete evidence that this is true to date. I'd wait for the performance tests results before we panic on this.

Lets talk rolling restart on a 15K Node HA cluster. We expect ~1/3 of the agents to be sending to each Konnectivity Server at any time. So we get about 5K nodes sending connection requests to a server. So we default this to starting interval of ~1 second. So even if the agents all started at the same second (they won't) you would expect 5K QPS connection requests toward the Konnectivity Server. As Backoff kicks in, with a 10 second cap this would drop to about 500 QPS at the worst. I believe it would handle the 5K QPS and that we will never actually see it that high. 

Also we are talking here with the cap not the initial period. The initial period is what should worry us with max volume.

cheftako:
Not approving (yet) to leave time for more inspection.

jkh52:
The auth perf concern (which we really have seen in some clusters) can be a separate issue, and this can be about reconciling the tuning parameters. Maybe making the retry cap a command line flag is a solution, like --sync-interval-max-backoff or similar.

Jefftree:
@cheftako @jkh52 Added a flag sync-interval-cap for the cap of the backoff. Does it make sense now to trigger the backoff on requests to a duplicate control plane instance or should that be a separate issue? (https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/agent/clientset.go#L188)

pkg/agent/clientset.go
```diff
@@ -43,6 +44,9 @@ type ClientSet struct {
 	// proxy server.
 	probeInterval time.Duration // The interval by which the agent
 	// periodically checks if its connections to the proxy server is ready.
+	syncIntervalCap time.Duration // The maximum interval
```
cheftako, line 47: Would be nice to add validation for probeInterval and syncIntervalCap. (Eg probeInterval < syncIntervalCap)

caesarxuchao: +1

caesarxuchao:
LGTM, too.

pkg/agent/clientset.go
Jefftree: Fixed in `cmd/agent/main.go`

Issue #171, opened 2021-02-23
Implement support for node to master network traffic (KEP-2025)


timoreimann:
This meta issue is to track all the work items necessary to implement [KEP-2025](https://github.com/kubernetes/enhancements/tree/master/keps/sig-cloud-provider/2025-extend-konnectivity-for-both-directions) ([permalink](https://github.com/kubernetes/enhancements/tree/846705f347e51537fbef5835b321f5933d088b5c/keps/sig-cloud-provider/2025-extend-konnectivity-for-both-directions)). Individual tasks can be broken out into more specific issues as needed.

At a high level, the following tasks must be completed (see the KEP for details):

- [ ] extension of the Konnectivity server to support the new flags (CLI options and feature gate) and implement the necessary protocol extensions.
- [ ] extension of the Konnectivity agent to support the new flags (CLI options and feature gate) and implement the necessary protocol extensions.
- [ ] extension of the agent to manage the dummy interface. (The NodeLocal DNSCache which does something similar could serve as inspiration here.)
- [ ] proper test coverage (to the extent feasible by the current testing framework; in particular, achieving full e2e test coverage could still be difficult at this point and may be postponed to a later point in time).
- [ ] documentation updates (perhaps also a recording).

Refs kubernetes/enhancements#2025
Refs #36

/cc @cheftako @irozzo-1A @youssefazrak

irozzo-1A:
thx @timoreimann ;-)

timoreimann:
It's that time of the quarter again

/reopen

PR #173, opened 2021-02-25
Add feature gate to enable master to cluster traffic

irozzo-1A:
This PR introduces the feature gate to enable the master to cluster traffic.

Despite this does not represent a very significant increase in the implementation of the functionality I created a dedicated PR as this will be used both on the agent and the server side.

Ref: #171


pkg/features/features.go
```diff
@@ -0,0 +1,42 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
cheftako, line 2: 2021

cheftako:
cc @deads2k 
Reminded of our conversation https://github.com/kubernetes/kubernetes/pull/92201#discussion_r452370878. I think even here there is something to be said for this feature being Kubernetes version bound rather than ANP version bound. It would suggest we might want this feature in k8s.io/api.

deads2k:
> cc @deads2k
> Reminded of our conversation [kubernetes/kubernetes#92201 (comment)](https://github.com/kubernetes/kubernetes/pull/92201#discussion_r452370878). I think even here there is something to be said for this feature being Kubernetes version bound rather than ANP version bound. It would suggest we might want this feature in k8s.io/api.

I think having featuregate names in k8s.io/API is a good idea, but I think I would draw a boundary for inclusion there at "used by a binary in k/k".  I think that having distinct featuregates for different binaries is pretty reasonable.

cheftako:
cc @andrewsykim

PR #176, opened 2021-03-01
Extend agent to proxy KAS traffic generated on node network

irozzo-1A:
This PR implements #177.

### DIFF
--- a/cmd/agent/app/options/options.go
+++ b/cmd/agent/app/options/options.go
@@ -82,6 +82,10 @@ func (pm *portMapping) Type() string {
 	return "portMapping"
 }
 
+func (pm *portMapping) IsEmpty() bool {
+	return *pm == portMapping{}
+}
+
 func (o *GrpcProxyAgentOptions) ClientSetConfig(dialOptions ...grpc.DialOption) *agent.ClientSetConfig {
 	return &agent.ClientSetConfig{
 		Address:                 fmt.Sprintf("%s:%d", o.ProxyServerHost, o.ProxyServerPort),
@@ -201,6 +205,11 @@ func (o *GrpcProxyAgentOptions) Validate() error {
 	if err := validateHostnameOrIP(o.BindAddress); err != nil {
 		return fmt.Errorf("agent bind address is invalid: %v", err)
 	}
+	// if we have the zero value the feature is disabled no need to do further
+	// checks
+	if o.ApiServerMapping.IsEmpty() {
+		return nil
+	}
 	if err := validateHostnameOrIP(o.ApiServerMapping.RemoteHost); err != nil {
 		return fmt.Errorf("apiserver address is invalid: %v", err)
 	}
@@ -271,8 +280,8 @@ func NewGrpcProxyAgentOptions() *GrpcProxyAgentOptions {
 		ServiceAccountTokenPath:   "",
 		WarnOnChannelLimit:        false,
 		SyncForever:               false,
-		ApiServerMapping:          portMapping{LocalPort: 6443, RemoteHost: "localhost", RemotePort: 6443},
 		BindAddress:               "127.0.0.1",
+		ApiServerMapping:          portMapping{},
 	}
 	return &o
 }
--- a/cmd/agent/app/options/options.go
+++ b/cmd/agent/app/options/options.go
@@ -2,16 +2,20 @@ package options
 
 import (
 	"fmt"
+	"net"
 	"net/url"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/google/uuid"
 	"github.com/spf13/pflag"
 	"google.golang.org/grpc"
+	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/klog/v2"
 
 	"sigs.k8s.io/apiserver-network-proxy/pkg/agent"
+	"sigs.k8s.io/apiserver-network-proxy/pkg/features"
 	"sigs.k8s.io/apiserver-network-proxy/pkg/util"
 )
 
@@ -55,7 +59,27 @@ type GrpcProxyAgentOptions struct {
 	// The check is an "unlocked" read but is still use at your own peril.
 	WarnOnChannelLimit bool
 
-	SyncForever bool
+	SyncForever      bool
+	BindAddress      string
+	ApiServerMapping portMapping
+}
+
+var _ pflag.Value = &portMapping{}
+
+// port mapping represents the mapping between a local port and a remote
+// destination.
+type portMapping agent.PortMapping
+
+func (pm *portMapping) String() string {
+	return (*agent.PortMapping)(pm).String()
+}
+
+func (pm *portMapping) Set(s string) error {
+	return (*agent.PortMapping)(pm).Parse(s)
+}
+
+func (pm *portMapping) Type() string {
+	return "portMapping"
 }
 
 func (o *GrpcProxyAgentOptions) ClientSetConfig(dialOptions ...grpc.DialOption) *agent.ClientSetConfig {
@@ -95,6 +119,10 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.AgentIdentifiers, "agent-identifiers", o.AgentIdentifiers, "Identifiers of the agent that will be used by the server when choosing agent. N.B. the list of identifiers must be in URL encoded format. e.g.,host=localhost&host=node1.mydomain.com&cidr=127.0.0.1/16&ipv4=1.2.3.4&ipv4=5.6.7.8&ipv6=:::::&default-route=true")
 	flags.BoolVar(&o.WarnOnChannelLimit, "warn-on-channel-limit", o.WarnOnChannelLimit, "Turns on a warning if the system is going to push to a full channel. The check involves an unsafe read.")
 	flags.BoolVar(&o.SyncForever, "sync-forever", o.SyncForever, "If true, the agent continues syncing, in order to support server count changes.")
+	flags.Var(&o.ApiServerMapping, "apiserver-port-mapping", "Mapping between a local port and the host:port used to reach the Kubernetes API Server")
+	flags.StringVar(&o.BindAddress, "bind-address", o.BindAddress, "Address used to listen for traffic generated on cluster network")
+	// add feature gates flag
+	features.DefaultMutableFeatureGate.AddFlag(flags)
 	return flags
 }
 
@@ -119,6 +147,10 @@ func (o *GrpcProxyAgentOptions) Print() {
 	klog.V(1).Infof("AgentIdentifiers set to %s.\n", util.PrettyPrintURL(o.AgentIdentifiers))
 	klog.V(1).Infof("WarnOnChannelLimit set to %t.\n", o.WarnOnChannelLimit)
 	klog.V(1).Infof("SyncForever set to %v.\n", o.SyncForever)
+	if features.DefaultMutableFeatureGate.Enabled(features.NodeToMasterTraffic) {
+		klog.V(1).Infof("AgentBindAddress set to %s.\n", o.BindAddress)
+		klog.V(1).Infof("Apiserver port mapping set to %s.\n", o.ApiServerMapping.String())
+	}
 }
 
 func (o *GrpcProxyAgentOptions) Validate() error {
@@ -166,6 +198,24 @@ func (o *GrpcProxyAgentOptions) Validate() error {
 	if err := validateAgentIdentifiers(o.AgentIdentifiers); err != nil {
 		return fmt.Errorf("agent address is invalid: %v", err)
 	}
+	if err := validateHostnameOrIP(o.BindAddress); err != nil {
+		return fmt.Errorf("agent bind address is invalid: %v", err)
+	}
+	if err := validateHostnameOrIP(o.ApiServerMapping.RemoteHost); err != nil {
+		return fmt.Errorf("apiserver address is invalid: %v", err)
+	}
+	if o.ApiServerMapping.LocalPort > 49151 {
+		return fmt.Errorf("please do not try to use ephemeral port %d for the apiserver local port", o.ApiServerMapping.LocalPort)
+	}
+	if o.ApiServerMapping.LocalPort < 1024 {
+		return fmt.Errorf("please do not try to use reserved port %d for the apiserver local port", o.ApiServerMapping.LocalPort)
+	}
+	if o.ApiServerMapping.RemotePort > 49151 {
+		return fmt.Errorf("please do not try to use ephemeral port %d for the apiserver remote port", o.ApiServerMapping.LocalPort)
+	}
+	if o.ApiServerMapping.RemotePort < 1 {
+		return fmt.Errorf("invalid port %d for the apiserver remote port", o.ApiServerMapping.RemotePort)
+	}
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


huxiaoliang:
this is a great feature that enable bi-directional communication between cloud and edge

cmd/agent/app/options/options.go
```diff
@@ -182,6 +245,8 @@ func NewGrpcProxyAgentOptions() *GrpcProxyAgentOptions {
 		ProbeInterval:             1 * time.Second,
 		SyncIntervalCap:           10 * time.Second,
 		ServiceAccountTokenPath:   "",
+		ApiServerMapping:          portMapping{LocalPort: 6443, RemoteHost: "localhost", RemotePort: 6443},
```
cheftako, line 248: I would prefer no default values so this feature is off unless both the feature gate is on and these flags are set. Also 6443 seems an odd remote port for the KAS.

cmd/agent/app/server.go
```diff
@@ -39,11 +42,19 @@ func (a *Agent) run(o *options.GrpcProxyAgentOptions) error {
 		return fmt.Errorf("failed to validate agent options with %v", err)
 	}
 
-	stopCh := make(chan struct{})
-	if err := a.runProxyConnection(o, stopCh); err != nil {
+	ctx := context.Background()
```
cheftako, line 45: Might be good to set the up as 

```
ctx, cancel := context.WithCancel(context.Background())
defer cancel()
```

cmd/agent/app/server.go
```diff
@@ -39,11 +42,19 @@ func (a *Agent) run(o *options.GrpcProxyAgentOptions) error {
...
+	var err error
+	var cs *agent.ClientSet
+	if cs, err = a.runProxyConnection(ctx, o); err != nil {
```
cheftako, line 48: Wondering if it would make sense to move this into app ala our server code.

cheftako: Rather than proxy connection, cluster egress might make more sense?

cmd/agent/app/server.go
```diff
@@ -39,11 +42,19 @@ func (a *Agent) run(o *options.GrpcProxyAgentOptions) error {
...
 		return fmt.Errorf("failed to run proxy connection with %v", err)
 	}
 
+	if features.DefaultMutableFeatureGate.Enabled(features.NodeToMasterTraffic) {
+		if err := a.runControlPlaneProxy(ctx, o, cs); err != nil {
```
cheftako, line 53: Might be good to only run the proxy if bind address and mapping are set.

cheftako: Rather than proxy connection, cluster egress might make more sense (as a name)?

cheftako:
Not seeing any server code changes or grpc proto changes. Assuming this is not the complete work?

cmd/agent/app/options/options.go
irozzo-1A: @cheftako I did it this way to match the default secure port of KAS that is `6443`, note that the two have to match for this feature to work (see the [KEP](https://github.com/kubernetes/enhancements/tree/846705f347e51537fbef5835b321f5933d088b5c/keps/sig-cloud-provider/2025-extend-konnectivity-for-both-directions#handling-the-traffic-from-the-pods-to-the-agent)). I thought that the feature gate was enough, but if you prefer this to be specified to enable the feature I can do it.

cmd/agent/app/server.go
irozzo-1A: > Might be good to only run the proxy if bind address and mapping are set.

I'm ok with that if we remove the defaults.

irozzo-1A:
> Not seeing any server code changes or grpc proto changes. Assuming this is not the complete work?

@cheftako Indeed this is not the complete work, this implements only the agent changes only. The server part is supposed to be implemented with a dedicated PR (see the [aggregator PR](https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/171) and [this](https://kubernetes.slack.com/archives/CPS4Z8G8N/p1626959711035400) thread on slack for more context).

I tried to implement this without the need for any grpc proto change, in order to distinguish `cluster -> master` connections from `master -> cluster` connections I adopted the convention of using even `ConnectID` for the first case and even for the latter (see https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/176/files#diff-c2eff45b2c7db0ae864aecb2e8b733dfbdfe7070ffe6f7feb1204e086c403a0aR523-R525). The server code where the `ConnectID` is generated will also need to be adapted accordingly.
If you see some problems with this approach, I did not so far but I could have missed something, the alternative would be to add something to identify the initiator (i.e. agent or server) besides the `ConnectID` in the proto (it should be optional for backward compatibility)

pkg/agent/client.go
```diff
@@ -64,42 +66,84 @@ func (c *connContext) send(msg []byte) {
 }
 
 type connectionManager struct {
-	mu          sync.RWMutex
+	connLock    sync.RWMutex
 	connections map[int64]*connContext
+
+	// pending dial
+	pendingDial     map[int64]*pendingDialContext
+	pendingDialLock sync.RWMutex
```
PratikDeoghare, line 74:
```suggestion
	pendingDialLock sync.RWMutex
	pendingDial     map[int64]*pendingDialContext
```

nit: move mutex above what it protects

cmd/agent/app/options/options.go
irozzo-1A: done

cmd/agent/app/server.go
irozzo-1A: done

SataQiu:
Great feature!

/ok-to-test

soider:
Hello! what is the current state of this PR?
@jnummelin and me at k0sproject.io used this PR as a branch for our feature branch and would be nice to sync up the changes we introduced to enable handling worker to kas traffic.

irozzo-1A:
Hi @soider, @PratikDeoghare prepared the server-side (PR is coming soon) and I'm preparing a manual e2e test to verify that everything works as expected. I should have something by the end of the week. After that, I will kindly ask @cheftako to do another pass :-)

soider:
@irozzo-1A @PratikDeoghare have you had it already implemented? we have some changes in our fork as well (also, no PR, but the current source is on the https://github.com/soider/apiserver-network-proxy/tree/node-to-kas-proxying) to support server side handling of the traffic.

irozzo-1A:
> @irozzo-1A @PratikDeoghare have you had it already implemented? we have some changes in our fork as well (also, no PR, but the current source is on the https://github.com/soider/apiserver-network-proxy/tree/node-to-kas-proxying) to support server side handling of the traffic.

Yes, this branch drafts the complete feature: https://github.com/irozzo-1A/apiserver-network-proxy/tree/extend-proxy
But as I mentioned we need to test it, polish it, and create some PRs out of this.

cmd/agent/app/server.go
irozzo-1A: I have no strong opinion on this, I can move it if you prefer.

jkh52:
@irozzo-1A we are interested in getting this merged, where does this stand?

irozzo-1A:
Hi @jkh52, this PR is ready for review. When this is merged we can proceed with the remaining part.

irozzo-1A:
Hey @cheftako, would it be possible to have another pass here?

tallclair:
Does this PR work as is? I don't see any server-side changes, and the server doesn't currently handle `DIAL_REQ` packets from the backend.

I'd like to propose an alternative approach: rather than changing the protocols to support dialing in both directions, what if you just added an agent on the control-plane node alongside the apiserver, and simply used the konnectivity-client to connect from the node side? A couple details:
- This might require additional backend selection logic to select the right agent backend for each direction, I'm not sure if the DestHost or DefaultRoute backend managers can handle this today.
- For the client connection, this could be implemented as a separate proxy component, or packaged into the agent binary. I'd suggest implementing it as a library, and starting with a separate command. Once it's proven out with that approach, we could consider building it into the agent. This will also help ensure that it's loosely coupled to the existing agent logic.

irozzo-1A:
> Does this PR work as is? I don't see any server-side changes, and the server doesn't currently handle `DIAL_REQ` packets from the backend.

Indeed, this only handles the Agent part, the Server part is almost ready though, it was waiting for this PR to enter. See:
https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/176#issuecomment-976640650

> I'd like to propose an alternative approach: rather than changing the protocols to support dialing in both directions, what if you just added an agent on the control-plane node alongside the apiserver, and simply used the konnectivity-client to connect from the node side? A couple details:
> - This might require additional backend selection logic to select the right agent backend for each direction, I'm not sure if the DestHost or DefaultRoute backend managers can handle this today.
> - For the client connection, this could be implemented as a separate proxy component, or packaged into the agent binary. I'd suggest implementing it as a library, and starting with a separate command. Once it's proven out with that approach, we could consider building it into the agent. This will also help ensure that it's loosely coupled to the existing agent logic.

I presume that the spirit of your proposal is doing something less intrusive or risky, I would like to point out that:
- The code path to handle the nodes->control_plane direction is protected by a feature gate.
- The messages are unchanged. The only protocol change is that I used the connection ID parity to determine the direction. This is not clean and it also means that server is no backward compatible when the feature is activated, but the problem can be solved by introducing an optional direction flag on the data messages. I was waiting for a feedback on this point.
This to say that IMO the risk of regression is low and the clean-up should be fairly easy in case the feature is dismissed.

That said, I think the proposal is interesting, I see the following downsides though:
- Increased operational complexity, we need to deploy the proxies plus the control-plane agents with a specific configuration.
- We should add filtering logic to make sure that who has access to the proxy, does not gain full access to control plain node net
- We should IMO make sure that the default selection logic excludes somehow the control plane agents, otherwise the risk of failure due to misconfiguration is high.
- Increased resource consumption, it would double the TCP connections between nodes and server in the worst case.

Also, I presume if we decide to go the way you proposed, we should update the KEP, and going through another validation, right?

tallclair:
> I presume that the spirit of your proposal is doing something less intrusive or risky

It's not only about reducing risk, it's also about minimizing complexity. It looks like your changes reimplement some of the agent functionality in the server, and vice-versa. Even if the changes are behind a feature gate, the additional code is still a maintenance burden.

I think my proposal reduces the amount of code by reusing existing mechanisms, but it also keeps the architecture conceptually simpler. There is only one connection type, dials flow in one direction, and we don't need to branch data packet handling based on connection ID divisibility.

> Increased operational complexity, we need to deploy the proxies plus the control-plane agents with a specific configuration.

Yes, but that also comes with more flexibility. For example if your proxy-server is on a different node than the apiserver.

> We should add filtering logic to make sure that who has access to the proxy, does not gain full access to control plain node net

Isn't this also true with your current approach? It looks like the server gets the dial address from the DialReq?

> We should IMO make sure that the default selection logic excludes somehow the control plane agents, otherwise the risk of failure due to misconfiguration is high.

Yes, I agree this is the biggest downside, but this is also aligned with other changes that we want to make to have smarter backend selection. I think this should work with `DestHost` selection as long as you don't have overlapping IPs in your control plane & cluster networks (or maybe you can just use localhost). Another option would be to make it configurable by source, so that agents get backends selected from a different pool. I'm not sure exactly what that would look like.

A benefit to this approach is this added complexity is siloed off to the backend manager component, and doesn't affect the main request code paths.

> Increased resource consumption, it would double the TCP connections between nodes and server in the worst case.

The proxy-server to control-plane agent connection is already multiplexed, so that connection overhead should be minimal. It does add more frontend connections since it's not reusing the backend connection from the worker nodes, but we want to make that multiplexed anyway.

---

As for next steps, I'll talk this over with some of the other konnectivity maintainers.

irozzo-1A:
Thanks @tallclair, let me share my final thoughts while waiting for an update about the next steps.

> I think my proposal reduces the amount of code by reusing existing mechanisms, but it also keeps the architecture conceptually simpler. There is only one connection type, dials flow in one direction, and we don't need to branch data packet handling based on connection ID divisibility.

Probably you're right but it's not completely obvious to me, IMO it largely depends on the changes required by the agent-selection logic. Also, the proxy will look like a hybrid between the current server and the agent, it's possible to factorize code to limit duplications, but the same could be done with the current approach.
The problem of determining the direction is definitely one of the bigger downsides, IMO having an optional direction flag would be already a bit better.

> Yes, but that also comes with more flexibility. For example if your proxy-server is on a different node than the apiserver.

The only advantage I see would be being able to scale differently server and control-plane agents, but I don't know how interesting this would be.

> > We should add filtering logic to make sure that who has access to the proxy, does not gain full access to control plain node net
> 
> Isn't this also true with your current approach? It looks like the server gets the dial address from the DialReq?

It's totally true, the main difference is that this filtering logic was supposed to be on the server, which is the component that is supposed to be on the control-plane side. Anyway, I think it would not be a bad thing to be able to constraint possible destinations in the nodes network as well.

> > We should IMO make sure that the default selection logic excludes somehow the control plane agents, otherwise the risk of failure due to misconfiguration is high.
> 
> Yes, I agree this is the biggest downside, but this is also aligned with other changes that we want to make to have smarter backend selection. I think this should work with `DestHost` selection as long as you don't have overlapping IPs in your control plane & cluster networks (or maybe you can just use localhost). Another option would be to make it configurable by source, so that agents get backends selected from a different pool. I'm not sure exactly what that would look like.
>

I think that using `DestHost` would imply that the node network proxies know the hosts where the control-plane agents reside, which may vary, and anyway it represents a configuration burden. I would see better a pool selection mechanism.
Also, it may be useful to avoid control-plane agents getting selected when using current strategies, because:
* we may have clashes with `DestHost` as you mentioned
* default strategy basically would be broken
This would require something to signal that the agent is on control-plane network and some special handling of this case.

> A benefit to this approach is this added complexity is siloed off to the backend manager component, and doesn't affect the main request code paths.
> 
> > Increased resource consumption, it would double the TCP connections between nodes and server in the worst case.
> 
> The proxy-server to control-plane agent connection is already multiplexed, so that connection overhead should be minimal. It does add more frontend connections since it's not reusing the backend connection from the worker nodes, but we want to make that multiplexed anyway.

Yeah, I don't know how big the problem would be. In a 5K nodes cluster with an agent per node, we would pass from 5K to 10K persistent connections to the server. But, probably in a 5K cluster, it would not be recommended to have 1 agent per node. Currently, the way traffic is vehiculated from the nodes network to the control plane network assumes that there is 1 agent per node, but decupling agents from node network proxies would reduce the problem to the proxies only.

tallclair:
I spoke with a few of the konnectivity maintainers today, and they agreed they're interested in further exploring this path, and also that that means a KEP update. Please tag me on the PR if/when you're able to get to that.

---

> Also, the proxy will look like a hybrid between the current server and the agent, it's possible to factorize code to limit duplications, but the same could be done with the current approach.

I think we can just reuse the konnectivity-client, so the cluster network listener looks pretty simple. This is oversimplified, but here's approximately what your `handleConnection` function could become:

```go
import (
	// ...
	kclient "sigs.k8s.io/apiserver-network-proxy/konnectivity-client/pkg/client"
)

func (a *Client) handleConnection(protocol, address string, conn net.Conn) error {
	// ...

	opts := []grpc.DialOption{ // copied from egress_selector.go
		grpc.WithBlock(),
		grpc.WithReturnConnectionError(),
		grpc.WithTimeout(30 * time.Second), // matches http.DefaultTransport dial timeout
		// grpc.WithTransportCredentials(insecure.NewCredentials()),
	}
	tunnel, err := kclient.CreateSingleUseGrpcTunnel(context.TODO(), serverAddress, opts...)
	if err != nil {
		return err
	}
	serverConn, err := tunnel.DialContext(context.TODO(), "tcp", address)
	if err != nil {
		return err
	}

	go io.Copy(clientConn, serverConn) // TODO: handle error
	go io.Copy(serverConn, clientConn)

	return nil
}
```

PR #179, opened 2021-03-03
Make dial request processing asynchonous.

ScheererJ:
This pull request fixes the issue if the net.Dial call based on a dial request blocks the receive loop and with that all communication from the proxy server. We have seen instances where all traffic was blocked for roughly 10 minutes.
The problem is solved by first making the dial request processing asynchronous. Secondly, we added a timeout to the net.Dial call to the target.

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -38,6 +38,8 @@ import (
 	"sigs.k8s.io/apiserver-network-proxy/proto/header"
 )
 
+const dialTimeout = 5 * time.Second
+
 // connContext tracks a connection from agent to node network.
 type connContext struct {
 	conn      net.Conn
@@ -361,7 +363,7 @@ func (a *Client) Serve() {
 			resp.GetDialResponse().Random = dialReq.Random
 
 			start := time.Now()
-			conn, err := net.Dial(dialReq.Protocol, dialReq.Address)
+			conn, err := net.DialTimeout(dialReq.Protocol, dialReq.Address, dialTimeout)
 			if err != nil {
 				resp.GetDialResponse().Error = err.Error()
 				if err := a.Send(resp); err != nil {


ScheererJ:
Signed CLA.

ydp:
Hi @ScheererJ I am learning the code base, does your change just move all the action in `PacketType_DIAL_REQ` into a go routine, which solved your problem?

ScheererJ:
> Hi @ScheererJ I am learning the code base, does your change just move all the action in `PacketType_DIAL_REQ` into a go routine, which solved your problem?

As mentioned in the pull request description above, the change essentially consists of only two changes, but the resulting indentation make the change look bigger than it is. We make the processing of `PacketType_DIAL_REQ` asynchronous by using a go routine. In addition to that, we added a timeout to the net.Dial call.

caesarxuchao:
@ydp if you add "?w=1" in to the url, you can see the diff with all the whitespace differences ignored, that would make this change easier to review.

@ScheererJ the dial timeout looks good. Moving the dial to a separate goroutine is reasonable as well. One concern I have is that if a goroutine is spawned to dial, and before the dial completes, the proxy server forwards a `PACKET_CLOSE_REQ`, then because the connManager doesn't have the connection yet, then the code will go to this [else block](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/847b98331f17962157ed3b56acf095e1d297d045/pkg/agent/client.go#L429-L440) without calling `ctx.cleanup()`. That would lead to leaking goroutines of `removeToProxy` and `proxyToRemote`, and leak the network connection if there is no timeout mechanism built there.

Perhaps we should bookkeep the pending dials, and cleanup pending dials upon receiving `PACKET_CLOSE_REQ`.

ScheererJ:
@caesarxuchao do you suggest to have separate book keeping of the pending dials or simply move the `a.connManager.Add(connID, ctx)` call before the dial call and ensure that the cleanFunc handles the situation properly by synchronizing with the dial (ensuring it finished)?

caesarxuchao:
I think we will have to add a separate book keeping, because before the dial returns, ctx.conn is not available.

caesarxuchao:
something like this:

```golang
type connContext struct {
  // add dialDone
  dialDone chan struct{}
  conn      net.Conn
  cleanFunc func()
  ...
  
}

// in client.Serve()
...
case PACKET_DIAL_REQUEST:
  dialDone := new(chan struct{})
  ctx := connContext {
    dialDone: dialDone,
  }
  go func {
    conn, err := net.DialTimeout(...)
    ctx.conn = conn
    close(dialDone)
  }
  ctx.cleanFunc := func() {
    <- dialDone
    if ctx.conn != nil {
       ctx.conn.Close()
    }
    ...
  }
```

ScheererJ:
I added the proposed change in a new commit, but I did not add the `if ctx.conn != nil` check as it cannot be nil after a successful dial.
Please advise in case I missed something.

ScheererJ:
and now it should work again...

pkg/agent/client.go
```diff
@@ -395,18 +389,34 @@ func (a *Client) Serve() {
...
+				}
+				a.connManager.Add(connID, ctx)
 
-			resp.GetDialResponse().ConnectID = connID
-			if err := a.Send(resp); err != nil {
-				klog.ErrorS(err, "stream send failure")
-				continue
-			}
+				start := time.Now()
+				conn, err := net.DialTimeout(dialReq.Protocol, dialReq.Address, 5 * time.Second)
```
caesarxuchao, line 396: nit: define a file-level constant for the timeout.

In the future we might make it a flag.

caesarxuchao (commented):
Thanks @ScheererJ.

We will need unit tests to make sure we don't introduce goroutine leaks. Let me know if you have time for adding tests. If not, I can add tests in followups.

pkg/agent/client.go
```diff
@@ -351,40 +352,33 @@ func (a *Client) Serve() {
...
+					<- dialDone
 					klog.V(4).InfoS("close connection", "connectionID", connID)
 					resp := &client.Packet{
 						Type:    client.PacketType_CLOSE_RSP,
 						Payload: &client.Packet_CloseResponse{CloseResponse: &client.CloseResponse{}},
 					}
 					resp.GetCloseResponse().ConnectID = connID
 
-					err := conn.Close()
+					err := ctx.conn.Close()
```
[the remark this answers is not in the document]
ScheererJ: Good catch.
I added `a.connManager.Delete(connID)` to the error handling of a failed dial attempt, but this indeed is still a race with the close request handling. Therefore, I now added the nil check as you proposed with the latest commit.

pkg/agent/client.go
ScheererJ: Done with the latest commit.

ScheererJ:
> Thanks @ScheererJ.
> 
> We will need unit tests to make sure we don't introduce goroutine leaks. Let me know if you have time for adding tests. If not, I can add tests in followups.

@caesarxuchao I would prefer if you could add the tests as a follow-up.

rtheis:
Posted test results with this fix to the associated issue: https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/180#issuecomment-852361629

relyt0925:
cc @cheftako is this still possible to get in as well since it's an important fix?

pkg/agent/client.go
```diff
@@ -351,62 +354,73 @@ func (a *Client) Serve() {
 
 		switch pkt.Type {
 		case client.PacketType_DIAL_REQ:
-			klog.V(4).Infoln("received DIAL_REQ")
-			resp := &client.Packet{
-				Type:    client.PacketType_DIAL_RSP,
-				Payload: &client.Packet_DialResponse{DialResponse: &client.DialResponse{}},
-			}
+			go func() {
```
caesarxuchao, line 357: If we move the entire handling of DIAL_REQ into a goroutine like this, we can still enter a race with the handling of CLOSE_REQ that I mentioned in https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/179#issuecomment-839283626.

How about we only move the net.Dial into a goroutine? That way `a.connManager.Add` is called sequentially and that would prevent the race. This is also what I suggested in https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/179#issuecomment-841347676.

caesarxuchao (commented):
Sorry for dropping the ball. I took another pass and left another comment about the race.

Given that @jkh52 and his team are working hard stabilizing ANP right now, we need to be extra cautious merging this PR.

How about we split this PR into two parts? In the first part we only add a 10s timeout on the net.Dial, so that we can avoid a 10min block. And in the second part we make this more risky goroutine change.

Hopefully in the future the ANP repo can have a multi-branch structure like k/k so that we can stabilize the code in one branch and merge the more risky PRs in another branch.

pkg/agent/client.go
ScheererJ: @caesarxuchao So what you are suggesting is that the go routine should start after the `a.connManager.Add` call. Right?
In the beginning the go routine simply started shortly before the `net.Dial` call, but after your suggested refactoring we can indeed move the start of the go routine to a later point in time, i.e. directly before the `net.Dial` call.

ScheererJ:
> Sorry for dropping the ball. I took another pass and left another comment about the race.
> 
> Given that @jkh52 and his team are working hard stabilizing ANP right now, we need to be extra cautious merging this PR.
> 
> How about we split this PR into two parts? In the first part we only add a 10s timeout on the net.Dial, so that we can avoid a 10min block. And in the second part we make this more risky goroutine change.
> 
> Hopefully in the future the ANP repo can have a multi-branch structure like k/k so that we can stabilize the code in one branch and merge the more risky PRs in another branch.

@caesarxuchao Should I only include the 10s timeout, i.e. the minimal change, into the first pull request or also include the reordering and change with regards to `a.connManager.Add`?

relyt0925:
This is just a note that without this it appears the Kubernetes e2e tests will not pass when Konnectivity is in use. cc @caesarxuchao 

I think this is critical toward stabilization of the system. See @rtheis's comment https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/179#issuecomment-852362599 for more details.

I know we (IBM + RedHat) have had to carry this patch in order to utilize Konnectivity in our systems.

pkg/agent/client.go
caesarxuchao: Yeah, the goroutine should start after the `a.connManager.Add` call. Or in other words, "move the start of the go routine to a later point in time, i.e. directly before the net.Dial call".

caesarxuchao (commented):
Sorry for not being responsive. I wan't monitoring this PR closely. Feel free to ping me on Slack.

caesarxuchao:
> Should I only include the 10s timeout, i.e. the minimal change, into the first pull request or also include the reordering and change with regards to a.connManager.Add?

In the first PR, let's only include the 10s timeout.

In the second PR, let's include the goroutine change, including the reordering and change with regards to a.connManager.Add.

relyt0925:
@ScheererJ do you still have time to make the changes? If not I can see if we can find someone to continue your great work here.

ScheererJ:
Added #252 as the second part of this improvement with the go routine change.

@relyt0925 I will adapt this pull request now to only include the timeout change.

ScheererJ:
Now, this pull request only includes the timeout change and the extended asynchronous processing via go routine is done in #252.

PR #185, opened 2021-03-24
Improve Makefile and README.md for new users

MaxRenaud:
The README.md lists mockgen as an explicit dependency
The Makefile has the following changes:

- Create $GOPATH/src if it doesn't exist. Without the dir, protoc fails
- ```make gen``` uses the protoc output .go files to build the licensed file. ```make clean``` deletes them from the repo and thus ```make gen``` never succeeded.
- Performs "go mod download" before generating files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,20 +85,30 @@ lint:
 	$(INSTALL_LOCATION)/golangci-lint run --no-config --disable-all --enable=gofmt,golint,gosec --fix --verbose --timeout 3m
 
 ## --------------------------------------
+## Go
+## --------------------------------------
+
+.PHONY: mod-download
+mod-download:
+	go mod download
+
+## --------------------------------------
 ## Proto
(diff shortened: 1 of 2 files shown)


MaxRenaud:
/retest
client_test.go:175: client.connContext not released
seems unrelated.

cheftako:
> /retest
> client_test.go:175: client.connContext not released
> seems unrelated.

Weirdly I see the prow robot complaining at you but the test shows as passing in the Review requested section. I don't see the test as having a rerun requested.

cheftako:
> > /retest
> > client_test.go:175: client.connContext not released
> > seems unrelated.
> 
> Weirdly I see the prow robot complaining at you but the test shows as passing in the Review requested section. I see you requested the test be rerun. So it seems like the test is flaky, which is a bit worrying.

PR #192, opened 2021-04-02
Add flags to allow tuning k8s client throttling.

jkh52:
I have seen that if enough agents exist, and one of a set of masters is unavailable (so every agent is sync-ing), proxy server goroutines grows unbounded. With pprof I find that client-go throttling (of token reviews) is where they are blocked.

In my repro, if I replace the client-go defaults (5 qps, 10 burst) with higher numbers, it mitigates the issue.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -121,6 +121,10 @@ type ProxyRunOptions struct {
 	authenticationAudience string
 	// Path to kubeconfig (used by kubernetes client)
 	kubeconfigPath string
+	// Client maximum QPS.
+	kubeconfigQPS float32
+	// Client maximum burst for throttle.
+	kubeconfigBurst int
 
 	// Proxy strategies used by the server.
 	// NOTE the order of the strategies matters. e.g., for list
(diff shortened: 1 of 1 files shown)


cmd/server/main.go
```diff
@@ -121,6 +121,10 @@ type ProxyRunOptions struct {
 	authenticationAudience string
 	// Path to kubeconfig (used by kubernetes client)
 	kubeconfigPath string
+	// Client maximum QPS.
+	kubeconfigQps float32
```
Jefftree, line 125: Stylistically capitalizing QPS is probably more correct.

Jefftree (commented):
One nit comment, otherwise LGTM.

cmd/server/main.go
```diff
@@ -375,6 +385,14 @@ func (p *Proxy) run(o *ProxyRunOptions) error {
 			return fmt.Errorf("failed to load kubernetes client config: %v", err)
 		}
 
+		if o.kubeconfigQps != 0 {
+			klog.V(1).Infoln("Setting k8s client QPS: %v", o.kubeconfigQps)
```
Jefftree, line 389: Also from the build:

```
# sigs.k8s.io/apiserver-network-proxy/cmd/server
cmd/server/main.go:389:4: Infoln call has possible formatting directive %v
cmd/server/main.go:393:4: Infoln call has possible formatting directive %v
```

jkh52:
Thanks.

cmd/server/main.go
```diff
@@ -154,6 +158,8 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.agentNamespace, "agent-namespace", o.agentNamespace, "Expected agent's namespace during agent authentication (used with agent-service-account, authentication-audience, kubeconfig).")
 	flags.StringVar(&o.agentServiceAccount, "agent-service-account", o.agentServiceAccount, "Expected agent's service account during agent authentication (used with agent-namespace, authentication-audience, kubeconfig).")
 	flags.StringVar(&o.kubeconfigPath, "kubeconfig", o.kubeconfigPath, "absolute path to the kubeconfig file (used with agent-namespace, agent-service-account, authentication-audience).")
+	flags.Float32Var(&o.kubeconfigQPS, "kubeconfig-qps", o.kubeconfigQPS, "Maximum client QPS (used with authentication).")
```
caesarxuchao, line 161: nit: there are multiple authentications in the system. Perhaps "proxy server uses this client to authenticate agent tokens" to be more specific.

caesarxuchao:
A nit. LGTM otherwise.

cmd/server/main.go
jkh52: Done.

PR #197, opened 2021-04-27
Allow agent to set ALPN protocol

jveski:
In some cases it is preferable to use a common network path i.e. load balancer rule to expose both proxy server and kube-apiserver.

It is currently possible to do so by managing two separate DNS records - one for apiserver, one for proxy server - and routing connections between services using a proxy that supports SNI-based "vhost-style" routing. This works well unless something obstructs connections to the (new) proxy hostname.

When agent nodes and apiserver run in separate networks, traffic from nodes to apiserver/proxy server might traverse a firewall. In such cases it's common to use a FQDN rule to allow agent nodes to access apiserver. But most firewalls implement FQDN rules using SNI. So without changes to firewall configurations, proxy agent->server connections using the new hostname will be blocked even if they resolve back to the same address. For managed k8s offerings, it might not be feasible to update any firewalls to allow access, so another option would be preferable.

The only other approach I can think of that doesn't involve reverse proxying all apiserver and proxy requests is to use a custom ALPN protocol. The implementation is simple: expose the TLS config's `NextProtos` field as a CLI flag. Then TCP proxies/load balancers (Envoy included) can make routing decisions given only the TLS client hello while still using a single SNI hostname.

### DIFF
--- a/pkg/util/certificates.go
+++ b/pkg/util/certificates.go
@@ -39,7 +39,7 @@ func getCACertPool(caFile string) (*x509.CertPool, error) {
 }
 
 // GetClientTLSConfig returns tlsConfig based on x509 certs
-func GetClientTLSConfig(caFile, certFile, keyFile, serverName string) (*tls.Config, error) {
+func GetClientTLSConfig(caFile, certFile, keyFile, serverName string, protos []string) (*tls.Config, error) {
 	certPool, err := getCACertPool(caFile)
 	if err != nil {
 		return nil, err
@@ -50,6 +50,7 @@ func GetClientTLSConfig(caFile, certFile, keyFile, serverName string) (*tls.Conf
 		tlsConfig := &tls.Config{
 			RootCAs:    certPool,
 			MinVersion: tls.VersionTLS12,
+			NextProtos: protos,
 		}
 		return tlsConfig, nil
 	}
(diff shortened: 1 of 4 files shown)


jveski:
I signed it

jveski:
I signed it

jveski:
Hey @caesarxuchao @Sh4d1, sorry to bug but could I get a review?

pkg/util/certificates.go
```diff
@@ -50,6 +50,7 @@ func GetClientTLSConfig(caFile, certFile, keyFile, serverName string) (*tls.Conf
 		tlsConfig := &tls.Config{
 			RootCAs:    certPool,
 			MinVersion: tls.VersionTLS12,
+			NextProtos: protos,
```
cheftako, line 53: Can we rewire this to only set the field with protos if protos was set as an input? I'd rather not modify the tlsConfig default value if the user did not set the --alpn-proto parameter.

cheftako:
Hi @jveski, mostly looks good. One minor point on init just to be safe.

pkg/util/certificates.go
jveski: I tested to make sure that the defaults are still used given a zero length NextProtos, but you're right — better to be safe here. Fixed.

jveski:
Thanks for the review @cheftako!

PR #198, opened 2021-04-28
Represent connection state in a pod condition

### LINKED ISSUE #196: Publish agent errors as events
Since getting pod logs is dependent on the agent, troubleshooting a disconnected agent is difficult.

One way to address this would be to publish k8s events for agent errors. If folks think this is a good idea I'd be happy to implement.

jveski:
Resolves #196 

Provide signal to users and other controllers by managing a pod condition that contains the current server connection state.

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -34,6 +34,8 @@ import (
 	"github.com/spf13/pflag"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/apiserver-network-proxy/pkg/agent"
 	"sigs.k8s.io/apiserver-network-proxy/pkg/util"
@@ -90,6 +92,14 @@ type GrpcProxyAgentOptions struct {
 
 	// file contains service account authorization token for enabling proxy-server token based authorization
 	serviceAccountTokenPath string
+
+	kubeconfigPath  string
+	kubeconfigQPS   float32
+	kubeconfigBurst int
+
+	enablePodCondition bool
+	podNamespace       string
+	podName            string
 }
 
 func (o *GrpcProxyAgentOptions) ClientSetConfig(dialOptions ...grpc.DialOption) *agent.ClientSetConfig {
@@ -102,6 +112,9 @@ func (o *GrpcProxyAgentOptions) ClientSetConfig(dialOptions ...grpc.DialOption) 
 		SyncIntervalCap:         o.syncIntervalCap,
 		DialOptions:             dialOptions,
 		ServiceAccountTokenPath: o.serviceAccountTokenPath,
+		EnablePodCondition:      o.enablePodCondition,
+		PodNamespace:            o.podNamespace,
+		PodName:                 o.podName,
 	}
 }
 
@@ -123,6 +136,10 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.DurationVar(&o.syncIntervalCap, "sync-interval-cap", o.syncIntervalCap, "The maximum interval for the syncInterval to back off to when unable to connect to the proxy server")
 	flags.StringVar(&o.serviceAccountTokenPath, "service-account-token-path", o.serviceAccountTokenPath, "If non-empty proxy agent uses this token to prove its identity to the proxy server.")
 	flags.StringVar(&o.agentIdentifiers, "agent-identifiers", o.agentIdentifiers, "Identifiers of the agent that will be used by the server when choosing agent. N.B. the list of identifiers must be in URL encoded format. e.g.,host=localhost&host=node1.mydomain.com&cidr=127.0.0.1/16&ipv4=1.2.3.4&ipv4=5.6.7.8&ipv6=:::::")
+	flags.StringVar(&o.kubeconfigPath, "kubeconfig", o.kubeconfigPath, "Absolute path to the kubeconfig file. Only required when --update-pod-status is set.")
+	flags.Float32Var(&o.kubeconfigQPS, "kubeconfig-qps", o.kubeconfigQPS, "Maximum client QPS")
+	flags.IntVar(&o.kubeconfigBurst, "kubeconfig-burst", o.kubeconfigBurst, "Maximum client burst")
+	flags.BoolVar(&o.enablePodCondition, "enable-pod-condition", o.enablePodCondition, "When set, manage a condition to represent connection status in the pod specified by POD_NAME and POD_NAMESPACE")
 	return flags
 }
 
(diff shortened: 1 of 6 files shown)


pkg/agent/clientset.go
```diff
@@ -24,10 +24,20 @@ import (
...
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 )
 
+const (
+	connectedPodConditionType      = "k8s.io/apiserver-network-proxy-connected"
+	connectedPodConditionReason    = "Connected"
+	disconnectedPodConditionReason = "Disconnected"
+	partitionPodConditionReason    = "Partition"
```
jveski, line 38: Not 100% on this terminology. Maybe `PartiallyConnected`?

cmd/agent/main.go
```diff
@@ -209,6 +232,12 @@ func newGrpcProxyAgentOptions() *GrpcProxyAgentOptions {
 		probeInterval:           1 * time.Second,
 		syncIntervalCap:         10 * time.Second,
 		serviceAccountTokenPath: "",
+		kubeconfigPath:          "",
+		kubeconfigQPS:           0,
+		kubeconfigBurst:         0,
+		enablePodCondition:      false,
+		podNamespace:            os.Getenv("POD_NAMESPACE"),
```
jveski, line 239: Env vars instead of flags for easy use with fieldRef values.

pkg/agent/clientset.go
```diff
@@ -210,3 +236,72 @@ func (cs *ClientSet) shutdown() {
...
+		klog.Warningf("error while getting pod: %s", err)
+		return
+	}
+
+	updatedPod := cs.appendPodCondition(pod, getPodCondition(cs.ClientsCount(), cs.serverCount))
+	if updatedPod == nil {
+		return
+	}
+
+	_, err = cs.k8sClient.CoreV1().Pods(updatedPod.Namespace).UpdateStatus(updatedPod)
```
jveski, line 256: Should we patch instead? I've seen controllers that go both ways so not sure.

tests/ha_proxy_server_test.go
```diff
@@ -37,26 +37,33 @@ func (lb *tcpLB) handleConnection(in net.Conn, backend string) {
...
+				return
+			default:
+			}
+			conn, err := ln.Accept()
+			if err != nil {
+				log.Printf("failed to accept: %s", err)
+				continue
+			}
+			// go lb.handleConnection(conn, lb.randomBackend())
+			if lb.backendCount() == 0 {
```
jveski, line 58: Avoid panic when passing `0` to rand.Intn

jveski:
cc @cheftako

jveski:
@dberkov @Jefftree - can I get a review on this? No particular rush, just want to make sure you're aware of this PR.

cmd/agent/main.go
cheftako: I have no issue with initializing podNamespace and podName using os.Getenv. I'm not certain these environment variables will actually be set in all environments. As such we should still have command line flags to allow for over-rides if needed.

cmd/agent/main.go
```diff
@@ -112,6 +125,10 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
...
 	flags.StringVar(&o.serviceAccountTokenPath, "service-account-token-path", o.serviceAccountTokenPath, "If non-empty proxy agent uses this token to prove its identity to the proxy server.")
 	flags.StringVar(&o.agentIdentifiers, "agent-identifiers", o.agentIdentifiers, "Identifiers of the agent that will be used by the server when choosing agent. N.B. the list of identifiers must be in URL encoded format. e.g.,host=localhost&host=node1.mydomain.com&cidr=127.0.0.1/16&ipv4=1.2.3.4&ipv4=5.6.7.8&ipv6=:::::")
+	flags.StringVar(&o.kubeconfigPath, "kubeconfig", o.kubeconfigPath, "Absolute path to the kubeconfig file. Only required when --update-pod-status is set.")
+	flags.Float32Var(&o.kubeconfigQPS, "kubeconfig-qps", o.kubeconfigQPS, "Maximum client QPS")
+	flags.IntVar(&o.kubeconfigBurst, "kubeconfig-burst", o.kubeconfigBurst, "Maximum client burst")
+	flags.BoolVar(&o.enablePodCondition, "enable-pod-condition", o.enablePodCondition, "When set, manage a condition to represent connection status in the pod specified by POD_NAME and POD_NAMESPACE")
```
cheftako, line 131: We should add validation that if enablePodCondition is true that any required fields, such as podname and namespace are set. We should also check at a minimum that kubeconfigPath is set to an existing path and readable.

cmd/agent/main.go
```diff
@@ -234,8 +263,28 @@ func (a *Agent) run(o *GrpcProxyAgentOptions) error {
 		return fmt.Errorf("failed to validate agent options with %v", err)
 	}
 
+	var k8sClient *kubernetes.Clientset
+	if o.kubeconfigPath != "" || o.enablePodCondition {
```
cheftako, line 267: If we do the previously mentioned validations, then the "|| o.enablePodCondition" check should be unnecessary. This is good as it will remove a confusing build config from flags edge case where kubeconfigPath == "". It also means we fail earlier on several invalid configurations, which is good.

pkg/agent/clientset.go
cheftako: I'd prefer something like PartiallyConnected. "partition" suggests a split brain problem and we (ANP agent) are not in a position to detect that. (Actually we almost might be but that is certainly not what this condition is indicating)

pkg/agent/clientset.go
cheftako: I would avoid patch (At least the old JSON one). If you can get this working with Server Side Apply, that would probably be an improvement.

cheftako: /cc @jpbetz

pkg/agent/clientset.go
jveski: Agreed that partition is a misnomer here. Changed to PartiallyConnected.

pkg/agent/clientset.go
jveski: Let's leave as-is then. I'd like to avoid taking a dependency on SSA for now since some users might still have its release gate disabled.

cmd/agent/main.go
jveski: Good call. Added for pod name/namespace. But the kubeconfig is actually optional since `clientcmd.BuildConfigFromFlags` defaults to the in-cluster config.

cmd/agent/main.go
jveski: I agree that this logic is precarious, but how should we handle cases where the kubeconfig flag is intentionally left unset to instead use the in cluster config?

jveski:
Ready for re-review whenever you are @cheftako

PR #199, opened 2021-05-01
Trivial: make agent log level more consistent with server.

jkh52:
Log network bytes detail at level 5, not 4. The default level is 4, and this log can be very noisy.

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -455,7 +455,7 @@ func (a *Client) remoteToProxy(connID int64, ctx *connContext) {
 
 	for {
 		n, err := ctx.conn.Read(buf[:])
-		klog.V(4).InfoS("received data from remote", "bytes", n, "connID", connID)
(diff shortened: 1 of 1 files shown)


pkg/agent/client.go
```diff
@@ -455,7 +455,7 @@ func (a *Client) remoteToProxy(connID int64, ctx *connContext) {
 
 	for {
 		n, err := ctx.conn.Read(buf[:])
-		klog.V(4).InfoS("received data from remote", "bytes", n, "connID", connID)
```
cheftako, line 458: Default log level of 4 was originally only supposed to be in the test client. Apparently I missed that in a recent review and we switched to 4 everywhere. Still trying to decide if its too late to switch back (backward compatibility problems). For now going to go ahead and approve. This switch still adheres to https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md as we can (and should) consider this to be trace level logging.

PR #203, opened 2021-05-11
Always remove client when Serve() returns

jveski:
Currently a client will be removed from the clientset when sending/receiving over it returns a non-EOF error, or when the probe observes `conn.GetState() != connectivity.Ready`. On EOF, the `Serve` loop will return without removing the client, but the probe will typically fail soon after so it gets cleaned up eventually.

But the client will never be cleaned up in cases where the probe doesn't fail. And since `Serve` is no longer running the client will count against the number of expected clients without actually being able to proxy connections. This occurs any time a connection is broken and reestablished by the underlying grpc client before the next probe.

I personally find the way connections are removed from the clientset confusing and error prone, so I'm proposing we defer a call to `RemoveClient` at the beginning of `Serve` and remove the (now unnecessary) existing calls in `Recv`/`Send`.

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -243,7 +243,6 @@ func (a *Client) Send(pkt *client.Packet) error {
 	err := a.stream.Send(pkt)
 	if err != nil && err != io.EOF {
 		metrics.Metrics.ObserveFailure(metrics.DirectionToServer)
-		a.cs.RemoveClient(a.serverID)
 	}
 	return err
 }
@@ -255,7 +254,6 @@ func (a *Client) Recv() (*client.Packet, error) {
 	pkt, err := a.stream.Recv()
 	if err != nil && err != io.EOF {
 		metrics.Metrics.ObserveFailure(metrics.DirectionFromServer)
-		a.cs.RemoveClient(a.serverID)
 	}
 	return pkt, err
 }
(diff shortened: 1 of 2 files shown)


jveski:
No rush but pinging @dberkov @Sh4d1 for a review whenever you get a chance.

caesarxuchao:
/lgtm

@Jefftree @cheftako do you want to double check?

pkg/agent/client.go
```diff
@@ -243,7 +243,6 @@ func (a *Client) Send(pkt *client.Packet) error {
 	err := a.stream.Send(pkt)
 	if err != nil && err != io.EOF {
 		metrics.Metrics.ObserveFailure(metrics.DirectionToServer)
-		a.cs.RemoveClient(a.serverID)
```
cheftako, line 246: Not sure this line is safe to delete here. Send() is called from remoteToProxy() as well as Serve(). RemoveClient() is deliberately idempotent.

pkg/agent/client.go
```diff
@@ -314,6 +312,7 @@ func (a *Client) initializeAuthContext(ctx context.Context) (context.Context, er
 // The requests include things like opening a connection to a server,
 // streaming data and close the connection.
 func (a *Client) Serve() {
+	defer a.cs.RemoveClient(a.serverID)
```
cheftako, line 315: Again RemoveClient() should be idempotent. So not seeing a problem adding this here.

pkg/agent/client.go
jveski: Yeah I think you're right. We should reconnect the client immediately when sending bytes back up to the server returns an error. As it stands, I think this PR would defer reconnection until after the probe fails — assuming it does.

jveski: Fixed.

PR #204, opened 2021-05-12
Add proxy-server current connections metric.

jkh52:


pkg/server/metrics/metrics.go
```diff
@@ -54,16 +56,41 @@ func newServerMetrics() *ServerMetrics {
 		},
 		[]string{},
 	)
+	connections := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "grpc_connections",
+			Help:      "Number of current grpc connections, partitioned by type.",
```
caesarxuchao, line 64: s/by type/by service method

pkg/server/metrics/metrics.go
```diff
@@ -22,11 +22,12 @@ import (
...
 )
 
-type Direction string
-
 const (
 	namespace = "konnectivity_network_proxy"
 	subsystem = "server"
+
+	Frontend = "frontend"
+	Backend = "backend"
```
caesarxuchao, line 30: How about using the actual method names (connect/proxy)?

Frontend/Backend can be confusing.

caesarxuchao (commented):
Nits one names. Otherwise lgtm.

pkg/server/metrics/metrics.go
jkh52: Done.

pkg/server/metrics/metrics.go
jkh52: Good idea, done.

Do you think it's worth adding service name as well, for completeness? (We only have one method per service though.)

caesarxuchao: I think the chance of us adding new methods of the same name is low, so "connect"/"proxy" are probably good enough.

caesarxuchao:
/lgtm

Omitting the "approval" to give @Jefftree or @cheftako a chance to take a look.

pkg/server/metrics/metrics.go
```diff
@@ -54,16 +59,40 @@ func newServerMetrics() *ServerMetrics {
...
 	prometheus.MustRegister(latencies)
-	return &ServerMetrics{latencies: latencies}
+	prometheus.MustRegister(connections)
+	return &ServerMetrics{latencies: latencies, connections: connections}
 }
 
 // Reset resets the metrics.
 func (a *ServerMetrics) Reset() {
 	a.latencies.Reset()
+	a.connections.Reset()
```
cheftako, line 82: Might be missing something but I didn't think one usually reset a gauge.

jkh52: Thanks, fixed.

I don't see any calls to metrics.Metrics.Reset(), is it called implicitly somewhere that I don't realize?

cheftako: Great question. I'm not a prometheus expert but if this is wired in properly I would think prometheus would call it (directly or indirectly) when it grabbed the metrics. Reset is used to stop double counting on counters as they are cumulative results.

PR #211, opened 2021-05-17
Clean up client testing.

cheftako:
pprof shows a leak during testing with built in client.
However the leak seems to be caused by a failure to call
CloseIdleConnections.
Also made sure we sent close message in client at the correct time.
Enhanced client to support parallel requests.
Added connectionId to a few close cases.
Standardized on the "connectionID" label in log messages.

### DIFF
--- a/cmd/client/main.go
+++ b/cmd/client/main.go
@@ -27,6 +27,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"sync"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -64,20 +65,22 @@ func main() {
 }
 
 type GrpcProxyClientOptions struct {
-	clientCert   string
-	clientKey    string
-	caCert       string
-	requestProto string
-	requestPath  string
-	requestHost  string
-	requestPort  int
-	proxyHost    string
-	proxyPort    int
-	proxyUdsName string
-	mode         string
-	userAgent    string
-	testRequests int
-	testDelaySec int
+	clientCert    string
+	clientKey     string
+	caCert        string
+	requestProto  string
+	requestPath   string
+	requestHost   string
+	requestPort   int
+	proxyHost     string
+	proxyPort     int
+	proxyUdsName  string
+	mode          string
+	userAgent     string
+	testRequests  int
+	testDelaySec  int
+	afterDelaySec int
+	closeIdleConn bool
 }
 
 func (o *GrpcProxyClientOptions) Flags() *pflag.FlagSet {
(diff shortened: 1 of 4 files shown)


cmd/client/main.go
```diff
@@ -241,17 +242,43 @@ func (c *Client) run(o *GrpcProxyClientOptions) error {
...
+			dialer, err := c.getDialer(o)
+			if err != nil {
+				return fmt.Errorf("failed to get dialer for client, got %v", err)
+			}
+			transport := &http.Transport{
+				DialContext: dialer,
+			}
+			client := &http.Client{
+				Transport: transport,
+			}
```
caesarxuchao, line 260: This changes the behavior of the client from reusing the same underlying connection to creating a new connection for each http request. I think both are ok. Do you change the behavior on purpose?

cmd/client/main.go
```diff
@@ -290,7 +322,7 @@ func (c *Client) getUDSDialer(o *GrpcProxyClientOptions) (func(ctx context.Conte
 
 	// Setup signal handler
 	ch := make(chan os.Signal, 1)
-	signal.Notify(ch)
+	defer signal.Notify(ch)
```
caesarxuchao, line 325: Sorry I don't get why we need to change this to `defer`. Could you explain a little bit?

caesarxuchao (commented):
LGTM. Got two questions.

cmd/client/main.go
cheftako: It was. As per the comment header "The apiserver network proxy does not support reusing the same dialer for multiple connections." So was attempting to ensure the test client reflects our primary use case. Up to this point our load was serial so it mattered less but as I was making the requests concurrent it was better to stay consistent, If we want a reuse dialer mode, I'm fine with adding it. I think that needs to be given some thought for the model to be correct. Also we should probably understand the requirement for that use case first.

cmd/client/main.go
cheftako: I was getting proxyConn closing early (i.e. during tunnel creation) some of the time and incorrectly attributed it to this and the following goroutine. Good catch.

cmd/client/main.go
```diff
@@ -96,6 +99,8 @@ func (o *GrpcProxyClientOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.userAgent, "user-agent", o.userAgent, "User agent to pass to the proxy server")
 	flags.IntVar(&o.testRequests, "test-requests", o.testRequests, "The number of times to send the request.")
 	flags.IntVar(&o.testDelaySec, "test-delay", o.testDelaySec, "The delay in seconds between sending requests.")
+	flags.IntVar(&o.afterDelaySec, "after-delay", o.afterDelaySec, "The delay in seconds after sending requests.")
+	flags.BoolVar(&o.closeIdleConn, "close-idle-conn", 	o.closeIdleConn, "Controls if the client calls closeIdleConnections.")
```
Jefftree, line 103: There's a weird space here, is there an extra tab?

cmd/client/main.go
```diff
@@ -170,25 +177,30 @@ func (o *GrpcProxyClientOptions) Validate() error {
...
+		requestPort:   8000,
+		proxyHost:     "localhost",
+		proxyPort:     8090,
+		proxyUdsName:  "",
+		mode:          "grpc",
+		userAgent:     "test-client",
+		testRequests:  1,
+		testDelaySec:  0,
+		afterDelaySec: 0,
+		closeIdleConn: true,
```
Jefftree, line 203: nit: Seems like it would make more sense for the default to be false since that's the default behavior in k/k and in general?

cmd/client/main.go
cheftako: Fixed.

cmd/client/main.go
cheftako: Torn. K/K I believe should be calling closeIdleConnection, but I agree it isn't. In case people are using this as an example, I would rather the default was the correct behavior. If the client doesn't close idle connections, then we have to maintain the tunnel all the way through, along with goroutines for each of the segments. The flag is here to allow that to be easily demonstrated.

PR #213, opened 2021-05-18
fix error msg

silenceper:
fix error msg

### DIFF
--- a/pkg/server/backend_manager.go
+++ b/pkg/server/backend_manager.go
@@ -234,7 +234,7 @@ func (s *DefaultBackendStorage) RemoveBackend(identifier string, idType pkgagent
 		}
 	}
 	if !found {
-		klog.V(1).InfoS("Cannot find connection for agent in backends", "connection", conn, "agentID", agentID)
+		klog.V(1).InfoS("Cannot find connection for agent in backends", "connection", conn, "identifier", identifier)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pkg/server/backend_manager.go
```diff
@@ -234,7 +234,7 @@ func (s *DefaultBackendStorage) RemoveBackend(identifier string, idType pkgagent
 		}
 	}
 	if !found {
-		klog.V(1).InfoS("Cannot find connection for agent in backends", "connection", conn, "agentID", agentID)
+		klog.V(1).InfoS("Cannot find connection for agent in backends", "connection", conn, "identifier", identifier)
```
cheftako, line 237: Could we rephrase this to not mention agent? Eg. "Could not find connection matching identifier to remove"?
Also could you fix the log message on line 205 to indicate failed to remove not failed to add?

pkg/server/backend_manager.go
```diff
@@ -202,7 +202,7 @@ func (s *DefaultBackendStorage) AddBackend(identifier string, idType pkgagent.Id
 // RemoveBackend removes a backend.
 func (s *DefaultBackendStorage) RemoveBackend(identifier string, idType pkgagent.IdentifierType, conn agent.AgentService_ConnectServer) {
 	if !containIDType(s.idTypes, idType) {
-		klog.ErrorS(&ErrWrongIDType{idType, s.idTypes}, "fail to add backend")
+		klog.ErrorS(&ErrWrongIDType{idType, s.idTypes}, "fail to add remove")
```
cheftako, line 205: Can we make this "fail to remove backend"

Issue #214, opened 2021-05-18
upgrade client-go package from v0.17.1 to v.0.19.10+


liufen90:
Client-go package has  newer releases. We should upgrade client-go to newer version.

liufen90:
@cheftako  I have finished upgrade client-go to v0.19.10.

liufen90:
@cheftako  Can you merge the PR #216  and release a new version as soon as possible? Thanks very much.

PR #216, opened 2021-05-19
upgrade client-go package from v0.17.1 to v0.19.10,

liufen90:
upgrade client-go package from v0.17.1 to v.0.19.10, #214


cheftako:
/lgtm
Have you validated that this works with older Kubernetes APIServers (prior to v1.19?)
If not we may need to implement a README for our releases.

liufen90:
> /lgtm
> Have you validated that this works with older Kubernetes APIServers (prior to v1.19?)
> If not we may need to implement a README for our releases.

No, I haven't.

cheftako:
@caesarxuchao may know if this is a concern. I don't think its worth gating this PR on this but it would be nice to know if we should have a README for it. Waiting on Chao's review to approve.

liufen90:
@caesarxuchao please take a look. Thank you.

pkg/server/server.go
```diff
@@ -508,14 +509,14 @@ func getAgentIdentifiers(stream agent.AgentService_ConnectServer) (pkgagent.Iden
...
-func (s *ProxyServer) validateAuthToken(token string) error {
+func (s *ProxyServer) validateAuthToken(ctx context.Context, token string) error {
 	trReq := &authv1.TokenReview{
 		Spec: authv1.TokenReviewSpec{
 			Token:     token,
 			Audiences: []string{s.AgentAuthenticationOptions.AuthenticationAudience},
 		},
 	}
-	r, err := s.AgentAuthenticationOptions.KubernetesClient.AuthenticationV1().TokenReviews().Create(trReq)
+	r, err := s.AgentAuthenticationOptions.KubernetesClient.AuthenticationV1().TokenReviews().Create(ctx, trReq, metav1.CreateOptions{})
```
caesarxuchao, line 519: This is the only use of client-go in the code base. I think this client call works for older apiservers.

caesarxuchao (commented):
/lgtm

cheftako:
For reference its in tag v0.0.19 (Coincidence).

PR #219, opened 2021-05-21
Prevent panic if DialRsp is returned multiple times

Jefftree:
Manually verified by forcing the proxy server to send two DIAL_RSP to ensure that the panic is resolved.

We shouldn't ever reach this case unless multiple DIAL_REQ are sent with the same randomID

### DIFF
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -93,6 +93,11 @@ func CreateSingleUseGrpcTunnel(address string, opts ...grpc.DialOption) (Tunnel,
 
 func (t *grpcTunnel) serve(c clientConn) {
 	defer c.Close()
+	defer func() {
+		if err := recover(); err != nil {
+			klog.V(1).InfoS("error in konnectivity serve", "err", err)
+		}
+	}()
 
 	for {
 		pkt, err := t.stream.Recv()


konnectivity-client/pkg/client/client.go
```diff
@@ -116,14 +117,15 @@ func (t *grpcTunnel) serve(c clientConn) {
...
-				ch <- dialResult{
-					err:    resp.Error,
-					connid: resp.ConnectID,
-				}
 				// We should only get DialResp once. If we get the DialResp
-				// twice, we will get an exception trying to write to a closed
-				// channel.
-				close(ch)
+				// multiple times, the subsequent responses will be ignored
+				t.handleDialOnce.Do(func() {
```
caesarxuchao, line 122: We want the code to fail to expose if we ever got two DIAL_RSP.

Can you check if a goroutine panics, does that interrupt the execution of the main code?

If not, then we don't need to do anything.

If it does, then we can call recover() in server() to handle the exception.

Jefftree: Updated. The recover function is a no-op if no panic is hit, so it shouldn't interrupt execution.

Jefftree:
https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/221 is a better alternative.

/close

PR #220, opened 2021-05-21
Use more efficient timer pattern.

jkh52:
Context: https://github.com/kubernetes/kubernetes/pull/102213


caesarxuchao:
LGTM.

@liggitt is this the pattern you expected?

PR #221, opened 2021-05-22
Improve handling of repeated dial response.

cheftako:
Make sure we do not block on repeated writes to pending channel
based on repeated dial response packets for one dial request.
Also make sure we do not throw exception writing to closed channel.


caesarxuchao:
This fix is nice!

/lgtm

PR #223, opened 2021-05-25
Exposing metrics for konnectivity network proxy server /ready check

yan-lgtm:


pkg/server/metrics/metrics.go
```diff
@@ -95,3 +113,13 @@ func (a *ServerMetrics) ConnectionInc(service_method string) {
 func (a *ServerMetrics) ConnectionDec(service_method string) {
 	a.connections.With(prometheus.Labels{"service_method": service_method}).Dec()
 }
+
+// SuccessInc increments a successful server health check.
+func (a *ServerMetrics) SuccessInc() {
+	a.healthSuccess.Inc()
```
cheftako, line 119: We seem to be confusing health and readiness. The code incrementing these is the readiness check and not the health check. (Granted the health check currently always returns healthy if it returns anything). However it is important we not confuse these endpoints as they are not the same thing.

yan-lgtm: Get it. Looks like health check was not implemented yet.

cmd/server/main.go
```diff
@@ -663,8 +664,10 @@ func (p *Proxy) runHealthServer(o *ProxyRunOptions, server *server.ProxyServer)
 		if ready {
 			w.WriteHeader(200)
 			fmt.Fprintf(w, "ok")
+			metrics.Metrics.SuccessInc()
```
caesarxuchao, line 667: This metric is incremented when someone sends a GET request to the the `/ready` path. Can you elaborate on the use case of this metric?

pkg/server/metrics/metrics.go
cheftako: The health check is implemented with 
```livenessHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		fmt.Fprintf(w, "ok")
	})```
Its pretty simple. We (currently) assume if the server is responding its "healthy". At some point this may be enhanced, if for example we are able to determine something like the GRPC server is wedged.

cmd/server/main.go
yan-lgtm: I probably put the metrics in the wrong place. What I want is to have a metrics showing the healthiness of konnectivity network server. I checked the liveness check, and it only returns ok without checking if the agent is running.

pkg/server/metrics/metrics.go
yan-lgtm: Yes I get your point here. I was hoping i can reuse some of the periodic health check to give me signal on whether the agent is healthy, looks like i misunderstand the scenarios here. We can discuss this on our Thursday sync meeting. Like I replied to Chao below. What I want is to have a way to expose a metrics whether the server is usable (up and be able to function). Look like this is not the right approach.

yan-lgtm:
Close this PR as not the right implementation per conversation with Walter and Chao

PR #224, opened 2021-05-27
Added histogram metric for send to frontend.

cheftako:
Extended range of standard metric down to nanoseconds.
Added histogram for write to frontend (KAS).
Improved Makefile to reflect server metrics dependency.
```
$ curl http://localhost:8095/metrics
...
# HELP konnectivity_network_proxy_server_frontend_write_duration_seconds Latency of write to the frontend in seconds
# TYPE konnectivity_network_proxy_server_frontend_write_duration_seconds histogram
konnectivity_network_proxy_server_frontend_write_duration_seconds_bucket{le="1e-07"} 0
konnectivity_network_proxy_server_frontend_write_duration_seconds_bucket{le="1e-06"} 3
konnectivity_network_proxy_server_frontend_write_duration_seconds_bucket{le="1e-05"} 3
konnectivity_network_proxy_server_frontend_write_duration_seconds_bucket{le="0.0001"} 3
konnectivity_network_proxy_server_frontend_write_duration_seconds_bucket{le="0.005"} 3
konnectivity_network_proxy_server_frontend_write_duration_seconds_bucket{le="0.025"} 3
konnectivity_network_proxy_server_frontend_write_duration_seconds_bucket{le="0.1"} 3
konnectivity_network_proxy_server_frontend_write_duration_seconds_bucket{le="0.5"} 3
konnectivity_network_proxy_server_frontend_write_duration_seconds_bucket{le="2.5"} 3
konnectivity_network_proxy_server_frontend_write_duration_seconds_bucket{le="12.5"} 3
konnectivity_network_proxy_server_frontend_write_duration_seconds_bucket{le="+Inf"} 3
konnectivity_network_proxy_server_frontend_write_duration_seconds_sum 3.9399999999999995e-07
konnectivity_network_proxy_server_frontend_write_duration_seconds_count 3
```

### DIFF
--- a/pkg/server/metrics/metrics.go
+++ b/pkg/server/metrics/metrics.go
@@ -45,7 +45,7 @@ var (
 type ServerMetrics struct {
 	latencies   *prometheus.HistogramVec
 	connections *prometheus.GaugeVec
-	backend *prometheus.GaugeVec
+	backend     *prometheus.GaugeVec
 }
 
 // newServerMetrics create a new ServerMetrics, configured with default metric names.
@@ -80,14 +80,14 @@ func newServerMetrics() *ServerMetrics {
 		},
 		[]string{},
 	)
-	
+
 	prometheus.MustRegister(latencies)
 	prometheus.MustRegister(connections)
 	prometheus.MustRegister(backend)
 	return &ServerMetrics{
-		latencies: latencies,
+		latencies:   latencies,
 		connections: connections,
-		backend: backend,
+		backend:     backend,
 	}
 }
 
--- a/Makefile
+++ b/Makefile
@@ -138,25 +138,25 @@ cfssljson:
 certs: export PATH := $(shell pwd):$(PATH)
 certs: easy-rsa-master cfssl cfssljson
 	# set up easy-rsa
-	cp -rf easy-rsa-master/easyrsa3 easy-rsa-master/master
+	cp -rf easy-rsa-master/easyrsa3 easy-rsa-master/frontend
 	cp -rf easy-rsa-master/easyrsa3 easy-rsa-master/agent
 	# create the client <-> server-proxy connection certs
-	cd easy-rsa-master/master; \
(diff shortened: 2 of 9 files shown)


pkg/server/metrics/metrics.go
```diff
@@ -19,6 +19,7 @@ package metrics
 import (
 	"time"
 
+	"k8s.io/klog/v2"
```
Jefftree, line 22: I think this is an unused import?

cheftako: Yup :D I was initially thrown a little about the way prometheus histogram reporting. I added some logging to understand what it was doing. I removed that logging but not the import it added.

pkg/server/metrics/metrics.go
```diff
@@ -34,8 +34,8 @@ const (
 )
 
 var (
-	// Use buckets ranging from 5 ms to 12.5 seconds.
-	latencyBuckets = []float64{0.005, 0.025, 0.1, 0.5, 2.5, 12.5}
+	// Use buckets ranging from 10 ns to 12.5 seconds.
```
caesarxuchao, line 37: nit: 10e-7 is 100 ns

10 ns sounds too small, that's just tens of CPU cycles, even for a local write I doubt 10 ns is enough, so let's just update the comment to 100ns.

caesarxuchao (commented):
One nit. Otherwise lgtm.

pkg/server/metrics/metrics.go
cheftako: Fair. I kept masking it smaller until the units actually didn't show up to confirm ```le``` was working.

cheftako:
Need to fix this as i crossed wires with a different PR.

cheftako:
Change has been lost. Closing.

cheftako:
Resused as https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/234

PR #227, opened 2021-06-01
refine cmd/agent/main.go and cmd/server/main.go

liufen90:
refine cmd/agent/main.go and cmd/server/main.go


cmd/agent/app/options/options.go
```diff
@@ -0,0 +1,186 @@
...
+		ProbeInterval:           o.ProbeInterval,
+		SyncIntervalCap:         o.SyncIntervalCap,
+		DialOptions:             dialOptions,
+		ServiceAccountTokenPath: o.ServiceAccountTokenPath,
+	}
+}
+
+func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
+	flags := pflag.NewFlagSet("proxy-agent", pflag.ContinueOnError)
+	flags.StringVar(&o.AgentCert, "agent-cert", o.AgentCert, "If non-empty secure communication with this cert.")
```
cheftako, line 63: I realize this is largely a forklift but slightly nervous about making sure all the flags are maintained (i.e. binaries are backward compatible). Any chance of tests for the Flags functions?

liufen90: I checked that I didn't modify the flags' names. And I tested the flags locally.

liufen90:
@cheftako  please have a look

cmd/server/app/server.go
```diff
@@ -0,0 +1,372 @@
...
+			fmt.Fprintf(w, "ok")
+			return
+		}
+		w.WriteHeader(500)
+		fmt.Fprintf(w, msg)
+	})
+
+	muxHandler := http.NewServeMux()
+	muxHandler.HandleFunc("/healthz", livenessHandler)
+	muxHandler.HandleFunc("/ready", readinessHandler)
```
cheftako, line 355: I realize this was a pre-existing bug but this should be ```/readyz```.

cmd/agent/app/server.go
```diff
@@ -0,0 +1,137 @@
...
+		fmt.Fprintf(w, "ok")
+	})
+	readinessHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintf(w, "ok")
+	})
+
+	muxHandler := http.NewServeMux()
+	muxHandler.Handle("/metrics", promhttp.Handler())
+	muxHandler.HandleFunc("/healthz", livenessHandler)
+	muxHandler.HandleFunc("/ready", readinessHandler)
```
cheftako, line 85: I realize this was a pre-existing bug but this should be ```/readyz```.

cmd/server/app/server.go
liufen90: I found that it is /ready in master branch. Maybe we can fix it in other PRs. This PR only refine main package.

cmd/agent/app/server.go
liufen90: I found that it is /ready in master branch. Maybe we can fix it in other PRs. This PR only refine main package.

PR #228, opened 2021-06-01
add support for default backend specification in destHost

relyt0925:
This allows an agent to specify using a enumerated host (open to discussion on the best way/value for this) that if no direct match based on IP or Host header is found to send traffic through agents with this specification. 

It enables a user to deploy a pool of agents in a network that they expect to handle *all* traffic besides a couple exceptions (in our case: the couple exceptions are Aggregated API Servers that are in a separate isolated network). The agent in that network will specifically call out the hosts it serves and only process those hosts.


charleszheng44:
@relyt0925 Could you elaborate on how could this change help to "handle all traffic beside a couple of exceptions"? It looks to me that if the `default-backend.konnectivity.io` is set, then all requests whose destination IP/host is not in the `backends` map will be sent to the default backend. If this is the case, IMO, it should be better to put this feature in the `DefaultBackendManager`.

relyt0925:
@charleszheng44 

The main gap with just the default random backend approach is the following:

Picture a scenario where your master control plane is in a isolated network from your intra cluster control plane (no direct IP connectivity). Then in this cluster picture you have APIServices (Aggregated API Servers) that are also running in the isolated master control plane network. In this scenario: the base kube apiserver needs to reach out to the aggregated apiservers in the isolated network. 

The issue is that APIServices in Kubernetes *always* have to be exposed as a in cluster service. This means that this line of the doc is wrong
```
https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1281-network-proxy#kubernetes-api-server-outbound-requests
Aggregated API Server (and OpenAPI requests for aggregated resources) Aggregated API Servers can be destined for a service or a URL. If destined for a service then the service rules apply. If destined for a URL then we will use the ‘master’ NetworkContext.
```
There is no possibility in the Kube API Spec of an APIService to expose it as a URL (shown here: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#servicereference-v1-apiregistration-k8s-io). Therefore: currently all APIService traffic always is classified as the "cluster" category of traffic. Note this is the main category where there is a gap on the classification of traffic. 

This can be solved by running two unique pools of agents: one in the isolated master control plane network and one in the intra cluster network (what a user typically sees as the Kubernetes cluster). The agent that connects to the server in the master control plane network *only* advertises the APIServices it needs to process. 

Where this gets tricky with the current approach is the default "pick a random backend if there is not an exact match behavior". With "pick a random backend" let's assume there's an equal amount of master control plane agents and intra cluster agents. There's now a 50% chance the random backend choosen will not be able to properly serve the request since it's in the wrong network. 

In the current approach in order to get this fully right: you will need each agent to advertise effectively every ip and every host it will ever be responsible in serving. That won't be feasible for the intra cluster agents since they will end up serving a vast amount of IPs/ services/ node IPs/ etc that are changing over the life of the cluster. However: the APIServices on the control plane side remain static over the life of the cluster (can change when the control plane goes through an update but is infrequent and at a managable scale). 

This is where the solution I posted is helpful: it allows us to say the intra cluster traffic agents should service all requests sent through connectivity unless a specific subset of IPs/Hostnames are contacted. In that case we know it needs to go to the Konnectivity agent running in the master network in order to properly route the request.

relyt0925:
Basically when you have isolated networks: in the current approach the only option for using destHost is to specify all hostnames/ips ever in use. 

With this approach: you can say "all traffic should go to this agent pool besides the ones I specifically call out to go to the agent pool I specify process this specific hostname/ip". It makes the config management of this setup feasible as it's not feasible to outline all the possible endpoints that the intra cluster agent (typical konnectivity agent that runs as a daemonset in the user kubernetes cluster) processes as it's a large dynamic range of traffic as services/nodes/pods/etc are created

Whereas the master control plane network is feasible to keep up to date overtime

relyt0925:
For those that are interested on how that typically looks what you can do to advertise that IP address is you can do the following
```
https://github.com/openshift/ibm-roks-toolkit/blob/62dba55df9e7fb0b0f5a352a653d7268d7558c56/assets/openshift-apiserver/service-template.yaml#L11
https://github.com/openshift/ibm-roks-toolkit/blob/62dba55df9e7fb0b0f5a352a653d7268d7558c56/assets/openshift-apiserver/openshift-apiserver-user-service.yaml
https://github.com/openshift/ibm-roks-toolkit/blob/62dba55df9e7fb0b0f5a352a653d7268d7558c56/assets/openshift-apiserver/openshift-apiserver-user-endpoint.yaml#L8
```

Basically you can create a headless service and create an endpoint that contains the IP in the master control plane network the kube-apiserver needs to contact. What will happen is the kube-apiserver will resolve this service to that ip and then reach out through konnectivity to connect to it. As the operator of the cluster: I will be able to configure konnectivity for these specific services that I know are in this isolated network to route through the agent in that network by using the agent-identifiers support as is today.

Then I know for the rest of all the other traffic I have going through konnectivity: it can be sent through the intra cluster agent pool that is going to explicity be configured to advertise this "default route enum".

relyt0925:
@charleszheng44

Hopefully that helps..... let me know if you want more details. The main fundamental idea is when master control plane components are in an isolated network from the intra cluster components the random picking a backend leads to requests failing if they go to a proxy in the wrong network. To prevent that as is today: every host/ip needs to be specified that an agent processes. This is very dynamic in the case of the intra cluster components since users are creating pods, services, nodes getting removed/added, etc. Therefore: the default route allows it to by default get selected for all traffic *besides* a subset of components (APIServices specifically that we know will run in the isolated control plane network).

charleszheng44:
@relyt0925 thanks for the detailed explanation! 

Sorry if I didn't explain it well. As @cheftako  mentioned, the BackendManager is pluggable and a running server can have multiple active BackendManagers working together. 

What I suggested was can we implement this feature in the [`Backend()`](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/6f3de56514e8b56b99da7d537026245b314d259b/pkg/server/backend_manager.go#L124) function of  the `DefaultBackendManager` (or implement a new `BackendManager`? ), and when starting the `konnectivity server`, we can specify two proxy strategies, i.e. `--proxy-strategies destHost,default`, then the server will go through both backend managers (in the order of `DestHostBackendManager` and `DefaultBackendManager`, for details, please refer to the  [`getBackend()`](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/6f3de56514e8b56b99da7d537026245b314d259b/pkg/server/server.go#L168)), and the forwarding process will look like the followings,

1. the `DestHostBackendManager` will check if it can find an agent with the same IP/Host as the destination IP/Host, if yes, forward the request to the agent.
2. if not, the `DefaultBackendManager` will check if the `default-backend.konnectivity.io` is set, if yes, the request will be sent to the agent associated with the `default-backend.konnectivity.io`.
3. else, a random agent will be chosen. 

The reason I suggest this is because the `DestHostBackendManager` is supposed to pick up an agent that has the IP/Host same as the request's destination, if we try to set a default agent, it may be confusing to future users.

relyt0925:
That works I can make that change!

relyt0925:
@charleszheng44 I adjusted to integrate this system into the DefaultBackend. There are no breaking changes to existing functionality but it adds support for agents to explicitly claim the default route. If that is done traffic is sent through a randomly selected agent out of that pool. If there is not an active member of that pool: the regular default strategy of selecting any active agent is used.

pkg/server/backend_manager.go
```diff
@@ -232,6 +239,15 @@ func (s *DefaultBackendStorage) RemoveBackend(identifier string, idType pkgagent
 				break
 			}
 		}
+		if idType == pkgagent.DefaultRoute {
+			for i := range s.defaultRouteAgentIDs {
+				if s.defaultRouteAgentIDs[i] == identifier {
+					s.defaultRouteAgentIDs[i] = s.defaultRouteAgentIDs[len(s.defaultRouteAgentIDs)-1]
+					s.defaultRouteAgentIDs = s.defaultRouteAgentIDs[:len(s.defaultRouteAgentIDs)-1]
+					break
```
charleszheng44, line 247: nit: 
```go
s.defaultRouteAgentIDs = append(s.defaultRouteAgentIDs[:i], s.defaultRouteAgentIDs[i+1:]...)
```

charleszheng44 (approved):
Thanks @relyt0925!  The PR /lgtm, I will leave it to @cheftako for another pass.

relyt0925:
Awesome! I'll update the unit tests while that occurs as well

relyt0925:
Unit tests added

relyt0925:
@charleszheng44 just FYI:

I had to add a new strategy in order to ensure in the total failure scenario of agents in the intra cluster network doesn't lead to traffic accidentally flowing to the master cluster network and wasting bandwidth. In this situation: It's better to have the requests fail and if we enhance the default strategy it will just fall back to selecting a random agent in the isolated master network.

The PR has been adjusted to create a new strategy `defaultRoute` that has the desired behavior of if an agent is available that has explicitly advertised the default route: send traffic to it. If not, then fail. It can still be chained with the basic default strategy to get that desired behavior if someone wants that but in our situation that is not ideal.

mihivagyok:
@relyt0925
After the some small fixes, the initial test looks good!
- startup of the konnectivity server
```
I0603 12:09:14.121398       1 main.go:169] ServerCert set to "/etc/kubernetes/konn-certs/konnectivity-server.pem".
I0603 12:09:14.121459       1 main.go:170] ServerKey set to "/etc/kubernetes/konn-certs/konnectivity-server-key.pem".
I0603 12:09:14.121467       1 main.go:171] ServerCACert set to "/etc/kubernetes/konn-certs/ca.crt".
I0603 12:09:14.121473       1 main.go:172] ClusterCert set to "/etc/kubernetes/server.crt".
I0603 12:09:14.121479       1 main.go:173] ClusterKey set to "/etc/kubernetes/server.key".
I0603 12:09:14.121485       1 main.go:174] ClusterCACert set to "".
I0603 12:09:14.121523       1 main.go:175] Mode set to "http-connect".
I0603 12:09:14.121529       1 main.go:176] UDSName set to "".
I0603 12:09:14.121534       1 main.go:177] DeleteUDSFile set to false.
I0603 12:09:14.121544       1 main.go:178] Server port set to 8090.
I0603 12:09:14.121551       1 main.go:179] Agent port set to 8091.
I0603 12:09:14.121557       1 main.go:180] Admin port set to 8095.
I0603 12:09:14.121562       1 main.go:181] Health port set to 8092.
I0603 12:09:14.121588       1 main.go:182] Keepalive time set to 1h0m0s.
I0603 12:09:14.121598       1 main.go:183] EnableProfiling set to false.
I0603 12:09:14.121648       1 main.go:184] EnableContentionProfiling set to false.
I0603 12:09:14.121660       1 main.go:185] ServerID set to af54a6a8-d091-473f-a7fe-2aff8e20e6e2.
I0603 12:09:14.121680       1 main.go:186] ServerCount set to 1.
I0603 12:09:14.121688       1 main.go:187] AgentNamespace set to "".
I0603 12:09:14.121694       1 main.go:188] AgentServiceAccount set to "".
I0603 12:09:14.121701       1 main.go:189] AuthenticationAudience set to "".
I0603 12:09:14.121726       1 main.go:190] KubeconfigPath set to "".
I0603 12:09:14.121785       1 main.go:191] KubeconfigQPS set to 0.000000.
I0603 12:09:14.121826       1 main.go:192] KubeconfigBurst set to 0.
I0603 12:09:14.121848       1 main.go:193] ProxyStrategies set to "destHost,defaultRoute".
I0603 12:09:14.121961       1 main.go:410] Starting master server for client connections.
I0603 12:09:14.123731       1 main.go:422] Starting agent server for tunnel connections.
I0603 12:09:14.141553       1 main.go:427] Starting admin server for debug connections.
I0603 12:09:14.141720       1 main.go:432] Starting health server for healthchecks.
I0603 12:09:18.798507       1 server.go:605] "Connect request from agent" agentID="cfcf2505-107a-477e-873e-e34d33113f75"
I0603 12:09:18.798608       1 server.go:192] "Add the agent to DestHostBackendManager" agent address="10.176.3.134"
I0603 12:09:18.798671       1 backend_manager.go:190] "Register backend for agent" connection=&{ServerStream:0xc0004540c0} agentID="10.176.3.134"
I0603 12:09:18.798710       1 server.go:205] "Add the agent to DefaultRouteBackendManager" agentID="cfcf2505-107a-477e-873e-e34d33113f75"
```

Then agent selection:
```
I0603 12:10:14.702905       1 default_route_backend_manager.go:49] "Picked agent as backend" agentID="cfcf2505-107a-477e-873e-e34d33113f75"
I0603 12:10:14.721258       1 server.go:678] "Received DIAL_RSP" random=4037200794235010051 agentID="cfcf2505-107a-477e-873e-e34d33113f75" connectionID=4
I0603 12:10:14.721389       1 server.go:250] "Register frontend for agent" frontend=&{Mode:http-connect Grpc:<nil> HTTP:0xc000061500 connected:0xc0002c6600 connectID:4 agentID:cfcf2505-107a-477e-873e-e34d33113f75 start:{wall:13845841557076682353 ext:60654283810 loc:0x23958a0} backend:0xc0001a6ac0} agentID="cfcf2505-107a-477e-873e-e34d33113f75" connectionID=4
I0603 12:10:14.721433       1 tunnel.go:121] "Starting proxy to host" host="172.31.7.11:80"
I0603 12:10:14.721493       1 tunnel.go:154] "Forwarding data on tunnel to agent" bytes=102 totalBytes=102 agentID="cfcf2505-107a-477e-873e-e34d33113f75" connectionID=4
I0603 12:10:14.741062       1 server.go:707] "Received data from agent" bytes=136 agentID="cfcf2505-107a-477e-873e-e34d33113f75" connectionID=4
I0603 12:10:14.741129       1 server.go:716] "DATA sent to frontend"
I0603 12:10:14.764959       1 tunnel.go:132] "EOF from host" host="172.31.7.11:80"
I0603 12:10:14.765026       1 tunnel.go:161] "Stopping transfer to host" host="172.31.7.11:80" agentID="cfcf2505-107a-477e-873e-e34d33113f75" connectionID=4
I0603 12:10:14.769045       1 server.go:721] "Received CLOSE_RSP" connectionID=4
E0603 12:10:14.769072       1 server.go:729] "CLOSE_RSP send to client stream error" err="tls: use of closed connection" connectionID=4
I0603 12:10:14.769103       1 server.go:271] "Remove frontend for agent" frontend=&{Mode:http-connect Grpc:<nil> HTTP:0xc000061500 connected:0xc0002c6600 connectID:4 agentID:cfcf2505-107a-477e-873e-e34d33113f75 start:{wall:13845841557076682353 ext:60654283810 loc:0x23958a0} backend:0xc0001a6ac0} agentID="cfcf2505-107a-477e-873e-e34d33113f75" connectionID=4
I0603 12:10:14.769138       1 server.go:734] "Close streaming" agentID="cfcf2505-107a-477e-873e-e34d33113f75" connectionID=4
```

PR with the fixes:
https://github.com/relyt0925/apiserver-network-proxy/pull/1

Thank you!

relyt0925:
This is ready for merge

relyt0925:
cc @cheftako

charleszheng44:
@relyt0925 Thanks for adding the test cases! I like the idea of adding the `DefaultRouteBackendManager`, it makes things more clear 😄

relyt0925:
Thank you for your time, quick review, and advice @charleszheng44 ! You are awesome to work with :)

pkg/agent/client.go
```diff
@@ -128,6 +130,8 @@ func GenAgentIdentifiers(addrs string) (Identifiers, error) {
 			agentIDs.Host = append(agentIDs.Host, ids...)
 		case CIDR:
 			agentIDs.CIDR = append(agentIDs.CIDR, ids...)
+		case DefaultRoute:
+			agentIDs.DefaultRoute = true
```
cheftako, line 134: We shouldn't assume that ids == "true". If someone tries to explicitly set "DefaultRoute=false" we should do the right thing. That can either be return an error in the validation or set it correctly here.

pkg/server/server.go
```diff
@@ -181,13 +181,13 @@ func (s *ProxyServer) getBackend(reqHost string) (Backend, error) {
 
 func (s *ProxyServer) addBackend(agentID string, conn agent.AgentService_ConnectServer) (backend Backend) {
 	for i := 0; i < len(s.BackendManagers); i++ {
+		agentIdentifiers, err := getAgentIdentifiers(conn)
```
cheftako, line 184: The default case is "random" across all backends without concern to agent identifiers. As such it should not care if there are any agent identifiers. So I don't think we can early out here. I think we do have to duplicate this check in both DestHostBackendManager and DefaultRouteBackendManager.

pkg/server/server.go
```diff
@@ -210,13 +215,13 @@ func (s *ProxyServer) addBackend(agentID string, conn agent.AgentService_Connect
 
 func (s *ProxyServer) removeBackend(agentID string, conn agent.AgentService_ConnectServer) {
 	for _, bm := range s.BackendManagers {
+		agentIdentifiers, err := getAgentIdentifiers(conn)
```
cheftako, line 218: Ditto

cheftako:
Two minor nits, otherwise looks good.

pkg/server/server.go
relyt0925: should be addressed!

pkg/server/server.go
relyt0925: should be addressed!

pkg/agent/client.go
relyt0925: should be addressed!

relyt0925:
@cheftako thank you for your time and insightful comments. They should be addressed and this should be ready for another review.

cheftako:
/approve
Approving hoping one more set of eyes will check and lgtm.

relyt0925:
cc @cheftako  

@charleszheng44 do you mind giving the lgtm?

PR #231, opened 2021-06-06
Fixed readiness url path to be /readyz

cheftako:
K8s specifies /readyz as the proper url path for the readiness check.
Added that path to the supported url pathes for the readiness check.
Keeping /ready for now for bacwkard compatability.
We are treating it as deprecated.

### DIFF
--- a/cmd/agent/app/server.go
+++ b/cmd/agent/app/server.go
@@ -82,7 +82,9 @@ func (a *Agent) runHealthServer(o *options.GrpcProxyAgentOptions) error {
 	muxHandler := http.NewServeMux()
 	muxHandler.Handle("/metrics", promhttp.Handler())
 	muxHandler.HandleFunc("/healthz", livenessHandler)
+	// "/ready" is deprecated but being maintained for backward compatibility
 	muxHandler.HandleFunc("/ready", readinessHandler)
+	muxHandler.HandleFunc("/readyz", readinessHandler)
 	healthServer := &http.Server{
 		Addr:           fmt.Sprintf(":%d", o.HealthServerPort),
 		Handler:        muxHandler,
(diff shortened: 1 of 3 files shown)


cmd/agent/app/server.go
```diff
@@ -83,6 +83,9 @@ func (a *Agent) runHealthServer(o *options.GrpcProxyAgentOptions) error {
 	muxHandler.Handle("/metrics", promhttp.Handler())
 	muxHandler.HandleFunc("/healthz", livenessHandler)
 	muxHandler.HandleFunc("/ready", readinessHandler)
+	// "/ready" is deprecated but being maintained for backward compatibility
+	muxHandler.HandleFunc("/ready", readinessHandler)
```
Jefftree, line 87: Duplicate of line 84?

cmd/server/app/server.go
```diff
@@ -353,6 +353,9 @@ func (p *Proxy) runHealthServer(o *options.ProxyRunOptions, server *server.Proxy
 	muxHandler := http.NewServeMux()
 	muxHandler.HandleFunc("/healthz", livenessHandler)
 	muxHandler.HandleFunc("/ready", readinessHandler)
+	// "/ready" is deprecated but being maintained for backward compatibility
+	muxHandler.HandleFunc("/ready", readinessHandler)
```
Jefftree, line 357: same comment as above

cmd/agent/app/server.go
cheftako: Yeah, I think this is from rebase. Fixed now.

cmd/server/app/server.go
cheftako: Ditto

Issue #232, opened 2021-06-07
Remove the term "master" from the code base


cheftako:
"master" is neither inclusive nor considerate of our members. As per https://github.com/kubernetes/community/blob/master/wg-naming/language-evaluation-framework.md#is-the-term-overtly-racist it is against K8s/CNCF policy. We need to clean up the code base to remove this term.

PR #233, opened 2021-06-08
First pass at fixing non-inclusive verbage.

cheftako:
Switched easy fix uses of master to either control plane or fronted as
appropriate. However we still have references to the git branch and also
to a feature gate. The latter requires a backward compatible method to
fix.

Issue: #232


Makefile
```diff
@@ -138,25 +138,25 @@ cfssljson:
 certs: export PATH := $(shell pwd):$(PATH)
 certs: easy-rsa-master cfssl cfssljson
```
Jefftree, line 139: We should be able to rename `easy-rsa-master` as well right? It seems like the package is just called `easy-rsa` (line 121)

Jefftree (commented):
One comment, otherwise LGTM

Makefile
cheftako: I can give it a go.

cheftako:
/remove hold
Removed easy-rsa-master

PR #235, opened 2021-06-10
Fix the wrong command in readme

SataQiu:
Fix the wrong command in readme


Jefftree:
LGTM after rebase conflict is fixed.

Issue #222, opened 2021-05-24
serveRecvBackend should not block on sending to frontend.


caesarxuchao:
This is derived from an offline discussion with @cheftako and @jkh52.

For each agent, the proxy server runs a [serveRecvBackend() goroutine](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/603aa670ea3d1e8449dfc7ddd9a303a07a53b904/pkg/server/server.go#L611) to serve the traffic from the proxy agent to the proxy server.

When debugging #167, we found that the serveRecvBackend can [block](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/603aa670ea3d1e8449dfc7ddd9a303a07a53b904/pkg/server/server.go#L675) on sending to a frontend if that frontend did not consume the data in the grpc connection and the amount of unread data reached the threshold set by the grpc flowcontrol. This blocked __all__ traffic that was proxied by the proxy agent.

We need to improve the isolation among frontends. A few thoughts we had:
* We can add a timeout on the send. GRPC does not provide a timeout interface, so we have to do this in the form of a select block.
* If sending to a frontend ever times out, proxy server marks this frontend as unhealthy, so that it can skip sending future packets to it to avoid more blocking.
* Unfortunately we can't terminate the GRPC connection from the server side, and in this case, the proxy server is the GRPC server. Thus, the proxy server can only passively wait for the frontend to terminate the GRPC connection. This can lead to memory leaks.
  * What the proxy server can do is that it can mark itself as unhealthy in its `/healthz`, so that kubelet will restart the proxy server, and that resets all broken GRPC connections.
* We can add histogram metrics for the latency of the send as the first step.

cheftako:
> marks this frontend as unhealthy
To be clear we mentioned marking the frontend stream an unhealthy. (As opposed to marking an entire KAS as unhealthy)

ydp:
Hi @caesarxuchao , thanks again for your work, if the purpose of marking proxy-server as unhealthy is to let kubelet restart it, could it be possible to just let it exit and make supervisor to restart it? Per my understanding, the ANP is meant to isolate control plane and cluster network, one way to achieve that is to remove kubelet from control plane, in this case, marking proxy-server unhealthy is not enough to make it restart.

caesarxuchao:
https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/234 adds the histogram metrics.

[...]

PR #238, opened 2021-06-17
Add context to CreateSingleUseGrpcTunnel.

jkh52:
This should make it possible for apiserver (egress selector) to
propagate cancel to proxy-server.

### DIFF
--- a/cmd/client/main.go
+++ b/cmd/client/main.go
@@ -337,9 +337,10 @@ func (c *Client) getDialer(o *GrpcProxyClientOptions) (func(ctx context.Context,
 }
 
 func (c *Client) getUDSDialer(o *GrpcProxyClientOptions) (func(ctx context.Context, network, addr string) (net.Conn, error), error) {
+	ctx := context.Background()
+
 	var proxyConn net.Conn
 	var err error
-
 	// Setup signal handler
 	ch := make(chan os.Signal, 1)
 	signal.Notify(ch)
@@ -364,13 +365,13 @@ func (c *Client) getUDSDialer(o *GrpcProxyClientOptions) (func(ctx context.Conte
 			}
 			return c, err
 		})
-		tunnel, err := client.CreateSingleUseGrpcTunnel(o.proxyUdsName, dialOption, grpc.WithInsecure(), grpc.WithUserAgent(o.userAgent))
+		tunnel, err := client.CreateSingleUseGrpcTunnel(ctx, o.proxyUdsName, dialOption, grpc.WithInsecure(), grpc.WithUserAgent(o.userAgent))
 		if err != nil {
 			return nil, fmt.Errorf("failed to create tunnel %s, got %v", o.proxyUdsName, err)
 		}
 
 		requestAddress := fmt.Sprintf("%s:%d", o.requestHost, o.requestPort)
-		proxyConn, err = tunnel.Dial("tcp", requestAddress)
+		proxyConn, err = tunnel.DialContext(ctx, "tcp", requestAddress)
 		if err != nil {
 			return nil, fmt.Errorf("failed to dial request %s, got %v", requestAddress, err)
 		}
@@ -410,6 +411,7 @@ func (c *Client) getUDSDialer(o *GrpcProxyClientOptions) (func(ctx context.Conte
 }
 
 func (c *Client) getMTLSDialer(o *GrpcProxyClientOptions) (func(ctx context.Context, network, addr string) (net.Conn, error), error) {
+	ctx := context.Background()
 	tlsConfig, err := util.GetClientTLSConfig(o.caCert, o.clientCert, o.clientKey, o.proxyHost, nil)
 	if err != nil {
 		return nil, err
@@ -434,13 +436,13 @@ func (c *Client) getMTLSDialer(o *GrpcProxyClientOptions) (func(ctx context.Cont
 		transportCreds := credentials.NewTLS(tlsConfig)
 		dialOption := grpc.WithTransportCredentials(transportCreds)
 		serverAddress := fmt.Sprintf("%s:%d", o.proxyHost, o.proxyPort)
-		tunnel, err := client.CreateSingleUseGrpcTunnel(serverAddress, dialOption)
+		tunnel, err := client.CreateSingleUseGrpcTunnel(ctx, serverAddress, dialOption)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create tunnel %s, got %v", serverAddress, err)
 		}
 
 		requestAddress := fmt.Sprintf("%s:%d", o.requestHost, o.requestPort)
-		proxyConn, err = tunnel.Dial("tcp", requestAddress)
+		proxyConn, err = tunnel.DialContext(ctx, "tcp", requestAddress)
 		if err != nil {
 			return nil, fmt.Errorf("failed to dial request %s, got %v", requestAddress, err)
 		}
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -35,7 +35,7 @@ import (
 type Tunnel interface {
 	// Dial connects to the address on the named network, similar to
 	// what net.Dial does. The only supported protocol is tcp.
-	Dial(protocol, address string) (net.Conn, error)
+	DialContext(ctx context.Context, protocol, address string) (net.Conn, error)
 }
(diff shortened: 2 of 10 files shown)


jkh52:
This is the first half of a suggestion at https://github.com/kubernetes/kubernetes/issues/102904. This should help further mitigate https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/222.

konnectivity-client/pkg/client/client.go
```diff
@@ -66,15 +66,15 @@ var _ clientConn = &grpc.ClientConn{}
 // gRPC based proxy service.
 // Currently, a single tunnel supports a single connection, and the tunnel is closed when the connection is terminated
 // The Dial() method of the returned tunnel should only be called once
-func CreateSingleUseGrpcTunnel(address string, opts ...grpc.DialOption) (Tunnel, error) {
-	c, err := grpc.Dial(address, opts...)
+func CreateSingleUseGrpcTunnel(ctx context.Context, address string, opts ...grpc.DialOption) (Tunnel, error) {
```
cheftako, line 69: We should add documentation/warning about the context on the tunnel. We should *NOT* be using request level context if there is any chance that the tunnel will be (re)used for more than 1 request.

konnectivity-client/pkg/client/client.go
```diff
@@ -86,12 +86,12 @@ func CreateSingleUseGrpcTunnel(address string, opts ...grpc.DialOption) (Tunnel,
...
 	}
 
-	go tunnel.serve(c)
+	go tunnel.serve(ctx, c)
 
 	return tunnel, nil
 }
 
-func (t *grpcTunnel) serve(c clientConn) {
+func (t *grpcTunnel) serve(ctx context.Context, c clientConn) {
```
cheftako, line 94: Not sure why we're bothering to add the context here if we're not planning on using it.

konnectivity-client/pkg/client/client.go
```diff
@@ -35,7 +35,7 @@ import (
 type Tunnel interface {
 	// Dial connects to the address on the named network, similar to
 	// what net.Dial does. The only supported protocol is tcp.
-	Dial(protocol, address string) (net.Conn, error)
+	DialContext(ctx context.Context, protocol, address string) (net.Conn, error)
```
Jefftree, line 38: This is a breaking interface change. Would it make sense to keep both interfaces here for backwards compatibility? The golang net dialer has both functions: `net.Dial` and `net.DialContext` 

cc: @cheftako

konnectivity-client/pkg/client/client_test.go
```diff
@@ -70,10 +73,10 @@ func TestData(t *testing.T) {
 		conns:       make(map[int64]*conn),
 	}
 
-	go tunnel.serve(&fakeConn{})
+	go tunnel.serve(ctx, &fakeConn{})
```
cheftako, line 76: Would be nice if we had a test which used a fakeDelayConn so we could add a test for timeout on the context.

tests/proxy_test.go
```diff
@@ -117,14 +117,14 @@ func TestProxyHandleDialError_GRPC(t *testing.T) {
 	time.Sleep(time.Second)
 
 	// run test client
-	tunnel, err := client.CreateSingleUseGrpcTunnel(proxy.front, grpc.WithInsecure())
```
cheftako, line 120: Can we add a context timeout test for SingleUse?

konnectivity-client/pkg/client/client.go
cheftako: This is a breaking change. Technically as a v0 release that would be allowed but certainly not preferred. As this is the piece which is wired into the KAS I'm less concerned about it. If we take this it means that whoever integrates this fix into K/K has to fix all the callers of this client. On the plus side it means that whoever integrates this fix into K/K has to fix ALL the callers of this client. 

I'm much more concerned with ensuring we remain backward compatible with the protocol/transport between the KAS and the Network Proxy-Server. The Network Proxy Server is a reference implementation and we expect that other groups can swap it out. So if we make a backward incompatible change there it isn't something we can fix.

liggitt: the current interface is a hang landmine... if we're still in v0 we should change the signature to require a context to ensure this doesn't hang when the parent request has moved on

konnectivity-client/pkg/client/client.go
liggitt: to be fair, the "SingleUse" name and "The Dial() method of the returned tunnel should only be called once" docs are already pretty blunt

konnectivity-client/pkg/client/client.go
liggitt: if there anywhere inside this method we block that we should also short circuit on `<-ctx.Done()` (like a peer to the `case <-timer.C` case)?

cheftako: If we find one then I'm fine changing the signature but till then I don't see the need.

konnectivity-client/pkg/client/client.go
cheftako: That's fair but I think slightly different than what I'm getting at. It is possible to send (serially) multiple requests over a connection returned by the dial/tunnel. We should not be using a request level context on the create tunnel request as it could cause the tunnel to terminate early.

liggitt: if the tunnel is being created just for one request (is it?) then we want the dial for establishing the tunnel to be bounded to that request's lifetime, right?

cheftako: Yes. However we seem to have a single place from which this is called in K/K (https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/staging/src/k8s.io/apiserver/pkg/server/egressselector/egress_selector.go#L206). I cannot convince myself that the tunnel is only used for a single request. In fact the hung log request/tunnel scenario seemed to suggest that tunnels were some times reused.

konnectivity-client/pkg/client/client.go
jkh52: I'll remove from here. I tried adding a check for <-ctx.Done() at the top of serve() for loop, but via unit test I found it unnecessary (because RecvMsg() will return an error)

konnectivity-client/pkg/client/client.go
jkh52: I also prefer making the client strict since the protocol will remain compatible, and there are extremely few call sites to adapt.

tests/proxy_test.go
jkh52: Done - but one of my two new tests isn't passing and I can't figure out why. (Is it the test itself or a remaining defect?)

konnectivity-client/pkg/client/client.go
jkh52: I also have a hard time tracing the kube-apiserver callsites. (oof: https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/staging/src/k8s.io/apimachinery/pkg/util/net/http.go#L224)

For this PR: In the worst case, we can update K/K with context.TODO()

cmd/client/main.go
```diff
@@ -344,13 +344,13 @@ func (c *Client) getUDSDialer(o *GrpcProxyClientOptions) (func(ctx context.Conte
 			}
 			return c, err
 		})
-		tunnel, err := client.CreateSingleUseGrpcTunnel(o.proxyUdsName, dialOption, grpc.WithInsecure(), grpc.WithUserAgent(o.userAgent))
+		tunnel, err := client.CreateSingleUseGrpcTunnel(context.Background(), o.proxyUdsName, dialOption, grpc.WithInsecure(), grpc.WithUserAgent(o.userAgent))
```
cheftako, line 347: Not worth holding the PR for but it would be a better example if we used 
```ctx := context.Background()```
then passed ctx in on these calls.

tests/proxy_test.go
cheftako: Filed https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/240 to get these tests merged. 
Would like to land these changes in K/K to confirm the fix, so not holding for these tests, yet...

cmd/client/main.go
jkh52: Done.

Issue #236, opened 2021-06-10
Performance of Konnectivity is limited when using proxy requests


mcginne:
In previous Kube versions we were using OpenVPN, but have switched to Konnectivity for Kube v1.21.

I have been running some performance tests to compare the performance of the 2 providers.

When running proxy requests to hit a service running in the cluster I am seeing much lower performance with Konnectivity:

I am using jmeter with 50 threads to hit the following URL:
```
https://<master URL>/api/v1/namespaces/default/services/http:httpperf-np-service:http/proxy
```
`httpperf-np-service` is a simple http server running in the cluster.

I ran tests comparing OpenVPN and then Konnectivity with different CPU limit settings for the `konnectivity-server` pod. The results are here:

<img width="737" alt="image" src="https://user-images.githubusercontent.com/27362732/121527303-2aaa3b00-c9f2-11eb-994f-c6bbd75647ae.png">

We can see OpenVPN has significantly higher throughput.

On looking at the Konnectivity logs it looks like it is doing a lot of opening and closing of connections to the nodes:
```
0609 16:00:18.218892       1 tunnel.go:121] "Starting proxy to host" host="172.30.80.4:8080"
...
I0609 16:00:18.229134       1 server.go:708] "Received CLOSE_RSP" connectionID=10280
E0609 16:00:18.229169       1 server.go:716] "CLOSE_RSP send to client stream error" err="tls: use of closed connection" connectionID=10280
I0609 16:00:18.229203       1 server.go:260] "Remove frontend for agent" frontend=&{Mode:http-connect Grpc:<nil> HTTP:0xc000868700 connected:0xc00098ede0 connectID:10280 agentID:4af8bf83-1f9a-4261-a0c8-ed7127db5ee2 start:{wall:13846413005822083990 ext:1129737651194 loc:0x21866c0} backend:0xc00020ed20} agentID="4af8bf83-1f9a-4261-a0c8-ed7127db5ee2" connectionID=10280
I0609 16:00:18.229228       1 server.go:721] "Close streaming" agentID="4af8bf83-1f9a-4261-a0c8-ed7127db5ee2" connectionID=10280
```
(Full logs can be found at -> 
[konnectivity-proxy-logs.txt](https://github.com/kubernetes-sigs/apiserver-network-proxy/files/6631460/konnectivity-proxy-logs.txt)

[...]

PR #239, opened 2021-06-17
don't use channels for moving packets

aojea:
channels have limitations on speed and size, hence they are not
a good solution for handling the proxied traffic

https://tpaschalis.github.io/channels-limitations-speed/

### DIFF
--- a/cmd/client/main.go
+++ b/cmd/client/main.go
@@ -210,6 +210,7 @@ func newGrpcProxyClientCommand(c *Client, o *GrpcProxyClientOptions) *cobra.Comm
 		Use:  "proxy-client",
 		Long: `A gRPC proxy Client, primarily used to test the Kubernetes gRPC Proxy Server.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			cmd.SilenceUsage = true
 			return c.run(o)
 		},
 	}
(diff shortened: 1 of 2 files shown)


aojea:
I don't know the logic well of the packet processing, so it needs a carefull review, i.e. there are several places where I don't know if it should return or just log and continue

aojea:
/hold
this is a test/poc

liggitt:
do we have benchmarks for throughput on this? if so, does this improve things? if not, we need them?

liggitt:
cc @jkh52

aojea:
> do we have benchmarks for throughput on this? if so, does this improve things? if not, we need them.

absolutely,

aojea:
IIUIC there is also another channel processing connection data in 

https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/5d2b5126a2f0c9640ca0394a0b3b5039fb96210b/konnectivity-client/pkg/client/client.go#L147

I've tried to give this a shot during this weekend but it is way more complex than I thought.

I don't know if this test is ok, but I only manage to get 1000 request with the client in the example https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/242

I'm wondering if these are one of the bottlenecks that show up here
https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/236 

Anyway, this rates seems too low to me, I want to speculate if this can be a probable cause of contention and and possible explaining some of the errors we saw in the e2e tests in k/k jobs

aojea:
wish I had more time to play with this, is really a nice project and challenge

PR #247, opened 2021-06-25
Change golang version to 1.16.4.

cheftako:


cheftako:
/close
Being handled in #248

PR #246, opened 2021-06-25
update to latest go to pickup critical http2 connection fixes

relyt0925:
This investigation done in Kubernetes has more details on the specific fixes:
 kubernetes/kubernetes#87615

There's critical failure paths in lower go versions below 1.15 where http2 connections when faced with certain network/connection errors can deadlock on open connections. The connections are left open but whenever a write or read is done from one it will log `: use of closed network connection` and do nothing.

I confirmed this behavior exists in the latest proxy:
```
Tylers-MacBook-Pro:armada-ansible tylerlisowski$ kubectl logs -n master-52643725 konnectivity-agent-8c6699584-nj6lk | grep 172.21.107.127
I0625 18:56:37.448913       1 options.go:98] AgentIdentifiers set to ipv4=172.21.224.93,ipv4=172.21.107.127,ipv4=172.21.161.36.
E0625 18:56:41.248418       1 client.go:473] "connection read failure" err="read tcp 172.30.210.115:58694->172.21.107.127:443: use of closed network connection"
E0625 18:56:41.248676       1 client.go:502] "conn write failure" err="write tcp 172.30.210.115:58694->172.21.107.127:443: use of closed network connection" connectionID=22
E0625 18:56:41.618133       1 client.go:473] "connection read failure" err="read tcp 172.30.210.115:58712->172.21.107.127:443: use of closed network connection"
E0625 18:56:41.618396       1 client.go:502] "conn write failure" err="write tcp 172.30.210.115:58712->172.21.107.127:443: use of closed network connection" connectionID=29
E0625 18:56:41.631325       1 client.go:473] "connection read failure" err="read tcp 172.30.210.115:58724->172.21.107.127:443: use of closed network connection"
E0625 18:56:42.452451       1 client.go:473] "connection read failure" err="read tcp 172.30.210.115:58746->172.21.107.127:443: use of closed network connection"
E0625 18:56:42.452548       1 client.go:502] "conn write failure" err="write tcp 172.30.210.115:58746->172.21.107.127:443: use of closed network connection" connectionID=45
E0625 18:57:13.675144       1 client.go:473] "connection read failure" err="read tcp 172.30.210.115:58872->172.21.107.127:443: use of closed network connection"
E0625 18:57:13.676248       1 client.go:502] "conn write failure" err="write tcp 172.30.210.115:58872->172.21.107.127:443: use of closed network connection" connectionID=47
[...]

PR #248, opened 2021-06-26
Enable http2 health checking with go 1.16.5

relyt0925:
builds on: https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/246

this pr will also turn on HTTP2 health checking which will auto heal the use of closed connection error.

### DIFF
--- a/cmd/client/main.go
+++ b/cmd/client/main.go
@@ -22,11 +22,13 @@ import (
 	"crypto/tls"
 	"flag"
 	"fmt"
+	"golang.org/x/net/http2"
 	"io/ioutil"
 	"net"
 	"net/http"
 	"os"
 	"os/signal"
+	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


relyt0925:
holding until go 1.16 pr goes in

go.mod
```diff
@@ -1,17 +1,17 @@
...
 	github.com/golang/mock v1.4.4
 	github.com/golang/protobuf v1.4.2
-	github.com/google/uuid v1.1.1
+	github.com/google/uuid v1.1.2
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
 	github.com/prometheus/client_golang v1.7.1
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/pflag v1.0.5
 	golang.org/x/net v0.0.0-20201110031124-69a78807bb2b
-	google.golang.org/grpc v1.27.0
```
cheftako, line 14: Not against upgrading grpc (I tested this myself a few weeks ago) but I would rather we separated this upgrade from the golang upgrade.

relyt0925: great point I agree

relyt0925: done

relyt0925:
cc @cheftako this should be good based off initial comments (or ready for another round of comments)

cmd/client/main.go
```diff
@@ -259,6 +261,10 @@ func (c *Client) run(o *GrpcProxyClientOptions) error {
 			transport := &http.Transport{
 				DialContext: dialer,
 			}
+			err = configureHTTP2Transport(transport)
```
cheftako, line 264: Can we wrap this in a CLI flag to control whether we have it turned on or not?
Otherwise looks good, thank you

relyt0925: can do!

relyt0925: done

PR #251, opened 2021-07-22
Fixed race condition writing to closed channel

cheftako:
Case is when agent gets a packet and attempts to proxy it forward.
However the channel to the connection we are going to forward to
has been closed.
As a result we get an error on writing to the channel.
One possibility is to use a lock to protect the channel for each packet.
Attempting a solution which does not need the lock.
Using a select does not work (https://play.golang.org/p/Az_YCJ24Cu1)
However also shown there is a send made safe using Recover().

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -50,6 +50,17 @@ func (c *connContext) cleanup() {
 	c.cleanOnce.Do(c.cleanFunc)
 }
 
+func (c *connContext) send(msg []byte) {
+	// TODO (cheftako@): Get perf test working and compare this solution with a lock based solution.
+	defer func() {
+		// Handles the race condition where we write to a closed channel
+		if err := recover(); err != nil {
+			klog.InfoS("Recovered from attempt to write to closed channel")
+		}
+	}()
+	c.dataCh <- msg
+}
+
 type connectionManager struct {
 	mu          sync.RWMutex
 	connections map[int64]*connContext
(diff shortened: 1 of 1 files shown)


pkg/agent/client.go
```diff
@@ -50,6 +50,14 @@ func (c *connContext) cleanup() {
 	c.cleanOnce.Do(c.cleanFunc)
 }
 
+func (c *connContext) send(msg []byte) {
+	defer func() {
+		// Handles the race condition where we write to a closed channel
+		recover()
```
jkh52, line 56: worth a warning log when this happens? Ideally would include connID.

caesarxuchao: Maybe an info log? We kind of know this will happen in normal execution flows.

pkg/agent/client.go
```diff
@@ -421,7 +429,7 @@ func (a *Client) Serve() {
 
 			ctx, ok := a.connManager.Get(data.ConnectID)
 			if ok {
-				ctx.dataCh <- data.Data
+				ctx.send(data.Data)
```
caesarxuchao, line 432: Walter and I talked about moving this dataCh write into the `connManager.Get`, and move the `close(dataCh)` at line 404 into the `connManager.Delete(connID)`. Then we use the `connManager.mu` to avoid a race. The problem with this solution is that holding `connManager.mu` affects all connections, not just this connection.

The other idea we had was adding a new lock to protect the dataCh. And since that Serve() function is single-threaded, we will not have any read contention. The only read-write contention will be when the cleanup function (which closes the dataCh) is called by another goroutine. The performance cost of adding such a low contention lock should be moderate, but we don't have benchmarks to validate this, so let's stick with this panic recovery fix for now.  @cheftako can you add a TODO about using a lock? Thanks.

PR #252, opened 2021-07-23
Make dial request processing asynchronous.

ScheererJ:
As suggested in #179, this is the second part of making the dial request processing asynchronous.

Previously, the net.Dial call based on a dial request may block the receive loop and with that all communication from the proxy server. We have seen instances where all traffic was blocked for roughly 10 minutes.
The problem is solved in this pull request by making the dial request processing asynchronous. (#179 already added a timeout.)

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -46,6 +46,7 @@ type connContext struct {
 	cleanFunc func()
 	dataCh    chan []byte
 	cleanOnce sync.Once
+	dialDone  chan struct{}
 }
 
 func (c *connContext) cleanup() {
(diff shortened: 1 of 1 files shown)


pkg/agent/client.go
```diff
@@ -399,21 +393,38 @@ func (a *Client) Serve() {
...
+					resp.GetDialResponse().Error = err.Error()
+					if err := a.Send(resp); err != nil {
+						klog.ErrorS(err, "could not send stream")
+					}
+					return
+				}
+				metrics.Metrics.ObserveDialLatency(time.Since(start))
+
+				ctx.conn = conn
+				close(dialDone)
```
caesarxuchao, line 417: Let's call this in a defer func in the beginning of the function, otherwise we might miss the `close` on early returns (e.g., the `return` at line 412), and that will lead to leaks.

caesarxuchao (commented):
cc @cheftako @jkh52 FYI

pkg/agent/client.go
ScheererJ: Moved the close as a deferred call directly to the beginning of the function as requested.

caesarxuchao:
Thanks, @ScheererJ. The changes look good.

Given the below reasons, I think we are not ready to merge this PR at the moment:
* @jkh52 and the team are working hard stablizing the K/K e2e tests with the network proxy;
* this changes the critical agent logic and doesn't have accompanying unit tests;
* #179 is merged which adds timeout on the dial, so I think this PR is less urgent to merge.

I think we can merge this PR when the k/k e2e tests are fixed.

jkh52 (changes requested):

PR #254, opened 2021-08-05
Updated server.go to include metrics for pending dial requests

jdnurme:


pkg/server/metrics/metrics.go
```diff
@@ -134,3 +154,13 @@ func (a *ServerMetrics) ConnectionDec(serviceMethod string) {
 func (a *ServerMetrics) SetBackendCount(count int) {
 	a.backend.WithLabelValues().Set(float64(count))
 }
+
+// PendingDialInc increments a pending agent dial.
```
cheftako, line 158: I think its safer to set the count then trust to always get the increment/decrement called.

pkg/server/server.go
```diff
@@ -418,6 +418,7 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
 				// The Dial is failing; no reason to keep this goroutine.
 				return
 			}
+			metrics.Metrics.PendingDialInc(metrics.ServeRecvFrontend)
```
cheftako, line 421: Can we do something like ```metrics.Metrics.PendingDialSet(len(s.PendingDial))```? That way if someone changes PendingDial without calling the metric inc/dec your haven't leaked a number.

pkg/server/metrics/metrics.go
jdnurme: ack

PR #257, opened 2021-08-18
add agent to backends after it is authenticated

PratikDeoghare:
In `ProxyServer.Connect` an agent is added to the backends and _then_ it is authenticated (if agent authentication is enabled). 
This means the agent is available for proxying requests while question of weather it is authentic or not is being decided. This clearly is a security risk. 

Please, take a look. Thanks!

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -614,16 +614,6 @@ func (s *ProxyServer) Connect(stream agent.AgentService_ConnectServer) error {
 	}
 
 	klog.V(2).InfoS("Connect request from agent", "agentID", agentID)
-	backend := s.addBackend(agentID, stream)
-	defer s.removeBackend(agentID, stream)
-
-	h := metadata.Pairs(header.ServerID, s.serverID, header.ServerCount, strconv.Itoa(s.serverCount))
-	if err := stream.SendHeader(h); err != nil {
-		return err
-	}
-
-	recvCh := make(chan *client.Packet, 10)
-	stopCh := make(chan error)
 
 	if s.AgentAuthenticationOptions.Enabled {
 		if err := s.authenticateAgentViaToken(stream.Context()); err != nil {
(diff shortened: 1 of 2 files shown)


jkh52:
Looks like a unit test needs fixing, and you need the CLA.

I agree with this change.

PratikDeoghare:
I signed it

cheftako:
I think you may need to make fixes near https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/server/server_test.go#L155

cheftako:
You may need to amend the commit as per https://github.com/kubernetes/community/blob/master/CLA.md#4-ensure-your-github-e-mail-address-matches-address-used-to-sign-cla to get the cla check to pass.

PratikDeoghare:
Thanks for looking. :pray:  I will fix the tests and then figure out whats up with the CLA.

cheftako:
Nice find and fix by the way. Thank you for helping :100:

dberkov:
/lgtm

@PratikDeoghare thanks for making this fix.

cheftako:
/lgtm

Now we just need to fix the CLA.

dberkov:
nit: squashing the changes to single commit might simplify CLA check to pass

cheftako:
I'd love to get this merged once the CLA passes.

PratikDeoghare:
Squashed the commits. Let's see if this passes. I never got HelloSign email tho. :confused: 

/check-cla

PratikDeoghare:
Fixed CLA. :tada:

pkg/server/server.go
```diff
@@ -632,12 +622,23 @@ func (s *ProxyServer) Connect(stream agent.AgentService_ConnectServer) error {
 		}
 	}
 
+	h := metadata.Pairs(header.ServerID, s.serverID, header.ServerCount, strconv.Itoa(s.serverCount))
+	if err := stream.SendHeader(h); err != nil {
```
cheftako, line 626: Reversing the order of SendHeader() and addBackend() is also a significant change. Might be worth at least a comment. Calling SendHeader() will unblock the agent. (https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/agent/client.go#L225) I think that's ok, because the agent only adds the connection to the list of connected Konn Servers. It won't start using that connection until a frontend is using that connection. An occurrence which cannot currently happen before the agent has been added to the list of backends below. Worth noting that this assumption becomes less true with #176.

PratikDeoghare: Ah I see. Thanks! 

Ptal https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/259

cheftako: Not convinced that reversing the orders is wrong, just its a change we need to think through. Until #176 lands I think this way is better. I think we should leave it this way for now and determine what we want to do once #176 is ready.

PratikDeoghare: I misunderstood your earlier comment then. Apologies. 
We can keep it this way for now and decide after #176. :+1:

PR #260, opened 2021-08-25
add/remove agent to default backend only if it accepts all traffic

PratikDeoghare:
The agents that specify `--agent-identifers` expect traffic going only to the specified destination. They shouldn't get any other traffic. 

`X`: Currently all the agents are added to the default backend. 

Scenario: 
1. Agent `a1` registers to handle traffic going to `a1.com`. 
2. Agent `a2` registers does not specify any agent identifiers so any traffic can be sent to it. 
3. A request is made to `b.com`. 
4. Since no agent is registered to specifically handle `b.com` we choose randomly from all the default agents. 
5. Because of `X` all the agents are there and `a1` gets chosen. 
6. The request fails. It should have gone to `a2`. 

This PR fixes the problem by adding an agent to the default backend only if it doesn't specify any agent identifiers.  

The fix is based on the assumption that an agent not specifying identifiers means any traffic can be sent to it. Let me know if I assumed wrong. 

Please, take a look. Thanks! :pray:

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -378,13 +378,13 @@ func (a *Client) Serve() {
 
 		switch pkt.Type {
 		case client.PacketType_DIAL_REQ:
-			klog.V(4).Infoln("received DIAL_REQ")
+			dialReq := pkt.GetDialRequest()
+			klog.V(4).Infoln("received DIAL_REQ to: ", dialReq.Address)
 			resp := &client.Packet{
 				Type:    client.PacketType_DIAL_RSP,
 				Payload: &client.Packet_DialResponse{DialResponse: &client.DialResponse{}},
 			}
 
-			dialReq := pkt.GetDialRequest()
 			resp.GetDialResponse().Random = dialReq.Random
 
 			start := time.Now()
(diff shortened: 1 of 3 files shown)


cheftako:
> The agents that specify --agent-identifers expect traffic going only to the specified destination. They shouldn't get any other traffic.

I would have thought we controller this via the proxy server's strategy flag. In a homogeneous agent set, if I don't want to send my traffic to fail through to the default backend-manager I can just not specify default in the fail-over strategies.  I can see where this might have value with a heterogeneous agent set. However in that case, I think either behavior might be desirable. Would it make sense to have an additional field such as agent-restriction or agent-strategy, which determines which backend-managers the agent supports?

PratikDeoghare:
Hi @cheftako  @mihivagyok 

I agree. It is possible to set things up to avoid the problem described. 

But... 

What if 
1. Agents simply specified what traffic they can handle
    a1: google.com, default
    a2: default
    a3: facebook.com, default 
    a4: google.com 
2. Proxy-server picks random agent out of subset of agents that can handle the given request. 
    req to google.com -- a1 or a4 will be picked 
    req to github.com -- a1, a2 or a3 will be picked 
    and so on...  

Wouldn't this be a simpler design? We won't need the concept of proxy strategies at proxy-server.

PR #264, opened 2021-09-10
Add additional lost wait request test.

cheftako:
Added a setting for the test server.
This allows us to test what happens if the dial request isn't returned.
Currently this causes the client to hang.

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -171,6 +171,9 @@ type Client struct {
 
 	cs *ClientSet // the clientset that includes this AgentClient.
 
+	serverChannel    chan *client.Packet
+	cleanChannel     sync.Once
+	serverError      error
 	stream           agent.AgentService_ConnectClient
 	agentID          string
 	agentIdentifiers string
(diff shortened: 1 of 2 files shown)


cmd/test-server/main.go
```diff
@@ -215,3 +276,91 @@ func (p *TestServer) runTestServer(ctx context.Context, o *TestServerRunOptions)
...
+	return &FlakyListener{
+		inner: ln,
+		flipper: false,
+		mode: mode,
+	}, nil
+}
+
+// Accept waits for and returns the next connection to the listener.
+func (f* FlakyListener) Accept() (net.Conn, error) {
+	c, err := f.inner.Accept()
```
caesarxuchao, line 330: I'm not familiar with how Accept() works. Maybe when inner.Accept() is called, from the client's perspective, the dial has succeeded, so we never trigger the dial timeout in the proxy agent.

The fact that flakyListener.Accept returns a net.Conn half of the time perhaps only affects how the following HTTP requests are handled but doesn't affect the dial.

PR #265, opened 2021-09-10
Improve logging around agent stream read/write.

cheftako:
Added connection id and server id to agent logging for stream
processing.
Also added logging to track when the relevant goroutines exit.

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -467,6 +467,13 @@ func (a *Client) Serve() {
 }
 
 func (a *Client) remoteToProxy(connID int64, ctx *connContext) {
+	defer func() {
+		if panicInfo := recover(); panicInfo != nil {
+			klog.V(2).InfoS("Exiting remoteToProxy with recovery", "panicInfo", panicInfo, "connectionID", connID)
+		} else {
+			klog.V(2).InfoS("Exiting remoteToProxy", "connectionID", connID)
+		}
+	}()
 	defer ctx.cleanup()
 
 	var buf [1 << 12]byte
@@ -479,11 +486,11 @@ func (a *Client) remoteToProxy(connID int64, ctx *connContext) {
 		klog.V(5).InfoS("received data from remote", "bytes", n, "connectionID", connID)
 
 		if err == io.EOF {
-			klog.V(2).Infoln("connection EOF")
+			klog.V(2).InfoS("connection EOF", "connectionID", connID)
 			return
 		} else if err != nil {
 			// Normal when receive a CLOSE_REQ
-			klog.ErrorS(err, "connection read failure")
+			klog.ErrorS(err, "connection read failure", "connectionID", connID)
 			return
 		} else {
 			resp.Payload = &client.Packet_Data{Data: &client.Data{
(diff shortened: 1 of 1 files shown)


pkg/agent/client.go
```diff
@@ -467,6 +467,13 @@ func (a *Client) Serve() {
 }
 
 func (a *Client) remoteToProxy(connID int64, ctx *connContext) {
+	defer func() {
+		if err := recover(); err != nil {
+			klog.V(2).InfoS("Exiting remoteToProxy with recovery", "error", err, "connectionID", connID)
+		} else {
+			klog.V(2).InfoS("Exiting remoteToProxy", "connectionID", connID)
+		}
+	}()
```
mainred, line 476: Thanks. So that the other existing connections won't be impacted by one connection failure.

pkg/agent/client.go
```diff
@@ -479,25 +486,35 @@ func (a *Client) remoteToProxy(connID int64, ctx *connContext) {
...
 
 func (a *Client) proxyToRemote(connID int64, ctx *connContext) {
+	defer func() {
+		if err := recover(); err != nil {
+			klog.V(2).InfoS("Exiting proxyToRemote with recovery", "error", err, "connectionID", connID)
+		} else {
+			klog.V(2).InfoS("Exiting proxyToRemote", "connectionID", connID)
+		}
+	}()
+	defer func() {
```
mainred, line 515: Remove this defer? Because we covered this in the above defer.
return of recover() is nil only when no panic happens, so L512 works in the same way.

pkg/agent/client.go
```diff
@@ -479,25 +486,35 @@ func (a *Client) remoteToProxy(connID int64, ctx *connContext) {
...
-				klog.ErrorS(err, "stream send failure")
+				klog.ErrorS(err, "stream send failure", "connectionID", connID)
 			}
 		}
 	}
 }
 
 func (a *Client) proxyToRemote(connID int64, ctx *connContext) {
+	defer func() {
+		if err := recover(); err != nil {
```
mainred, line 509: Not sure `error` is the correct name for the return value of `recover()` because it's more like info from panic, instead of an error caused by `recover()`

mainred (commented):
Thanks for your change.

pkg/agent/client.go
cheftako: Side effect of what I was trying to accomplish. I think that's ok but happy to chat if others think its better to crash and restart. Mostly want to track whawt could be causing these listeners to go down.

pkg/agent/client.go
cheftako: Sorry this was leftover which should have been removed.

pkg/agent/client.go
cheftako: I agree "error" is close but not quite right. "Info" is to vague. "panic"? "exception"?

mainred: panicInfo to be specific?

caesarxuchao: In Golang's [blog](https://go.dev/blog/defer-panic-and-recover) they just use 'r'. I think `err` is ok. After all, this is just a local variable.

caesarxuchao (commented):
LGTM.

Please address https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/265/files#r707074459.

pkg/agent/client.go
cheftako: I think we are also implicitly discussing the structured logging name. So in that sense it is more than just the local variable.

mainred (approved):
Thanks

Issue #167, opened 2021-01-21
quick and multiple Ctrl+C interrupt for kubectl logs -f cause konnectivity-agent stuck


ydp:
Hi team,

I met a strange issue in our product that when I use `kubectl logs -f konnectivity-agent-xxx` to streaming logs, because there are many logs there, I did not wait it to finish and use `Ctrl+C` to interrupt the streaming, since it did not exit immediately like cat big file, I issued `Ctrl+C` at least 3 times as quick as I can. Then the streaming process exit, but konnectivity got stuck. This looks like a random behavior, but after several experiments, it actually can be reproduced in any of my setups.

konnectivity server/client version: v0.0.12
setup: one proxy-server on k8s master vm and one proxy-agent run as pod(one replica deployment).

The logs when `Ctrl+C` issued:
server:
```
I0121 06:53:12.616717       7 server.go:224] proxy request from client, userAgent [grpc-go/1.27.0]
I0121 06:53:12.616941       7 server.go:257] start serving frontend stream
I0121 06:53:12.617098       7 server.go:268] >>> Received DIAL_REQ
I0121 06:53:12.617118       7 backend_manager.go:170] pick agentID=a89e1d16-8908-4f83-bbc1-bea45ba34ea4 as backend
I0121 06:53:12.617168       7 server.go:290] >>> DIAL_REQ sent to backend
I0121 06:53:12.740584       7 server.go:308] >>> Received 24 bytes of DATA(id=195)
I0121 06:53:12.740750       7 server.go:324] >>> DATA sent to Backend
I0121 06:53:12.740818       7 server.go:294] >>> Received CLOSE_REQ(id=195)
I0121 06:53:12.740842       7 server.go:303] >>> CLOSE_REQ sent to backend
I0121 06:53:13.353224       7 server.go:308] >>> Received 24 bytes of DATA(id=194)
I0121 06:53:13.353335       7 server.go:324] >>> DATA sent to Backend
I0121 06:53:13.353347       7 server.go:294] >>> Received CLOSE_REQ(id=194)
I0121 06:53:13.353357       7 server.go:303] >>> CLOSE_REQ sent to backend
I0121 06:53:13.405226       7 server.go:308] >>> Received 24 bytes of DATA(id=189)
I0121 06:53:13.405460       7 server.go:324] >>> DATA sent to Backend
I0121 06:53:13.405484       7 server.go:294] >>> Received CLOSE_REQ(id=189)
[...]

PR #266, opened 2021-09-18
remove unnecessary goroutines and channels

PratikDeoghare:
I think we don't really need the channels and goroutines there. Please, take a look.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -361,46 +361,30 @@ func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
 	userAgent := md.Get(header.UserAgent)
 	klog.V(2).InfoS("proxy request from client", "userAgent", userAgent)
 
-	recvCh := make(chan *client.Packet, 10)
-	stopCh := make(chan error)
-
-	go s.serveRecvFrontend(stream, recvCh)
-
-	defer func() {
-		close(recvCh)
-	}()
-
-	// Start goroutine to receive packets from frontend and push to recvCh
-	go func() {
-		for {
-			in, err := stream.Recv()
-			if err == io.EOF {
-				close(stopCh)
-				return
-			}
-			if err != nil {
-				klog.ErrorS(err, "Stream read from frontend failure")
-				close(stopCh)
-				return
-			}
-
-			recvCh <- in
-		}
-	}()
-
-	return <-stopCh
+	s.serveRecvFrontend(stream)
+	return nil
 }
 
-func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer, recvCh <-chan *client.Packet) {
+func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer) {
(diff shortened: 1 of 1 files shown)


pkg/server/server.go
```diff
@@ -361,46 +361,30 @@ func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
...
+	// backend from the BackendManger then.
+	var backend Backend
 
-	// Start goroutine to receive packets from frontend and push to recvCh
-	go func() {
-		for {
-			in, err := stream.Recv()
+	for {
+		pkt, err := stream.Recv()
+		if err != nil {
```
jkh52, line 378: inadvertent? Otherwise is redundant with line 382.

pkg/server/server.go
```diff
@@ -361,46 +361,30 @@ func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
 	userAgent := md.Get(header.UserAgent)
 	klog.V(2).InfoS("proxy request from client", "userAgent", userAgent)
 
-	recvCh := make(chan *client.Packet, 10)
```
jkh52, line 364: we're losing this 10 packet buffer. Having it could have possibly helped avoid grpc-layer lock-up issues between proxy-server and proxy-agent, similar to what we saw between KAS and proxy-server at #167.

Mentioning for awareness, not necessarily as a blocker.

jkh52:
LGTM otherwise.

cheftako:
Not convinced this is safe. The channel stops blocking send (Eg. https://github.com/grpc/grpc-go/blob/master/stream.go#L1387 which is what frontend.Send() ends up calling for grpc mode) from preventing us from listening. This stops working when the channel is fulll but I think the first step is to add a warning when the channel is full. If the channel does become full, we need a different fix.

jkh52:
Are you still looking to proceed (there is a subset of this PR that could still be merged, just not all of it)?

PratikDeoghare:
@jkh52 
> Are you still looking to proceed (there is a subset of this PR that could still be merged, just not all of it)?

Sorry I have been a bit busy. 

Nothing urgent about this PR. It was created because it looked like the code can be simplified. 

I will try to adjust the PR to keep the buffer.

PratikDeoghare:
@cheftako 
> Not convinced this is safe. The channel stops blocking send (Eg. https://github.com/grpc/grpc-go/blob/master/stream.go#L1387 which is what frontend.Send() ends up calling for grpc mode) from preventing us from listening. 

I see. Yes. 

> This stops working when the channel is fulll but I think the first step is to add a warning when the channel is full. If the channel does become full, we need a different fix.

I have not observed channel getting full.

PratikDeoghare:
> I will try to adjust the PR to keep the buffer.

Looks like it is hard to do without essentially reverting these changes. 
We will need a goroutine to run Recv and push it into the channel that acts as buffer. 
Then we read from the channel. This is what the code without the changes in this PR does anyway. 

Not sure about the next move. Happy to close the PR if it is not useful. 

Thanks for looking and discussing. :pray:

PR #269, opened 2021-10-03
Instrumenting connection issues on Proxy Server

cheftako:
Improving debugging of anything causing problems sending on Proxy
Server.  Further instrumenting connection possible connection
issues on the Proxy Server.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -367,6 +367,7 @@ func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
 	go s.serveRecvFrontend(stream, recvCh)
 
 	defer func() {
+		klog.V(2).InfoS("Receive channel on Proxy is stopping", "userAgent", userAgent, "serverID", s.serverID)
 		close(recvCh)
 	}()
 
@@ -375,6 +376,7 @@ func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
 		for {
 			in, err := stream.Recv()
 			if err == io.EOF {
+				klog.V(2).InfoS("Stream closed on Proxy", "userAgent", userAgent, "serverID", s.serverID)
 				close(stopCh)
 				return
 			}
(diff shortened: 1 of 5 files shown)


pkg/server/server.go
```diff
@@ -652,6 +664,9 @@ func (s *ProxyServer) Connect(stream agent.AgentService_ConnectServer) error {
 				return
 			}
 
+			if len(recvCh) > 9 {
```
jkh52, line 667: nit: I would declare the other constant (recvCh size) so the two never get out of sync.

pkg/server/server.go
```diff
@@ -652,6 +664,9 @@ func (s *ProxyServer) Connect(stream agent.AgentService_ConnectServer) error {
...
+				klog.V(2).InfoS("Receive channel on Connect is full", "agentID", agentID, "serverID", s.serverID)
```
jkh52, line 668: nit: maybe warning level?

jkh52:
This is a really useful change, thanks!

LGTM - two nits only (if you do either, note the pattern is present in two places).

pkg/server/server.go
cheftako: Not if we want to stay with structured logging.
https://github.com/kubernetes/klog/pull/184#issuecomment-717405511

jkh52:
Calling len() or cap() on a closed channel seems safe (I double checked in Go Playground), so what precisely is unsafe? I want to make sure I understand.

Nits:
- It would be nice for some unit test codepath to enable the new flag, and/or we could enable in k/k e2e.
- I think there's a corollary for the agent (dataCh?) we could potentially make (similar new flag).

All non-blockers, this seems a fine checkpoint.

/lgtm

cheftako:
> Calling len() or cap() on a closed channel seems safe (I double checked in Go Playground), so what precisely is unsafe? I want to make sure I understand.

This issue is when len() is called on the live channel. We have two goroutines each modifying the size of the channel (one adding to it; the other removing from it). Observing the size then is not safe on either goroutine. Currently we are observing the size on the goroutine adding to the channel. To make it safe (not effected by a race condition) I would need to observe the length under a lock. However I would then need to remove from the channel under lock on the other goroutine. However removing/dequeuing from the channel under lock is a problem. The dequeue will block if the channel is empty. That means I am then blocked holding the lock and it becomes impossible to enqueue data to the lock. At that point we are just deadlocked.

As it stands I don't think the code will cause any problems. Sadly if I test the code, it will trigger the race detection we have turned on in our tests. I generally find the race detection in our tests useful. It has caught several real bugs and made it easy to both fix and feel confident we have fixed the issues. However that leaves me needing to hide this race condition from our tests.

PR #270, opened 2021-10-06
Fix obscure HTTP CONNECT goroutine leak

jveski:
The connection chan is not closed for DIAL_RSP, which means http handler goroutines stall if the agent fails to dial. This probably went unnoticed because the client connections close as expected and it takes a very long time for enough goroutines to pile up and consume noticeable resources.

To reproduce, just generate a few thousand connections through the tunnel while watching the go_goroutines metric.

### DIFF
--- a/tests/proxy_test.go
+++ b/tests/proxy_test.go
@@ -10,6 +10,7 @@ import (
 	"net/http/httptest"
 	"net/url"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


tests/proxy_test.go
```diff
@@ -409,10 +477,17 @@ func runHTTPConnProxyServer() (proxy, func(), error) {
 	proxy.agent = localAddr(lis.Addr())
 
 	// http-connect
+	active := int32(0)
+	proxy.getActiveHTTPConnectConns = func() int { return int(atomic.LoadInt32(&active)) }
```
cheftako, line 481: Can we get this added to metrics?

PR #279, opened 2021-10-29
Add an agent side keep alive.

cheftako:
Added a agent side grpc keepalive.
Currently defaults to 1 hour as per server side keepalive.
Configurable to a much shorter period.


jkh52:
As @Jefftree pointed out this may be overlapping with: https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/agent/client.go#L536, do we need both?

Issue #276, opened 2021-10-27
Leaking/Too many open File descriptors in Konnectivity server with GRPC mode and UDS transport


rata:
Hi!

This investigation is joint work we are doing with @ipochi, I’m just opening the issue because it is late in his timezone now :)

We are trying to adopt Konnectivity v0.0.24 and we faced some memory issues when using it in GRPC mode with UDS (unix sockets). In production the mem usage of the konnectivity container was too high and restarted due to that, on several clusters. For now, we switched to http-connect mode and we are having more luck with that so far.

Nevertheless, we are trying to investigate the GRPC issue in a lab setup. For this, @ipochi created [a CLI tool][konnscen] using client-go to reproduce different scenarios. This just gets pod logs on some random pods, with different levels of concurrency.

Using that tool, we can easily reproduce a scenario where we see a lot of open fds on the konnectivity server, even when the proxy is idle for ~1 day. And if we hit it with more load, the number of fds seems to always increase (>= than before). We also see a big number of goroutines too. On a quick look it seems several might be per connection, to receive and such. But we haven’t dug deeper into that yet.

I’m using an example with a small number of fds here on what I will paste, but with more load we can easily open 2k+ fds and the same observations here are true when more fds are open too. One konnscen config.yaml that can be used for that is:
```
concurrent_connections:
  number_of_concurrent_users: 50
  number_of_times: 2
```

Taking a closer look into the fds, we can see that the fds are sockets connections. In this case, 60020 is the PID of the konnectivity proxy server:
```
# ls -l /proc/60020/fd
lrwx------. 1 root root 64 Oct 26 21:59 0 -> /dev/null
l-wx------. 1 root root 64 Oct 26 21:59 1 -> 'pipe:[203909]'
lrwx------. 1 root root 64 Oct 25 16:24 10 -> 'socket:[201974]'
lrwx------. 1 root root 64 Oct 25 16:24 11 -> 'socket:[204824]'
lrwx------. 1 root root 64 Oct 27 13:18 12 -> 'socket:[8739504]'
lrwx------. 1 root root 64 Oct 25 16:24 13 -> 'socket:[199329]'
lrwx------. 1 root root 64 Oct 27 13:18 16 -> 'socket:[8735695]'
lrwx------. 1 root root 64 Oct 27 13:18 19 -> 'socket:[8737576]'
l-wx------. 1 root root 64 Oct 26 21:59 2 -> 'pipe:[203910]'
[...]

PR #280, opened 2021-10-30
proxy-server: in gRPC mode, set same keepalive for UDS and mTLS.

jkh52:
Fix an inconsistency that was identified in #276

### DIFF
--- a/cmd/server/app/server.go
+++ b/cmd/server/app/server.go
@@ -170,7 +170,10 @@ func (p *Proxy) runUDSFrontendServer(ctx context.Context, o *options.ProxyRunOpt
 	}
 	var stop StopFunc
 	if o.Mode == "grpc" {
-		grpcServer := grpc.NewServer()
+		frontendServerOptions := []grpc.ServerOption{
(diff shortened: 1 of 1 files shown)


jkh52:
/retest

TestFailedDial_HTTPCONN seems like a flake (only modifed gRPC codepath)

cheftako:
> TestFailedDial_HTTPCONN seems like a flake (only modifed gRPC codepath)
Worrying. Would be good to capture the error and fix at some point. If you see this again, can you capture the failure in a bug?

rata:
Thanks!

Sorry, with @ipochi we were waiting for confirmation if this helped or not before opening the PR!

rata:
@jkh52 Just curious, do you happen to know if the keepalive was not used for the UDS case before on purpose?

By looking at the git history, it seems the keepalive config was _never_ there for the UDS case and I'm not sure if the grpc PING command (that the keepalive.Time config sends), that AFAIK uses the http/2 ping command underneath, works over UDS sockets. Do you happen to know? I guess it does, but not _sure_.

I agree, though, that this won't probably hurt in any way, though, and seems like it was forgotten. But we were trying to validate that before opening the PR, we can open the PR without doing those things the next time :)

PR #281, opened 2021-11-01
Bubble up errors on grpc connections.

rata:
The Connect() and Proxy() method always return nil today, as they return
what is received in the stopCh (see [here][link1] and [here][link2]).
When the channel is closed, the nil error is received and then returned.

This patch just changes to send the error (instead of closing the
channel) when an error was received. This way, we report the proper
error to the grpc layer.

This is mostly a no-op IIUC, but the recommended way in the grpc
quick start.

[link1]: https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/a007c42f6c9d06ff869da5b93c428546658c414e/pkg/server/server.go#L402
[link2]: https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/a007c42f6c9d06ff869da5b93c428546658c414e/pkg/server/server.go#L680

Signed-off-by: Rodrigo Campos <rodrigo@kinvolk.io>

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -388,6 +388,7 @@ func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
 			}
 			if err != nil {
 				klog.ErrorS(err, "Stream read from frontend failure")
+				stopCh <- err
 				close(stopCh)
 				return
 			}
@@ -666,6 +667,7 @@ func (s *ProxyServer) Connect(stream agent.AgentService_ConnectServer) error {
 			}
 			if err != nil {
 				klog.ErrorS(err, "stream read failure")
+				stopCh <- err
 				close(stopCh)
 				return
 			}


rata:
I'm member of the kubernetes org, but not kubernetes-sigs, so can't ok-to-test this. Can someone ok-to-test this? :)

cheftako:
I generally like the fix but would prefer if we still made sure something closes the stop channel.

rata:
@cheftako it is common to not close channels in go. Please note [this comment][link] from Ian Lance Taylor, go maintainer:
 > Note that it is only necessary to close a channel if the receiver is
looking for a close. Closing the channel is a control signal on the
channel indicating that no more data follows

I can change the code to send the error and then close the channel, but the receiver is not waiting on that close. Let me know what you prefer :)

[link]: https://groups.google.com/g/golang-nuts/c/pZwdYRGxCIk/m/qpbHxRRPJdUJ

cheftako:
I agree it is unnecessary to close the channel. However we are spending a lot of our time right now trying to ensure we aren't leaking resources. This means we are looking to ensure things like connections are properly closed. This tends to trigger at least my basic pattern match for what a leak looks like. I think it will save time in the long run to not resemble a leak.

rata:
@cheftako ohh, good point. Yeah, that will make the life simpler for our future selves :)

Updated to send the error before the close, then! PTAL :)

PR #282, opened 2021-11-03
add option to change cipher suites

Tamas-Biro1:
resolves: https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/283

This PR adds an option to change ANP server default cipher suite to specify a more strict list.

### DIFF
--- a/cmd/server/app/options/options.go
+++ b/cmd/server/app/options/options.go
@@ -11,6 +11,7 @@ import (
 	"k8s.io/klog/v2"
 
 	"sigs.k8s.io/apiserver-network-proxy/pkg/server"
+	"sigs.k8s.io/apiserver-network-proxy/pkg/util"
 )
 
 type ProxyRunOptions struct {
@@ -77,6 +78,13 @@ type ProxyRunOptions struct {
 	// blocking call has its own problems, so it cannot easily be made race condition safe.
 	// The check is an "unlocked" read but is still use at your own peril.
 	WarnOnChannelLimit bool
+
+	// Cipher suites used by the server.
+	// If empty, the default suite will be used from tls.CipherSuites(),
+	// also checks if given comma separated list contains cipher from tls.InsecureCipherSuites().
+	// NOTE that cipher suites are not configurable for TLS1.3,
+	// see: https://pkg.go.dev/crypto/tls#Config, so in that case, this option won't have any effect.
+	CipherSuites string
 }
 
 func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
@@ -108,6 +116,7 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.AuthenticationAudience, "authentication-audience", o.AuthenticationAudience, "Expected agent's token authentication audience (used with agent-namespace, agent-service-account, kubeconfig).")
 	flags.StringVar(&o.ProxyStrategies, "proxy-strategies", o.ProxyStrategies, "The list of proxy strategies used by the server to pick a backend/tunnel, available strategies are: default, destHost.")
 	flags.BoolVar(&o.WarnOnChannelLimit, "warn-on-channel-limit", o.WarnOnChannelLimit, "Turns on a warning if the system is going to push to a full channel. The check involves an unsafe read.")
+	flags.StringVar(&o.CipherSuites, "cipher-suites", o.CipherSuites, "The comma separated list of allowed cipher suites. Has no effect on TLS1.3. Empty means allow default list.")
 	return flags
 }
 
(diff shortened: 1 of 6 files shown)


cmd/server/app/options/options.go
```diff
@@ -108,6 +115,7 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.AuthenticationAudience, "authentication-audience", o.AuthenticationAudience, "Expected agent's token authentication audience (used with agent-namespace, agent-service-account, kubeconfig).")
 	flags.StringVar(&o.ProxyStrategies, "proxy-strategies", o.ProxyStrategies, "The list of proxy strategies used by the server to pick a backend/tunnel, available strategies are: default, destHost.")
 	flags.BoolVar(&o.WarnOnChannelLimit, "warn-on-channel-limit", o.WarnOnChannelLimit, "Turns on a warning if the system is going to push to a full channel. The check involves an unsafe read.")
+	flags.StringVar(&o.CipherSuites, "cipher-suites", o.CipherSuites, "The comma separated list of allowed cipher suites.")
```
cheftako, line 118: Would be good to add something like "Empty means allow default list."

cmd/server/app/server.go
```diff
@@ -46,6 +47,30 @@ func NewProxyCommand(p *Proxy, o *options.ProxyRunOptions) *cobra.Command {
...
+	for _, v := range tls.CipherSuites() {
+		acceptedCiphers[v.Name] = v.ID
+	}
+
+	// compare user given ciphers and accepted ciphers, and collect them
+	ciphersIntSlice := make([]uint16, 0)
+	for _, cipher := range cipherNames {
+		intValue, ok := acceptedCiphers[cipher]
+		if !ok {
+			return nil, fmt.Errorf("cipher suite %s not supported, doesn't exist or considered as insecure", cipher)
```
cheftako, line 67: Any chance this could be moved into the validation functions?

jkh52:
The test failure looks related:

```
# sigs.k8s.io/apiserver-network-proxy/cmd/server/app
cmd/server/app/server.go:57:49: undefined: tls.CipherSuites
cmd/server/app/server.go:58:20: undefined: tls.CipherSuites
note: module requires Go 1.17
```

Note that #275 introduces 1.17 here. But the prow log have "golang:1.13" (see Job Pod info), does anybody know how to update that?

cheftako:
> The test failure looks related:
> 
> ```
> # sigs.k8s.io/apiserver-network-proxy/cmd/server/app
> cmd/server/app/server.go:57:49: undefined: tls.CipherSuites
> cmd/server/app/server.go:58:20: undefined: tls.CipherSuites
> note: module requires Go 1.17
> ```
> 
> Note that #275 introduces 1.17 here. But the prow log have "golang:1.13" (see Job Pod info), does anybody know how to update that?

https://github.com/kubernetes/test-infra/blob/d527235e962a094b65925249afd6dce0585194bd/config/jobs/kubernetes-sigs/apiserver-network-proxy/apiserver-network-proxy-presubmits.yaml

cmd/server/app/options/options.go
Tamas-Biro1: thanks for the suggestion, added

cmd/server/app/server.go
Tamas-Biro1: moved into the validation functions

cmd/server/app/options/options.go
```diff
@@ -77,6 +78,13 @@ type ProxyRunOptions struct {
 	// blocking call has its own problems, so it cannot easily be made race condition safe.
 	// The check is an "unlocked" read but is still use at your own peril.
 	WarnOnChannelLimit bool
+
+	// Cipher suites used by the server.
+	// If empty, the default suite will be used from tls.CipherSuites(),
+	// also checks if given comma separated list contains cipher from tls.InsecureCipherSuites().
+	// NOTE that cipher suites are not configurable for TLS1.3,
+	// see: https://pkg.go.dev/crypto/tls#Config, so in that case, this option won't have any effect.
```
mihivagyok, line 86: Maybe we could add the TLS version to the configuration also. Otherwise, user would not know what TLS version is used in the code. Default would TLS1.2, but would be possible to set it to TLS1.3. What do you think? Thank you!

go.mod
```diff
@@ -20,38 +20,4 @@ require (
 	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.0
 )
 
-require (
```
jkh52, line 23: I see now some 'go mod tidy' issues in tests. I think this is an incorrect deletion; try running tidy with go 1.17.

cmd/server/app/options/options.go
Tamas-Biro1: TLS version is the min version that server could accept. But you are right, for TLS1.3 it won't have any effect, so I added this info to the option arguments. Thank you!

go.mod
Tamas-Biro1: fixed

cmd/server/app/options/options.go
```diff
@@ -139,6 +148,7 @@ func (o *ProxyRunOptions) Print() {
 	klog.V(1).Infof("KubeconfigBurst set to %d.\n", o.KubeconfigBurst)
 	klog.V(1).Infof("ProxyStrategies set to %q.\n", o.ProxyStrategies)
 	klog.V(1).Infof("WarnOnChannelLimit set to %t.\n", o.WarnOnChannelLimit)
+	klog.V(1).Infof("CipherSuites set to %q.\n", o.CipherSuites)
```
cheftako, line 151: Might be nice to print what the default suite is as a follow up if that's not too difficult.

PR #287, opened 2021-11-05
Make sure deps are clean prior to running update vendor.

cheftako:
Update instructions to ensure K/K dependencies are clean before running update vendor.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -41,5 +41,6 @@ Creating a new release of network proxy involves releasing a new version of the 
 
     ```
     ./hack/pin-dependency.sh sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.15
+    make clean generated_files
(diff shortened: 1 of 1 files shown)


Jefftree:
Just out of curiosity what was the scenario in which this was needed? Network proxy doesn't have any code generation in k/k so if we start from a clean workspace, upgrade network proxy and run update-vendor, I don't see any reason the generated files would change.

cheftako:
> Just out of curiosity what was the scenario in which this was needed? Network proxy doesn't have any code generation in k/k so if we start from a clean workspace, upgrade network proxy and run update-vendor, I don't see any reason the generated files would change.

https://github.com/kubernetes/kubernetes/pull/106188#discussion_r743898778 is an example. I ended up with spurious changes in the go.mod file. Running the clean and then the update vendor fixed it. I assume this is from have older generated files lying around.

PR #289, opened 2021-11-11
konnectivity-client: improve logs for DIAL_RSP not recognized case.

jkh52:


konnectivity-client/pkg/client/client.go
```diff
@@ -114,13 +114,13 @@ func (t *grpcTunnel) serve(c clientConn) {
 			t.pendingDialLock.RUnlock()
 
 			if !ok {
-				klog.V(1).Infoln("DialResp not recognized; dropped")
+				klog.V(1).InfoS("DialResp not recognized; dropped", "connectionID", resp.ConnectID, "dialID", resp.Random)
```
cheftako, line 117: I dislike that we call it "random" on most of the other log statements but we do mostly call it that. Can we change other locations which use "random" to use "dialID"?

jkh52: Done (let me know if I missed any)

jkh52:
@cheftako

PR #290, opened 2021-11-12
Don't leak konn-client connections if DialResp is not recognized

rata:
If the DialResp is not recognized, it seems we never will receive a
close request in this stream. So, we can close this connection by finishing the goroutine (a defer makes sure the connection is closed when we finish)

This patch greatly helps to reduce the leaks reported in #276. Although,
this doesn't completely fix the issue. The number of leaked fds seems to be ~1/10 with this patch, sometimes even less.

However, I have some open questions that you can maybe answer:
1. Is the close request/response for some reason always expected to come, even in this case, and therefore is there some other part of the code that we also need to fix?
2. Even if the answer to ^ is yes, I think closing the connection here (if we will only receive a close afterwards) seems valid, as we will prevent issues in case we have a regression in the future. Is closing it here legit? What do you think?

Updates: #276

### DIFF
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -115,6 +115,7 @@ func (t *grpcTunnel) serve(c clientConn) {
 
 			if !ok {
 				klog.V(1).InfoS("DialResp not recognized; dropped", "connectionID", resp.ConnectID, "dialID", resp.Random)
+				return
 			} else {
 				result := dialResult{
 					err:    resp.Error,


konnectivity-client/pkg/client/client.go
```diff
@@ -115,12 +115,13 @@ func (t *grpcTunnel) serve(c clientConn) {
 
 			if !ok {
 				klog.V(1).Infoln("DialResp not recognized; dropped")
+				return
```
cheftako, line 118: 2 concerns but I think we should deal with them later.
1) Ideally we would send a PacketType_DIAL_CLS is c was a grpc channel. For that to happen c should be a *client.conn rather than a *grpc.ClientConn. (http-connect is a separate issue)
2) I would like to get away for SingleUseGRPCTunnel. I think this is wrong for that. However we can fix this change to work with that change when we are ready.

rata: Oh, will look into 1. I'm not sure if we close the connection here, if the proxy-server also needs to close some connection with the agent too. Should the DIAL_CLS solve that (if c was *client.conn)?

I'll have a look in the next days, but let me know if you already know the answer :)

cheftako: > Should the DIAL_CLS solve that (if c was *client.conn)?
Yes client.conn::Close will send either a CLOSE_REQ or DIAL_CLS depending on if it has received a DIAL_RSP. I think there is a bit of an issue in that we're crossing the streams. The 'grpc.clientConn' is the tunnel connection, in fact we later wrap it in a tunnel connection which makes that more clear. 'client.conn' is the request connection. So arguably what we should probably do is have tunnel Close() which walks the list of connections and calls Close on each of them.

cheftako:
Can we get this rebased?

iaguis:
Rebased :)

cheftako:
Thanks :D

rata:
@cheftako sorry, I was out sick. Thanks! And thanks @iaguis for covering me too :)

PR #293, opened 2021-11-15
avoid getting empty server id

mainred:
If `authenticateAgentViaToken` failed in server for client-side token expiration or being rejected by webhook in case of Kubernetes by accident, the client will print logs like this, which is too opaque to debug.
`"cannot sync once" err="expected one server ID in the context, got []"`

The error is caused by the fact that when server return from grpc call `Connect`, the stream is closed in client side and channel headerChan is closed from which metadata header will read, thus empty header is returned
https://github.com/grpc/grpc-go/blob/cf8b64e2c5bf11e00856a29794e434460eb67b90/internal/transport/transport.go#L331


pkg/server/server.go
```diff
@@ -634,6 +634,14 @@ func (s *ProxyServer) Connect(stream agent.AgentService_ConnectServer) error {
 	if s.AgentAuthenticationOptions.Enabled {
 		if err := s.authenticateAgentViaToken(stream.Context()); err != nil {
 			klog.ErrorS(err, "Client authentication failed", "agentID", agentID)
+			// Before sending metadata header, returning early will close stream from client side and give any empty value
+			// to client call Header(), which is a bit misleading as the client expect to receive server info to continue
+			// or error to break.
+			h := metadata.Pairs(header.AuthenticationError, err.Error())
```
mainred, line 640: This is the only error type I can think of for now before sending metadata, if we have more types of error in the future, we may think to use a more generalized error type.

pkg/agent/client.go
```diff
@@ -502,7 +511,7 @@ func (a *Client) remoteToProxy(connID int64, ctx *connContext) {
 			return
 		} else if err != nil {
 			// Normal when receive a CLOSE_REQ
-			klog.ErrorS(err, "connection read failure", "connectionID", connID)
+			klog.V(5).InfoS("connection read failure", "connectionID", connID, "err", err)
```
mainred, line 514: We have customers bring concerns about this error though the code has comments to show there is no harm.

pkg/agent/client.go
```diff
@@ -237,12 +238,17 @@ func (a *Client) Connect() (int, error) {
 		conn.Close() /* #nosec G104 */
 		return 0, err
 	}
-	serverID, err := serverID(stream)
+	md, err := getAndValidateStreamHeader(stream)
```
mainred, line 241: `conn.Close()` and returning 0 and error called in multiple times in this function. I tried to use the named return type and defer function to cover these cases but considering that makes the code less easy to be maintainable, I copied and pasted the error handling here.

mainred:
@cheftako , could we move forward with this PR? Thanks.

andyliuliming:
@mainred when will this PR be merged?

mainred:
Hi @andyliuliming, this change is waiting for review, hope the problem fixed in this PR does not confuse too many users.

pkg/agent/client.go
jkh52: Please move this to a separate PR, only because it's unrelated to the agent authentication fix.

pkg/server/server.go
jkh52: An alternative approach that would make it easier in the future to reason about backward compatibility: define a new Packet type in konnectivity-client/proto/client/client.proto.

But, it may be nice to keep Packet out of this Connect bootstrap phase and stay with headers only. We could move this definition into proto (enum?), to make the API slightly stronger.

@cheftako WDYT?

jkh52:
I'm overall LGTM on this PR. See two comments.

pkg/agent/client.go
mainred: Moved to https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/317, thanks.

pkg/agent/client.go
```diff
@@ -289,11 +295,19 @@ func (a *Client) Recv() (*client.Packet, error) {
 	return pkt, err
 }
 
-func serverCount(stream agent.AgentService_ConnectClient) (int, error) {
+func getAndValidateStreamHeader(stream agent.AgentService_ConnectClient) (metadata.MD, error) {
+	// TODO: this is a blocking call. Add a timeout?
 	md, err := stream.Header()
```
jkh52, line 300: For the reader / talking to myself: this change adds a new cleaner 403 signal from server to agent during bootstrapping. So it's natural to worry about the new agent binary / old server binary scenario, and the possibility that the agent will get stuck waiting for this new signal from the server. But, the old agent already blocked on .Header() in the same place as before (serverCount and serverID) so this alone isn't a risk.

jkh52:
@mainred friendly ping

mainred:
@jkh52 sorry for this pending PR. please let me know what else is required, thanks.

jkh52:
@mainred this PR needs rebase.

/assign @tallclair 
PTAL

PR #295, opened 2021-11-24
fix the proxy server backend metric error

### LINKED ISSUE #294: [BUG] The backend ServerMetric is not correct
As the konnectivity server may has multiple BackendManagers, the `backend` metric should record the total backends stored in various of BackendManager even if we can record them with a label. 

However, the `backend` metric is recorded by DefaultBackendStorage, which is the underlying storage implementation. It will result in only some BackendManager record  its number of backends.

The solution is that `backend` metric should be recorded by konnectivity server instead of DefaultBackendStorage.

YRXING:
#### What type of PR is this?
/kind bug

#### What this PR does / why we need it:
the backend metric should be recorded by proxy server
#### Which issue(s) this PR fixes:

Fixes #294


pkg/server/metrics/metrics.go
```diff
@@ -164,6 +167,16 @@ func (a *ServerMetrics) SetBackendCount(count int) {
 	a.backend.WithLabelValues().Set(float64(count))
 }
 
+// BackendCountInc increments a new backend connection.
+func (a *ServerMetrics) BackendCountInc(manager string, idType string) {
```
mainred, line 171: we have only one managed supported by the server once, so I guess it's not necessary to record metrics for each backend manager. the introduction of `idType` is valuable, but for current use cases, I'd suggest we keep the code as it is now.

mainred: multiple backend managers are supported for one proxy-sever, close this comment

YRXING: anything else I need to do about this issue?

pkg/server/server.go
```diff
@@ -200,14 +200,17 @@ func (s *ProxyServer) addBackend(agentID string, conn agent.AgentService_Connect
 			for _, ipv4 := range agentIdentifiers.IPv4 {
 				klog.V(5).InfoS("Add the agent to DestHostBackendManager", "agent address", ipv4)
 				s.BackendManagers[i].AddBackend(ipv4, pkgagent.IPv4, conn)
+				metrics.Metrics.BackendCountInc("DestHostBackendManager", "ipv4")
```
mainred, line 203: I think it's leaving this logic in `AddBackend`  is preferable, which makes adding backend logic more clear.
How about adding filed to backend storage to differentiate backend count for different kinds of backend managers.

YRXING: Please review the code.

pkg/server/default_route_backend_manager.go
```diff
@@ -32,20 +34,30 @@ var _ BackendManager = &DefaultRouteBackendManager{}
...
-			[]agent.IdentifierType{agent.DefaultRoute})}
+			[]pkgagent.IdentifierType{pkgagent.DefaultRoute})}
 }
 
 // Backend tries to get a backend associating to the request destination host.
-func (dibm *DefaultRouteBackendManager) Backend(ctx context.Context) (Backend, error) {
-	dibm.mu.RLock()
-	defer dibm.mu.RUnlock()
-	if len(dibm.backends) == 0 {
+func (drbm *DefaultRouteBackendManager) Backend(ctx context.Context) (Backend, error) {
```
mainred, line 41: :thumbsup:

pkg/server/backend_manager.go
```diff
@@ -204,7 +214,6 @@ func (s *DefaultBackendStorage) AddBackend(identifier string, idType pkgagent.Id
 		return addedBackend
 	}
 	s.backends[identifier] = []*backend{addedBackend}
-	metrics.Metrics.SetBackendCount(len(s.backends))
```
mainred, line 207: Thanks for your contribution, @YRXING .
We are complicating the logic of the backend count metrics. we don't have to rewrite `AddBackend` for each manager to collect backend count metrics. code maintenance and reusability should be taken into consideration.

How about we keep `SetBackendCount` here and add the manager name, idtype into it. manager name, can be defined as member of backendstore.

YRXING: Does the DefaultBackendStorage need record different idTypes count metric or just record the total backend count?

mainred: Following your thought already in this PR, record different backend counts from different idTypes and backend managers. `SetBackendCount` will finally look like this way after other changes are also included.
```suggestion
	metrics.Metrics.SetBackendCount(s.backendType, idType, len(s.backends))
````
cc @cheftako and @jkh52 who will give the final say to this PR.

Comment (YRXING) [reply pkg/server/backend_manager.go:207]: we can only set the total number of backends stored in this manager through `SetBackendCount(s.backendType, idType, len(s.backends))` method. Because the storage can not distinguish different types of backends. (unless we change the way it stores the backends)

```go
type DefaultBackendStorage struct {
    //...
    backends map[string][]*backend
    //...
}
```

If we want to record diffrent backends, we can use `BackendCountDec` and `BackendCountInc` method which I have defined in `metrics.go` file. It's just that the caller becomes DefaultBackendStorage.

Is that Ok?

jkh52:
/lgtm

I like depending on len() only, compared with Inc() / Dec().

pkg/server/metrics/metrics.go
```diff
@@ -99,7 +99,10 @@ func newServerMetrics() *ServerMetrics {
 			Name:      "ready_backend_connections",
 			Help:      "Number of konnectivity agent connected to the proxy server",
 		},
-		[]string{},
+		[]string{
+			"manager",
+			"idType",
+		},
```
jkh52, line 105: @logicalhan Is a change like this backward compatible, and if not is it more appropriate to create a new metric and deprecate the old?

logicalhan: This is definitely not backwards compatible. This isn't in k8s-proper, so the same rules of API conformance do not necessarily hold here. I would at the very least add a release note clearly denoting the API change to the metric.

logicalhan: Also what's the expected cardinality of manager and idType?

jkh52:
/remove-lgtm

I think we should keep this metric as-is, and add a new one. That way we avoid current integrations with metrics readers.

pkg/server/metrics/metrics.go
jkh52: I think cardinality is likely not a concern. (manager is currently 3, idType is about 3 or 4).

cheftako: > This is definitely not backwards compatible. This isn't in k8s-proper, so the same rules of API conformance do not necessarily hold here. I would at the very least add a release note clearly denoting the API change to the metric.

This is a binary we include in some distros of K8s (i.e. its an optional binary in K8s) so we should probably consider it part of K8s.

cheftako: > I think cardinality is likely not a concern. (manager is currently 3, idType is about 3 or 4).

Technically the idType is part of a CLI flag on the agent (Eg https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/addons/konnectivity-agent/konnectivity-agent-ds.yaml#L43), so theoretically its not bounded. 
Practically the number of managers and id types's are under the cloud provider's control. 
So I would expect this number to be small.

jkh52: See:
https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/cmd/agent/app/options/options.go#L178-L186

cheftako: Agreed. However that is on the agent side and the agent is at some level a reference implementation. 
If we had similar code on the server side to limit the cardinality of idType, then I think we could reasonably claim it was limited.

PR #299, opened 2021-12-02
Undo the most recent konnectivity-client/go.mod updates.

jkh52:
This change was making it hard to use 0.0.26 in k/k. See
https://github.com/kubernetes/kubernetes/pull/106634. After this commit, we can create tag 0.0.27 and retry.

The grpc version in particular is newer than k/k, and if we want the ability to cherry pick mitigations to older minor versions we need to be conservative. Here is a table of k/k branches and grpc version, for reference:

master: 1.40.0
release-1.23: 1.40.0
release-1.22: 1.38.0
release-1.21: 1.27.1
release-1.20: 1.27.1
release-1.19: 1.27.0


konnectivity-client/go.mod
```diff
@@ -3,8 +3,8 @@ module sigs.k8s.io/apiserver-network-proxy/konnectivity-client
 go 1.17
 
 require (
-	github.com/golang/protobuf v1.4.3
-	google.golang.org/grpc v1.42.0
+	github.com/golang/protobuf v1.4.1
+	google.golang.org/grpc v1.27.0
```
cheftako, line 7: https://github.com/kubernetes/kubernetes/blob/master/go.mod#L102 shows v1.40.0 . Would it make more sense to make that our version?

liggitt:
because we want to pick the memory leak fix back several versions, sticking to the grpc version before the latest bump will let us do that non-disruptively

konnectivity-client/go.mod
```diff
@@ -3,8 +3,8 @@ module sigs.k8s.io/apiserver-network-proxy/konnectivity-client
 go 1.17
 
 require (
-	github.com/golang/protobuf v1.4.3
-	google.golang.org/grpc v1.42.0
+	github.com/golang/protobuf v1.4.1
```
cheftako, line 6: https://github.com/kubernetes/kubernetes/blob/master/go.mod#L47 shows v1.5.2. Would it make sense to make that our version? (Ditto https://github.com/kubernetes/kubernetes/blob/release-1.23/go.mod)

konnectivity-client/go.mod
cheftako: (Ditto https://github.com/kubernetes/kubernetes/blob/release-1.23/go.mod#L102)

rata: Hmm, they pasted the grpc versions used in different kube versions we want to backport and from their message I understand that we need the lowest here so we can backport without issues.

Am I missing something?

rata, line 7: Can we add a comment explaining why we need this version and the same thing you pasted in the PR description with the grpc versions the different kube releases use?

rata (approved):
@jkh52 If this fixes the vendoring issues in kube, this LGTM. Thanks for putting me in cc and sorry for this regression!

This LGTM, just adding a comment so we don't make the same mistake will be great :)

konnectivity-client/go.mod
jkh52: There is tension between wanting newer konnectivity-client grpc (to reduce version skew between KAS and proxy-server), and keeping older konnectivity-client grpc (keeping the client backport friendly).

As a balance/compromise, updated to support backport to 1.21.

jkh52: Good idea I'll add a comment.

PR #300, opened 2021-12-03
Defer sending close response after having cleaned connection manager to

irozzo-1A:
Defer sending close response after having cleaned connection manager to avoid race conditions in `TestServeData_HTTP` unit test.

> --- FAIL: TestClose_Client (0.00s)
    client_test.go:338: client.connContext not released

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -415,6 +415,9 @@ func (a *Client) Serve() {
 					}
 					resp.GetCloseResponse().ConnectID = connID
 
+					close(dataCh)
+					a.connManager.Delete(connID)
+
 					err := conn.Close()
 					if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/agent/client.go
```diff
@@ -420,12 +420,12 @@ func (a *Client) Serve() {
 						resp.GetCloseResponse().Error = err.Error()
 					}
 
+					close(dataCh)
```
cheftako, line 423: I like the change but I think we can do this even more aggressively. If we close the channel and remove the connContext just before line 418, I think it stops us from sending data toward a connection we have already closed.

cheftako: P.S. I think both lines can move up to before ```err := conn.Close()```.

PR #301, opened 2021-12-06
Update cloudbuild.yaml to support docker buildx

Jefftree:
Not sure if it'll fix our build problem, following instructions from https://github.com/kubernetes/test-infra/tree/master/config/jobs/image-pushing#simple-build-example

/assign @cheftako

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -10,6 +10,9 @@ steps:
     - DOCKER_CLI_EXPERIMENTAL=enabled
     args:
     - release-staging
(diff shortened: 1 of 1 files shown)


cloudbuild.yaml
```diff
@@ -10,6 +10,9 @@ steps:
     - DOCKER_CLI_EXPERIMENTAL=enabled
     args:
     - release-staging
+    # default cloudbuild has HOME=/builder/home and docker buildx is in /root/.docker/cli-plugins/docker-buildx
```
cheftako, line 13: Would explicitly setting the path help?

PR #274, opened 2021-10-20
use docker buildx to build multi-arch image

mainred:
Before this change, the image built through our Makefile `docker-push/xxx` is with the same arch as the host as no arch simulation is introduced, even though ARCH is specified, as shown below I used `REGISTRY=qingchuanhao.azurecr.io ARCH=arm64 make docker-push/proxy-server` to push an arm64 image to ACR.

```
➜  apiserver-network-proxy git:(qinhao/docker-buildx) ✗ docker manifest inspect qingchuanhao.azurecr.io/proxy-server-arm64:4e8b12d2cd77db7ae5acecb5b706d5ecb3bf93e3 --verbose
{
        "Ref": "qingchuanhao.azurecr.io/proxy-server-arm64:4e8b12d2cd77db7ae5acecb5b706d5ecb3bf93e3",
        "Descriptor": {
                "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
                "digest": "sha256:1b7c6cda30a9107b6f3663f14aec7a8d8345db0edfb2d5a27df0be625120639a",
                "size": 529,
                "platform": {
                        "architecture": "amd64",
                        "os": "linux"
                }
        },
        "SchemaV2Manifest": {
                "schemaVersion": 2,
                "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
                "config": {
                        "mediaType": "application/vnd.docker.container.image.v1+json",
                        "size": 1626,
                        "digest": "sha256:6c8176319e0266e592a7c196fc59452c4f911264a28de871a6e7f6595b03bfaf"
                },
                "layers": [
                        {
                                "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
                                "size": 16844398,
                                "digest": "sha256:ac1c233a6102ecd2569ae1b62251b35c56aab32519e5cb6f85384230515c9374"
                        }
                ]
        }
}
```

After this change, the same command `REGISTRY=qingchuanhao.azurecr.io ARCH=arm64 make docker-push/proxy-server` return a arm64-based image
```
➜  apiserver-network-proxy git:(qinhao/docker-buildx) ✗ docker manifest inspect qingchuanhao.azurecr.io/proxy-server-arm64:4e8b12d2cd77db7ae5acecb5b706d5ecb3bf93e3 --verbose
{
        "Ref": "qingchuanhao.azurecr.io/proxy-server-arm64:4e8b12d2cd77db7ae5acecb5b706d5ecb3bf93e3",
        "Descriptor": {
[...]

PR #303, opened 2021-12-07
Increase cloudbuild timeout

Jefftree:
Increase cloudbuild timeout to 2.5 hours based on measurements provided [here](https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/274#issuecomment-970228867)

/assign @cheftako


cheftako:
Would be nice if we could work out how to make the build a bit shorter.

PR #304, opened 2021-12-09
Only publish image for amd64

Jefftree:
Only publish official image for amd64

This is a temporary measure to fix our broken image push pipeline.
/assign @cheftako

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,8 @@
 # limitations under the License.
 
 ARCH ?= amd64
-ALL_ARCH = amd64 arm arm64 ppc64le s390x
+ARCH_LIST ?= amd64 arm arm64 ppc64le s390x
+RELEASE_ARCH_LIST = amd64
 # The output type could either be docker (local), or registry.
 OUTPUT_TYPE ?= docker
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+DEFAULT_ARCH = amd64
```
jkh52, line 15: The difference between DEFAULT_ARCH and ARCH seems unclear.

As an alternative, just a new `RELEASE_LIST = amd64`?

Makefile
```diff
@@ -308,7 +309,7 @@ docker-push-manifest/http-test-server: ## Push the fat manifest docker image.
 
 .PHONY: release-staging
 release-staging: ## Builds and push container images to the staging bucket.
-	REGISTRY=$(STAGING_REGISTRY) $(MAKE) docker-push-all release-alias-tag
+	REGISTRY=$(STAGING_REGISTRY) ARCH_LIST=$(DEFAULT_ARCH) $(MAKE) docker-push-all release-alias-tag
```
jkh52, line 312: IIUC, should this be string quoted.

PR #305, opened 2021-12-09
Re-work of #252 - make dial request handling async

mihivagyok:
+ rework of https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/252
+ make dial request handling async in agent
+ ~~use infoS instead of infoln~~

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -394,15 +394,15 @@ func (a *Client) Serve() {
 			connID := atomic.AddInt64(&a.nextConnID, 1)
 			dataCh := make(chan []byte, xfrChannelSize)
 			dialDone := make(chan struct{})
-			ctx := &connContext{
+			connCtx := &connContext{
 				dataCh:    dataCh,
 				dialDone:  dialDone,
 				warnChLim: a.warnOnChannelLimit,
 			}
-			ctx.cleanFunc = func() {
+			connCtx.cleanFunc = func() {
 				// block on purpose
 				<-dialDone
-				if ctx.conn != nil {
+				if connCtx.conn != nil {
 					klog.V(4).InfoS("close connection", "connectionID", connID)
 					closeResp := &client.Packet{
 						Type:    client.PacketType_CLOSE_RSP,
@@ -410,7 +410,7 @@ func (a *Client) Serve() {
 					}
 
 					closeResp.GetCloseResponse().ConnectID = connID
-					err := ctx.conn.Close()
+					err := connCtx.conn.Close()
 					if err != nil {
 						closeResp.GetCloseResponse().Error = err.Error()
 					}
@@ -424,12 +424,10 @@ func (a *Client) Serve() {
 				}
 			}
 			go func() {
-				a.connManager.Add(connID, ctx)
 				defer close(dialDone)
 				start := time.Now()
 				conn, err := net.DialTimeout(dialReq.Protocol, dialReq.Address, dialTimeout)
 				if err != nil {
-					a.connManager.Delete(connID)
 					dialResp.GetDialResponse().Error = err.Error()
 					if err := a.Send(dialResp); err != nil {
 						klog.ErrorS(err, "could not send dialResp")
@@ -437,14 +435,15 @@ func (a *Client) Serve() {
 					return
 				}
 				metrics.Metrics.ObserveDialLatency(time.Since(start))
-				ctx.conn = conn
+				connCtx.conn = conn
+				a.connManager.Add(connID, connCtx)
 				dialResp.GetDialResponse().ConnectID = connID
 				if err := a.Send(dialResp); err != nil {
 					klog.ErrorS(err, "could not send dialResp")
 					return
 				}
-				go a.remoteToProxy(connID, ctx)
-				go a.proxyToRemote(connID, ctx)
+				go a.remoteToProxy(connID, connCtx)
+				go a.proxyToRemote(connID, connCtx)
 			}()
 
 		case client.PacketType_DATA:
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -49,6 +49,7 @@ type connContext struct {
 	dataCh    chan []byte
 	cleanOnce sync.Once
 	warnChLim bool
+	dialDone  chan struct{}
 }
 
 func (c *connContext) cleanup() {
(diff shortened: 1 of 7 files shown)


pkg/agent/client.go
```diff
@@ -375,70 +376,73 @@ func (a *Client) Serve() {
...
-
-					if err := a.Send(resp); err != nil {
+					if err := a.Send(closeResp); err != nil {
 						klog.ErrorS(err, "close response failure")
 					}
-				},
-				warnChLim: a.warnOnChannelLimit,
+					close(dataCh)
+					a.connManager.Delete(connID)
+				}
```
jkh52, line 421: please add an else case with log, to help debugging possible leaks in the future.

pkg/agent/client.go
```diff
@@ -375,70 +376,73 @@ func (a *Client) Serve() {
...
 			}
 			a.connManager.Add(connID, ctx)
```
jkh52, line 423: I think a.connManager.Add should be moved to the go func(), so that we preserve old behavior (available before dial response is sent)

jkh52:
This looks like a good improvement for reliability, thanks!

pkg/agent/client.go
mihivagyok: Thanks! I have added.

pkg/agent/client.go
mihivagyok: Right. I hope I moved to the right place :) Thank you!

mihivagyok:
@jkh52 @caesarxuchao Could you please take a look at this PR? If I need to provide any more details, please let me know. Thank you! Adam

pkg/agent/client.go
```diff
@@ -375,70 +376,75 @@ func (a *Client) Serve() {
...
+					closeResp := &client.Packet{
 						Type:    client.PacketType_CLOSE_RSP,
 						Payload: &client.Packet_CloseResponse{CloseResponse: &client.CloseResponse{}},
 					}
-					resp.GetCloseResponse().ConnectID = connID
-
-					close(dataCh)
-					a.connManager.Delete(connID)
 
-					err := conn.Close()
```
caesarxuchao, line 421: Who's responsible to close the `conn` now?

pkg/agent/client.go
```diff
@@ -375,70 +376,75 @@ func (a *Client) Serve() {
...
+					closeResp.GetCloseResponse().ConnectID = connID
 					if err != nil {
-						resp.GetCloseResponse().Error = err.Error()
+						closeResp.GetCloseResponse().Error = err.Error()
```
caesarxuchao, line 414: `err` used to be tracking the error of `conn.Close()`. What is it tracking now?

pkg/agent/client.go
```diff
@@ -375,70 +376,75 @@ func (a *Client) Serve() {
...
+			go func() {
+				a.connManager.Add(connID, ctx)
+				defer close(dialDone)
+				start := time.Now()
+				conn, err := net.DialTimeout(dialReq.Protocol, dialReq.Address, dialTimeout)
+				if err != nil {
+					a.connManager.Delete(connID)
+					dialResp.GetDialResponse().Error = err.Error()
+					if err := a.Send(dialResp); err != nil {
+						klog.ErrorS(err, "could not send stream")
```
caesarxuchao, line 434: nit: could not send dialResp

pkg/agent/client.go
```diff
@@ -375,70 +376,75 @@ func (a *Client) Serve() {
...
+					}
+					return
+				}
+				metrics.Metrics.ObserveDialLatency(time.Since(start))
+				ctx.conn = conn
+				dialResp.GetDialResponse().ConnectID = connID
+				if err := a.Send(dialResp); err != nil {
+					klog.ErrorS(err, "stream send failure")
```
caesarxuchao, line 442: nit: could not send dialResp

caesarxuchao (commented):
Can you move the changes of Infoln -> InfoS to another PR?

Sorry my review availability will be on and off. Please ping @jkh52 / @cheftako / @Jefftree if you need more reviews.

pkg/agent/client.go
mihivagyok: Orignally this was the log, but sure, I could change it to be more clear about the error! Thank you!

pkg/agent/client.go
mihivagyok: You are so right. The ctx.conn.Close() was missing. Thank you!

pkg/agent/client.go
mihivagyok: You are so right. The ctx.conn.Close() was missing. Thank you!

mihivagyok:
Hi team! It is ready for review: @jkh52 @cheftako @Jefftree
 Thank you!

pkg/agent/client.go
```diff
@@ -375,70 +376,76 @@ func (a *Client) Serve() {
...
+					a.connManager.Delete(connID)
+				} else {
+					klog.ErrorS(fmt.Errorf("connection is nil"), "cannot send CLOSE_RESP to nil connection")
+				}
 			}
-
-			go a.remoteToProxy(connID, ctx)
-			go a.proxyToRemote(connID, ctx)
+			go func() {
+				a.connManager.Add(connID, ctx)
```
cheftako, line 427: I believe adding the ctx (connContext) to the connManager this early opens us up to race conditions. 

Once the ctx.conn has been set I think it should be safe to add the ctx to the connManager.

pkg/agent/client.go
```diff
@@ -375,70 +376,76 @@ func (a *Client) Serve() {
...
-				}
-				continue
-			}
-			metrics.Metrics.ObserveDialLatency(time.Since(start))
+			dialResp.GetDialResponse().Random = dialReq.Random
 
 			connID := atomic.AddInt64(&a.nextConnID, 1)
 			dataCh := make(chan []byte, xfrChannelSize)
+			dialDone := make(chan struct{})
 			ctx := &connContext{
```
cheftako, line 397: While we're cleaning up names to avoid confusion, can we rename ctx to something like connCtx?

pkg/agent/client.go
mihivagyok: Thanks! Good idea.

pkg/agent/client.go
mihivagyok: Of course, it sounds good! Thanks!

PR #307, opened 2021-12-10
fix: client should return instantly when no backend

mainred:
Two changes are included in this PR:
- If no backend is registered to the server, and the grpc client request will hang for 30s instead of returning instantly with `no backend` error. This is caused by the issue that random id is not included in dia resp from server to client, thus the client does not recognize the dial resp and return, dropping the error.
- for a normal client-server close case, close req is sent to backend twice, one as proxy from client, and the other one is as a generalized one. The latter one is not necessary, and bring misleading error message for both proxy-server and proxy-agent

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -425,8 +425,12 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer, 
 				klog.ErrorS(err, "Failed to get a backend", "serverID", s.serverID)
 
 				resp := &client.Packet{
-					Type:    client.PacketType_DIAL_RSP,
-					Payload: &client.Packet_DialResponse{DialResponse: &client.DialResponse{Error: err.Error()}},
+					Type: client.PacketType_DIAL_RSP,
+					Payload: &client.Packet_DialResponse{
+						DialResponse: &client.DialResponse{
+							Random: pkt.GetDialRequest().Random,
+							Error:  err.Error(),
+						}},
 				}
 				if err := stream.Send(resp); err != nil {
 					klog.V(5).Infoln("Failed to send DIAL_RSP for no backend", "error", err, "serverID", s.serverID)
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -459,8 +459,10 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer, 
 			if err := backend.Send(pkt); err != nil {
 				// TODO: retry with other backends connecting to this agent.
 				klog.ErrorS(err, "CLOSE_REQ to Backend failed", "serverID", s.serverID, "connectionID", connID)
+			} else {
+				klog.V(5).Infoln("CLOSE_REQ sent to backend", "serverID", s.serverID, "connectionID", connID)
+				return
 			}
-			klog.V(5).Infoln("CLOSE_REQ sent to backend", "serverID", s.serverID, "connectionID", connID)
 
 		case client.PacketType_DIAL_CLS:
 			random := pkt.GetCloseDial().Random
--- a/pkg/server/server_test.go
+++ b/pkg/server/server_test.go
@@ -253,7 +253,7 @@ func prepareAgentConnMD(ctrl *gomock.Controller, proxyServer *ProxyServer) *agen
 	return agentConn
 }
 
-func baseTestWithoutBackend(t *testing.T, validate func(*agentmock.MockAgentService_ConnectServer)) {
+func baseServerProxyTestWithoutBackend(t *testing.T, validate func(*agentmock.MockAgentService_ConnectServer)) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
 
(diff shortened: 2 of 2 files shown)


pkg/server/server_test.go
```diff
@@ -218,3 +222,72 @@ func TestAddRemoveFrontends(t *testing.T) {
 		t.Errorf("expected %v, got %v", e, a)
 	}
 }
+
+func TestServerProxyNormalClose(t *testing.T) {
```
mainred, line 226: Before we have other tests sharing these definitions, let's first start in this way.
I am planning to add more UT with gomock to protect further behaviors and make it simpler for others contributors to write UT for their changes.

jkh52: This test coverage looks good. Would be nice to also add coverage of the DialResponse "with error" case that you fixed above.

pkg/server/server.go
```diff
@@ -425,8 +425,12 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
 				klog.ErrorS(err, "Failed to get a backend", "serverID", s.serverID)
 
 				resp := &client.Packet{
-					Type:    client.PacketType_DIAL_RSP,
-					Payload: &client.Packet_DialResponse{DialResponse: &client.DialResponse{Error: err.Error()}},
+					Type: client.PacketType_DIAL_RSP,
+					Payload: &client.Packet_DialResponse{
+						DialResponse: &client.DialResponse{
+							Random: pkt.GetDialRequest().Random,
```
jkh52, line 431: +1

Nice fix, thank you!

pkg/server/server.go
```diff
@@ -459,8 +463,10 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
 			if err := backend.Send(pkt); err != nil {
 				// TODO: retry with other backends connecting to this agent.
 				klog.ErrorS(err, "CLOSE_REQ to Backend failed", "serverID", s.serverID, "connectionID", connID)
+			} else {
+				klog.V(5).Infoln("CLOSE_REQ sent to backend", "serverID", s.serverID, "connectionID", connID)
+				return
```
jkh52, line 468: Please clarify why this is needed here, but not also in other "close" codepaths?

Also- I know we currently have single use grpc tunnel, but this change may make it harder in the future to have a multi-use tunnel.

mainred: 1. after sending close req to the backend, the work for `serveRecvFrontend` is done, right?
2. without this change, we'll send a second close req to the backend at the end of `serveRecvFrontend`. The latter one is caused by closing `recvCh` for closed connection.
Here are some log proof:
### proxy-server:
- The first time to send close req to backend

I1211 12:56:53.137935    6025 server.go:472] CLOSE_REQ sent to backend serverID 8fb992ef-fb34-475d-9fa6-de569516b226 connectionID 1
I1211 12:56:53.138643    6025 server.go:770] "Received CLOSE_RSP" serverID="8fb992ef-fb34-475d-9fa6-de569516b226" agentID="5b03efa7-411f-4a2b-9ea3-088b7235daad" connectionID=1
I1211 12:56:53.138691    6025 server.go:780] CLOSE_RSP sent to frontend connectionID 1

- skip some log
- the second time to send close rep to second, we have no log but you can tell from CLOSE_RSP

I1211 12:56:53.139887    6025 server.go:770] "Received CLOSE_RSP" serverID="8fb992ef-fb34-475d-9fa6-de569516b226" agentID="5b03efa7-411f-4a2b-9ea3-088b7235daad" connectionID=1
E1211 12:56:53.139942    6025 server.go:773] "could not get frontend client" err="can't find agentID 5b03efa7-411f-4a2b-9ea3-088b7235daad in the frontends" serverID="8fb992ef-fb34-475d-9fa6-de569516b226" agentID="5b03efa7-411f-4a2b-9ea3-088b7235daad" connectionID=1

### proxy-agent

- the first time to receive close req

I1211 12:56:53.138121    6012 client.go:456] "received CLOSE_REQ" connectionID=1
I1211 12:56:53.138161    6012 client.go:411] "close connection" connectionID=1
I1211 12:56:53.138281    6012 client.go:498] "received data from remote" bytes=0 connectionID=1
E1211 12:56:53.138312    6012 client.go:505] "connection read failure" err="read tcp 172.28.0.2:60474->172.28.0.2:8000: use of closed network connection" connectionID=1
I1211 12:56:53.138425    6012 client.go:486] "Exiting remoteToProxy" connectionID=1
I1211 12:56:53.138445    6012 client.go:524] "Exiting proxyToRemote" connectionID=1

- the second time to receive close req

I1211 12:56:53.139584    6012 client.go:375] "[tracing] recv packet" type="CLOSE_REQ"
I1211 12:56:53.139608    6012 client.go:456] "received CLOSE_REQ" connectionID=1
I1211 12:56:53.139643    6012 client.go:462] "Failed to find connection context for close" connectionID=1
E1211 13:00:53.143622    6012 client.go:371] "could not read stream" err="rpc error: code = Unavailable desc = closing transport due to: connection error: desc = \"error reading from server: EOF\", received prior goaway: code: ENHANCE_YOUR_CALM, debug data: \"too_many_pings\""

mainred: Fortunately, this change is not a must, considering the server and agent can handle this well, but reducing these noisy error log from konnectivity will make customers feel safe.

pkg/server/server_test.go
mainred: Sure, of course

andrewsykim:
Sorry for the delay getting to this PR, I will spend some time this week testing this patch

pkg/server/server.go
andrewsykim, line 468: @mainred I think it might be simpler to scope this PR to only include DialResponse containing random IDs and exclude the fix for sending CLOSE_REQ twice in a separate PR.  @mihivagyok also has a PR out regarding this https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/322

mainred:
Close this PR since the issue has been address by https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/356

Issue #306, opened 2021-12-09
Do images for proxy-test-client and http-server need to be pushed?


Jefftree:
Our release pipeline currently pushes images for the proxy-test-client and http-server in additional to the required proxy-server and proxy-agent binaries. We're using the proxy-test-client and http-server strictly for local testing currently, do they need to be pushed to the container registry?

cheftako:
I don't believe so. proxy-test-client and http-server exist specifically to allow us to do testing. So unless our test suites are pulling them in from a reqistry, they should not need to be released.

mainred:
Agree, and I think we don't have customers using proxy-test-client and http-server, or using konnectivity without kubernetes.

yue9944882:
for those who looking for a prebuilt image of {proxy-server,proxy-agent,proxy-test-server/client}, as a workaround we can use `quay.io/open-cluster-management/cluster-proxy:latest` of which the latest release is built over `v0.0.27` of this project.

PR #308, opened 2021-12-10
Avoid pushing docker images for proxy-test-client and http-test-server

Jefftree:
Remove the code for building and pushing the docker images for proxy-test-client and http-test-server since they are used only for testing and development purposes.

The images can still be built locally via `make build` (`make bin/proxy-test-client` and `make bin/proxy-test-server`)

I also removed the docker build step for the images in this PR. Are there any use cases of using docker build without docker push?

Fixes https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/306

/cc @cheftako 
/cc @mainred 
/cc @jkh52

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -179,10 +179,10 @@ buildx-setup:
 	${DOCKER_CMD} buildx inspect img-builder > /dev/null || docker buildx create --name img-builder --use
 
 .PHONY: docker-build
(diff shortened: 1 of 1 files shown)


mainred (approved):
Thanks. And our check-in will be a bit faster.

PR #310, opened 2021-12-13
Decouple read remote from write to server.

cheftako:
Currently we read from the remote and on the same routine, we write to
the konn server. This means that reading from subsequent reads from
remote are effectively blocked on the write to the konn server. This PR
decouples the read and write. It combines all the writes on to a
channel. Then the writes are on their own routine which reads from that
channel.

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -172,6 +172,9 @@ type Client struct {
 
 	cs *ClientSet // the clientset that includes this AgentClient.
 
+	serverChannel    chan *client.Packet
+	cleanChannel     sync.Once
+	serverError      error // Currently checking and setting of error all occur in writeToProxyServer
 	stream           agent.AgentService_ConnectClient
 	agentID          string
 	agentIdentifiers string
@@ -250,9 +253,33 @@ func (a *Client) Connect() (int, error) {
 	}
 	a.conn = conn
 	a.stream = stream
+	a.serverChannel = make(chan *client.Packet, xfrChannelSize)
+	a.serverError = nil
 	a.serverID = serverID
+	go a.writeToProxyServer()
 	klog.V(2).InfoS("Connect to", "server", serverID)
 	return serverCount, nil
+}
+
+func (a *Client) writeToProxyServer() {
+	defer func() {
+		if panicInfo := recover(); panicInfo != nil {
+			klog.V(2).InfoS("Exiting writeToProxyServer with recovery", "panicInfo", panicInfo, "serverID", a.serverID)
+		} else {
+			klog.V(2).InfoS("Exiting writeToProxyServer", "serverID", a.serverID)
+		}
+	}()
+
+	for pkt := range a.serverChannel {
+		klog.V(5).InfoS("writeToProxyServer recevied packet to send to KonnServer", "serverID", a.serverID)
+		err := a.stream.Send(pkt)
+		if err != nil && err != io.EOF {
+			metrics.Metrics.ObserveFailure(metrics.DirectionToServer)
+			a.cs.RemoveClient(a.serverID)
+			klog.V(2).InfoS("writeToProxyServer removing server from clientset", "serverID", a.serverID)
+			a.serverError = err
+		}
+	}
 }
 
 // Close closes the underlying connection.
@@ -271,12 +298,23 @@ func (a *Client) Send(pkt *client.Packet) error {
 	a.sendLock.Lock()
 	defer a.sendLock.Unlock()
 
-	err := a.stream.Send(pkt)
-	if err != nil && err != io.EOF {
-		metrics.Metrics.ObserveFailure(metrics.DirectionToServer)
-		a.cs.RemoveClient(a.serverID)
+	if a.serverError != nil {
+		// We know the return connection is closed, so return the error.
+		// Failing to return an error here does not mean the send will succeed.
+		// It just means we do not know yet if it will fail.
+		// Slight back-flips here to ensure the write is closing the channel.
+		a.cleanChannel.Do(func() {
+			klog.V(2).InfoS("Data channel to server has errored out", "serverID", a.serverID)
+			close(a.serverChannel)
+		})
+		return a.serverError
 	}
-	return err
+
+	if a.warnOnChannelLimit && len(a.serverChannel) >= xfrChannelSize {
+		klog.V(2).InfoS("Data channel to server on agent is full", "serverID", a.serverID)
+	}
+	a.serverChannel <- pkt
+	return nil
 }
 
 func (a *Client) Recv() (*client.Packet, error) {
@@ -495,9 +533,6 @@ func (a *Client) remoteToProxy(connID int64, ctx *connContext) {
 	defer ctx.cleanup()
 
 	var buf [1 << 12]byte
-	resp := &client.Packet{
-		Type: client.PacketType_DATA,
-	}
 
 	for {
 		n, err := ctx.conn.Read(buf[:])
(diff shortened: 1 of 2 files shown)


pkg/agent/client.go
```diff
@@ -505,10 +544,16 @@ func (a *Client) remoteToProxy(connID int64, ctx *connContext) {
 			klog.ErrorS(err, "connection read failure", "connectionID", connID)
 			return
 		} else {
-			resp.Payload = &client.Packet_Data{Data: &client.Data{
-				Data:      buf[:n],
-				ConnectID: connID,
-			}}
+			data := make([]byte, 0, n)
```
mainred, line 547: This decoupling is based on the fact we can store/cache the data received from the remote before sending it to the server. 
If the connection to the server is unhealthy or closed, maybe it's in some sense a good idea to block receiving more data from remote until the healthy connection to the server is back, because the cached data won't be sent to the server for connection is closed otherwise.

mainred (commented):
Thanks for your work, left a few comments.

pkg/agent/client.go
```diff
@@ -270,12 +296,28 @@ func (a *Client) Send(pkt *client.Packet) error {
 	a.sendLock.Lock()
 	defer a.sendLock.Unlock()
 
-	err := a.stream.Send(pkt)
+	/*err := a.stream.Send(pkt)
```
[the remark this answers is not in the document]
cheftako: Yes. (My bad for not removing prior to sending this out)

pkg/agent/client.go
```diff
@@ -249,11 +252,34 @@ func (a *Client) Connect() (int, error) {
...
+		}
+	}()
+
+	for pkt := range a.serverChannel {
+		klog.V(5).InfoS("writeToKonnServer recevied packet to send to KonnServer", "serverID", a.serverID)
+		err := a.stream.Send(pkt)
+		if err != nil && err != io.EOF {
+			metrics.Metrics.ObserveFailure(metrics.DirectionToServer)
+			a.cs.RemoveClient(a.serverID)
+			a.serverError = err
```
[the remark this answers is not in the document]
cheftako: I'm making an assumption that *this* channel/connection to the server is no longer functional. By removing the serverID from the ClientSet we are enabling https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/b9bf6c51031b435180466a6ccf2c7e9a531f12c6/pkg/agent/clientset.go#L176 to now establish a new channel to the server. Anyone using the old channel/connection will get an error. All new requests should use the new channel/connection once it has been established. This does mean in flight requests will fail. However this is supposed to be an exception occurence.

pkg/agent/client.go
cheftako: That's a really nice idea for an improvement. However it would require we implement a resume all the way through. It involved being able to move in flight connections from 1 channel to newer channel. That means we need to coordinate that in both agent and server. While I really like this idea I think its better done in a subsequent change.

pkg/agent/client.go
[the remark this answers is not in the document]
mainred: Thanks for your explanation, do you mind adding a log when removing server ID happens, like around https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/310/files#diff-c2eff45b2c7db0ae864aecb2e8b733dfbdfe7070ffe6f7feb1204e086c403a0aR277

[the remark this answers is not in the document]
mainred: I can add a UT to protect the UT which removes and adds a server back later.

mainred (approved):

pkg/agent/client.go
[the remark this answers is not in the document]
cheftako: Done

pkg/agent/client.go
```diff
@@ -249,11 +252,35 @@ func (a *Client) Connect() (int, error) {
...
+	}()
+
+	for pkt := range a.serverChannel {
+		klog.V(5).InfoS("writeToKonnServer recevied packet to send to KonnServer", "serverID", a.serverID)
+		err := a.stream.Send(pkt)
+		if err != nil && err != io.EOF {
+			metrics.Metrics.ObserveFailure(metrics.DirectionToServer)
+			a.cs.RemoveClient(a.serverID)
+			klog.V(2).InfoS("WriteToKonnServer removing server from clientset", "serverID", a.serverID)
+			a.serverError = err
```
andrewsykim, line 279: Do we need to lock when setting a.serverError?

pkg/agent/client.go
```diff
@@ -249,11 +252,35 @@ func (a *Client) Connect() (int, error) {
 	}
 	a.conn = conn
 	a.stream = stream
+	a.serverChannel = make(chan *client.Packet, xfrChannelSize)
+	a.serverError = nil
```
jkh52, line 256: serverError could deserve some comments about: why it doesn't need lock protection, and when set the client is effectively dead.

jkh52:
/hold

(to allow address my single comment)

/lgtm

pkg/agent/client.go
cheftako: No. Checking the value of serverError and the setting of serverError to err, all occur on the same goroutine. Adding a comment to that effect.

pkg/agent/client.go
cheftako: Done.

pkg/agent/client.go
```diff
@@ -249,11 +252,35 @@ func (a *Client) Connect() (int, error) {
...
 	a.serverID = serverID
+	go a.writeToKonnServer()
 	klog.V(2).InfoS("Connect to", "server", serverID)
 	return serverCount, nil
 }
 
+func (a *Client) writeToKonnServer() {
```
mihivagyok, line 263: I think it is already hard to understand the naming in the code (proxy, server, agent, client, frontend, backend, remote etc), that's why I think the `writeToKonnServer` function name is not the best - also so far this Konn name is not appearing in the code. That's why I would like to suggest some other name like `writeToProxyServer` or something like that. What do you think? Thanks! Adam

cheftako: Done

pkg/agent/client.go
```diff
@@ -271,12 +298,23 @@ func (a *Client) Send(pkt *client.Packet) error {
...
-	if err != nil && err != io.EOF {
-		metrics.Metrics.ObserveFailure(metrics.DirectionToServer)
-		a.cs.RemoveClient(a.serverID)
+	if a.serverError != nil {
+		// We know the return connection is closed, so return the error.
+		// Failing to return an error here does not mean the send will succeed.
+		// It just means we do not know yet if it will fail.
+		// Slight back-flips here to ensure the write is closing the channel.
+		a.cleanChannel.Do(func() {
+			klog.V(2).InfoS("Data channel to server has errored out", "serverID", a.serverID)
```
andrewsykim, line 307: Should this be klog.Error including a.serverError as one of the keys?

pkg/agent/client.go
```diff
@@ -250,11 +253,35 @@ func (a *Client) Connect() (int, error) {
...
 	klog.V(2).InfoS("Connect to", "server", serverID)
 	return serverCount, nil
 }
 
+func (a *Client) writeToProxyServer() {
+	defer func() {
+		if panicInfo := recover(); panicInfo != nil {
+			klog.V(2).InfoS("Exiting writeToProxyServer with recovery", "panicInfo", panicInfo, "serverID", a.serverID)
+		} else {
+			klog.V(2).InfoS("Exiting writeToProxyServer", "serverID", a.serverID)
```
andrewsykim, line 269: v=2 seems a little low for this doesn't it? I would expect this to be v=4 at least

pkg/agent/client.go
```diff
@@ -250,11 +253,35 @@ func (a *Client) Connect() (int, error) {
...
+		}
+	}()
+
+	for pkt := range a.serverChannel {
+		klog.V(5).InfoS("writeToProxyServer recevied packet to send to KonnServer", "serverID", a.serverID)
+		err := a.stream.Send(pkt)
+		if err != nil && err != io.EOF {
```
andrewsykim, line 276: Should we be cleaning up this goroutine if err == io.EOF?

Issue #311, opened 2021-12-15
Idle connections are stuck after establishing connection. They never send PacketType_DATA


ipochi:
While exploring more on issue #276, it was noticed that the idle connections are somehow stuck in a weird state where the connection is established i.e the kube-apiserver receives a DIAL_RSP from the proxy-server, but then thereafter there is complete inactivity  from the apiserver i.e no packets of PacketType_DATA are sent.

In order to reproduce this, I added some debug statements to the `konnectivity-client` library which is vendored in Kubernetes and built a local cluster.

This is the diff in the kubernetes repo to build a local cluster
```
diff --git a/hack/local-up-cluster.sh b/hack/local-up-cluster.sh
index abe85bb29fc..49b526ef080 100755
--- a/hack/local-up-cluster.sh
+++ b/hack/local-up-cluster.sh
@@ -553,7 +553,20 @@ kind: EgressSelectorConfiguration
 egressSelections:
 - name: cluster
   connection:
-    proxyProtocol: Direct
+    # This controls the protocol between the API Server and the Konnectivity
+    # server. Supported values are "GRPC" and "HTTPConnect". There is no
+    # end user visible difference between the two modes. You need to set the
+    # Konnectivity server to work in the same mode.
+    proxyProtocol: GRPC
+    transport:
+      # This controls what transport the API Server uses to communicate with the
+      # Konnectivity server. UDS is recommended if the Konnectivity server
+      # locates on the same machine as the API Server. You need to configure the
+      # Konnectivity server to listen on the same UDS socket.
+      # The other supported transport is "tcp". You will need to set up TLS
+      # config to secure the TCP transport.
+      uds:
+        udsName: /tmp/uds-proxy
 - name: controlplane
   connection:
     proxyProtocol: Direct
@@ -617,6 +630,7 @@ EOF
       --requestheader-allowed-names=system:auth-proxy \
       --proxy-client-cert-file="${CERT_DIR}/client-auth-proxy.crt" \
       --proxy-client-key-file="${CERT_DIR}/client-auth-proxy.key" \
+      --api-audiences=https://kubernetes.default.svc,system:konnectivity-server \
       --cors-allowed-origins="${API_CORS_ALLOWED_ORIGINS}" >"${APISERVER_LOG}" 2>&1 &
     APISERVER_PID=$!

[...]

PR #312, opened 2021-12-15
(WIP - Do not review) Terminate Idle connections by setting a timeout to free up resources

### LINKED ISSUE #276: Leaking/Too many open File descriptors in Konnectivity server with GRPC mode and UDS transport

ipochi:
This PR fixes #276 by addressing the issue #311.

We introduce a new flag `--grpc-max-idle-time` after which any idle connections from frontend(kube-apiserver) to proxy-server will be terminated.

To test this PR: reproduce the steps:
1. checkout this PR/branch locally and build the proxy-server binary
2. Follow steps described in the issue #311.

Start the proxy-server with:
```
./bin/proxy-server --server-port=0 --uds-name=/tmp/uds-proxy --cluster-ca-cert=certs/agent/issued/ca.crt --cluster-cert=certs/agent/issued/proxy-frontend.crt --cluster-key=certs/agent/private/proxy-frontend.key --warn-on-channel-limit --v 5 --keepalive-time 1m --frontend-keepalive-time 1m  --enable-profiling --grpc-max-idle-time 3m 2>&1 | tee server.log
```

Before running the stress test, get the following metrics
```
$ curl -s localhost:8095/metrics | grep -E "grpc"
# HELP konnectivity_network_proxy_server_grpc_connections Number of current grpc connections, partitioned by service method.
# TYPE konnectivity_network_proxy_server_grpc_connections gauge
konnectivity_network_proxy_server_grpc_connections{service_method="Connect"} 1
konnectivity_network_proxy_server_grpc_connections{service_method="Proxy"} 0


$ curl -s localhost:8095/metrics | grep -E "open_fds"
# HELP process_open_fds Number of open file descriptors.
# TYPE process_open_fds gauge
process_open_fds 13
```

Take a note of the metrics, during the stress test:

Once the stress test is complete and enough time has passed to trigger a `--grpc-max-idle-time` timeout, you'll notice that the metrics values would be pre-stress test levels.

/cc @rata @iaguis

### DIFF
--- a/cmd/server/app/options/options.go
+++ b/cmd/server/app/options/options.go
@@ -41,6 +41,9 @@ type ProxyRunOptions struct {
 	// pings the client to see if the transport is still alive.
 	KeepaliveTime         time.Duration
 	FrontendKeepaliveTime time.Duration
+	// Time that a gRPC stream between frontend and the proxy-server can be inactive (no pkt received) before
+	// it is closed.
+	GrpcMaxIdleTime time.Duration
 	// Enables pprof at host:AdminPort/debug/pprof.
 	EnableProfiling bool
 	// If EnableProfiling is true, this enables the lock contention
@@ -117,6 +120,7 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.ProxyStrategies, "proxy-strategies", o.ProxyStrategies, "The list of proxy strategies used by the server to pick a backend/tunnel, available strategies are: default, destHost.")
 	flags.BoolVar(&o.WarnOnChannelLimit, "warn-on-channel-limit", o.WarnOnChannelLimit, "Turns on a warning if the system is going to push to a full channel. The check involves an unsafe read.")
 	flags.StringVar(&o.CipherSuites, "cipher-suites", o.CipherSuites, "The comma separated list of allowed cipher suites. Has no effect on TLS1.3. Empty means allow default list.")
+	flags.DurationVar(&o.GrpcMaxIdleTime, "grpc-max-idle-time", o.GrpcMaxIdleTime, "Time that a gRPC connection can be inactive (no pkt received) before it is closed.")
 	return flags
 }
 
@@ -136,6 +140,7 @@ func (o *ProxyRunOptions) Print() {
 	klog.V(1).Infof("Health port set to %d.\n", o.HealthPort)
 	klog.V(1).Infof("Keepalive time set to %v.\n", o.KeepaliveTime)
 	klog.V(1).Infof("Frontend keepalive time set to %v.\n", o.FrontendKeepaliveTime)
+	klog.V(1).Infof("gRPC maxIdleTime time set to %v.\n", o.GrpcMaxIdleTime)
 	klog.V(1).Infof("EnableProfiling set to %v.\n", o.EnableProfiling)
 	klog.V(1).Infof("EnableContentionProfiling set to %v.\n", o.EnableContentionProfiling)
 	klog.V(1).Infof("ServerID set to %s.\n", o.ServerID)
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -145,6 +145,9 @@ type ProxyServer struct {
 	AgentAuthenticationOptions *AgentTokenAuthenticationOptions
 
 	proxyStrategies []ProxyStrategy
+
+	// Klls inactive (no app payload) Proxy connections after this duration.
+	grpcMaxIdleTime time.Duration
 }
 
 // AgentTokenAuthenticationOptions contains list of parameters required for agent token based authentication
@@ -326,7 +329,7 @@ func (s *ProxyServer) getFrontendsForBackendConn(agentID string, backend Backend
 }
 
 // NewProxyServer creates a new ProxyServer instance
-func NewProxyServer(serverID string, proxyStrategies []ProxyStrategy, serverCount int, agentAuthenticationOptions *AgentTokenAuthenticationOptions, warnOnChannelLimit bool) *ProxyServer {
+func NewProxyServer(serverID string, proxyStrategies []ProxyStrategy, serverCount int, agentAuthenticationOptions *AgentTokenAuthenticationOptions, warnOnChannelLimit bool, grpcMaxIdleTime time.Duration) *ProxyServer {
 	var bms []BackendManager
 	for _, ps := range proxyStrategies {
 		switch ps {
@@ -352,6 +355,7 @@ func NewProxyServer(serverID string, proxyStrategies []ProxyStrategy, serverCoun
 		Readiness:          bms[0],
 		proxyStrategies:    proxyStrategies,
 		warnOnChannelLimit: warnOnChannelLimit,
+		grpcMaxIdleTime:    grpcMaxIdleTime,
 	}
 }
 
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -24,6 +24,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"google.golang.org/grpc/metadata"
@@ -53,6 +54,8 @@ type ProxyClientConnection struct {
 	agentID   string
 	start     time.Time
 	backend   Backend
+	streamID  int64
+	idleTimer *time.Timer
 }
 
 const (
@@ -148,6 +151,18 @@ type ProxyServer struct {
 
 	// Klls inactive (no app payload) Proxy connections after this duration.
 	grpcMaxIdleTime time.Duration
+
+	// Tracks the stream ID.
+	nextStreamID int64
+
+	// Map to track the connections per stream ID.
+	connectionIDsByStreamID map[int64]map[int64]struct{}
+
+	// Mutex to avoid race conditions while reading writing to the map.
+	connectionIDsByStreamIDMutex sync.Mutex
+
+	idleTimerByStreamID      map[int64]*time.Timer
+	idleTimerByStreamIDMutex sync.Mutex
 }
 
 // AgentTokenAuthenticationOptions contains list of parameters required for agent token based authentication
@@ -352,10 +367,12 @@ func NewProxyServer(serverID string, proxyStrategies []ProxyStrategy, serverCoun
 		BackendManagers:            bms,
 		AgentAuthenticationOptions: agentAuthenticationOptions,
 		// use the first backend-manager as the Readiness Manager
-		Readiness:          bms[0],
-		proxyStrategies:    proxyStrategies,
-		warnOnChannelLimit: warnOnChannelLimit,
-		grpcMaxIdleTime:    grpcMaxIdleTime,
+		Readiness:               bms[0],
+		proxyStrategies:         proxyStrategies,
+		warnOnChannelLimit:      warnOnChannelLimit,
+		grpcMaxIdleTime:         grpcMaxIdleTime,
+		connectionIDsByStreamID: map[int64]map[int64]struct{}{},
+		idleTimerByStreamID:     map[int64]*time.Timer{},
 	}
 }
 
@@ -374,7 +391,61 @@ func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
 	recvCh := make(chan *client.Packet, xfrChannelSize)
 	stopCh := make(chan error)
 
-	go s.serveRecvFrontend(stream, recvCh)
+	// If this param is specified, we set a timer. Otherwise, the timer
+	// channel will never receive anything, therefore never fire and act as
+	// there is no inactivity timeout for this connection.
+
+	// Assign a stream ID.
+	var t *time.Timer
+
+	streamID := int64(0)
+
+	if s.grpcMaxIdleTime != 0 {
+		streamID = atomic.AddInt64(&s.nextStreamID, 1)
+		t = time.NewTimer(s.grpcMaxIdleTime)
+		s.idleTimerByStreamIDMutex.Lock()
+		s.idleTimerByStreamID[streamID] = t
+		s.idleTimerByStreamIDMutex.Unlock()
+
+		// This goroutine closes the idle streams.
+		// We get all the connectionIDs by stream ID and if the timer fires, we send a close request
+		// to the recvCh itself. This is the same channel on which the proxy-server sends packets after
+		// receiving from the client (i.e kube-apiserver). This way we simulate the behaviour as if the
+		// kube-apiserver is sending the connection close request, hence completing the cycle of a packet
+		// and keeping the map that stores the connection information clean.
+		// Under certain conditions, it was noticed that apiserver establishes a connection by sending the DIAL_REQ
+		// and receiving the DIAL_RSP, but no PacketType_DATA is ever sent, hence we never get the connID by
+		// which we can close the connection. In order to mitigate that, we can take advantage of the
+		// fact that since the stream is idle and we close the stream. More details regarding the issue in
+		// https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/311
+		go func() {
+			select {
+			case <-stopCh:
+				return
+			case <-t.C:
+				s.connectionIDsByStreamIDMutex.Lock()
+				klog.V(2).InfoS("Closing stream due to inactivity", "streamID", streamID, "serverID", s.serverID)
+				for connID := range s.connectionIDsByStreamID[streamID] {
+					pkt := &client.Packet{
+						Type: client.PacketType_CLOSE_REQ,
+						Payload: &client.Packet_CloseRequest{
+							CloseRequest: &client.CloseRequest{
+								ConnectID: connID,
+							},
+						},
+					}
+
+					klog.V(5).InfoS("Connection inactivity timeout, sending CLOSE_REQ", "serverID", s.serverID, "connectionID", connID, "streamID", streamID)
+					recvCh <- pkt
+
+					stream.Send(nil)
+				}
+				s.connectionIDsByStreamIDMutex.Unlock()
+			}
+		}()
+	}
+
+	go s.serveRecvFrontend(stream, recvCh, streamID, t)
 
 	defer func() {
 		klog.V(2).InfoS("Receive channel on Proxy is stopping", "userAgent", userAgent, "serverID", s.serverID)
@@ -385,6 +456,7 @@ func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
 	go func() {
 		for {
 			in, err := stream.Recv()
+
 			if err == io.EOF {
 				klog.V(2).InfoS("Stream closed on Proxy", "userAgent", userAgent, "serverID", s.serverID)
 				close(stopCh)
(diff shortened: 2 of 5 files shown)


pkg/server/server.go
```diff
@@ -471,6 +545,15 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
 
 		case client.PacketType_DATA:
 			connID := pkt.GetData().ConnectID
+
+			// Store the connID in the map.
```
invidian, line 549: Ditto condition.

pkg/server/server.go
```diff
@@ -145,6 +146,18 @@ type ProxyServer struct {
...
+	grpcMaxIdleTime time.Duration
+
+	// Tracks the stream ID.
+	nextStreamID int64
+
+	// Map to track the connections per stream ID.
+	connectionIDsByStreamID map[int64]map[int64]struct{}
+
+	// Mutex to avoid race conditions while reading writing to the map.
+	connectionIDsByStreamIDMutex sync.RWMutex
```
jkh52, line 160: Currently this is only used as a mutex, can replace with sync.Mutex

pkg/server/server.go
```diff
@@ -460,6 +530,10 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
 				// TODO: retry with other backends connecting to this agent.
 				klog.ErrorS(err, "CLOSE_REQ to Backend failed", "serverID", s.serverID, "connectionID", connID)
 			}
+
+			s.connectionIDsByStreamIDMutex.Lock()
+			delete(s.connectionIDsByStreamID[streamID], connID)
+			s.connectionIDsByStreamIDMutex.Unlock()
```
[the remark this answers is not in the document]
rata: Probably in its own method too :)

pkg/server/server.go
rata: And ditto method :)

pkg/server/server.go
```diff
@@ -370,7 +385,58 @@ func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
...
+	go func() {
+		select {
+		case <-stopCh:
+			return
+		case <-streamIdleTimer.C:
+			s.connectionIDsByStreamIDMutex.Lock()
+			switch len(s.connectionIDsByStreamID[streamID]) {
+			case 0:
+				klog.V(2).InfoS("Closing stream due to inactivity", "streamID", streamID, "serverID", s.serverID)
+				stream.Send(nil)
```
rata, line 419: Why send nil?

rata (commented):
Added some small comments after a cursory look. Will have a more detailed look later. @ipochi please note that tests are failing

I'm curious what @cheftako and @jkh52 think about this "big hammer" hack. This is what is mentioned in the 4th item [here](https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/276#issuecomment-974141778), I explain why I'd like some big hammer like that (basically, when the leak is in the konn-client, we need to backport to k8s versions and having a fix in the proxy-server is way easier to deploy, besides helping with future regressions too until they are fixed. Currently we had to disable konnectivity until we fix these leaks, with a hammer, we expect that is not needed to disable it but can use the hammer until we fix the leak).

@ipochi does this address the mem issues or not? I will test locally, but not all runs give the same results, and you probably run quite a lot already :)

cmd/server/app/options/options.go
```diff
@@ -41,6 +41,9 @@ type ProxyRunOptions struct {
 	// pings the client to see if the transport is still alive.
 	KeepaliveTime         time.Duration
 	FrontendKeepaliveTime time.Duration
+	// Time that a gRPC connection can be inactive (no pkt received) before
```
cheftako, line 44: "gRPC connection" is a bit confusing. We have a long lived gRPC connection between the KonnServer and KonnAgent which makes up part of the tunnel. I think it would be good to elaborate on the connections we are effecting. Proxied connections? Request connections? Client connections?

cheftako:
> I'm curious what Walter Fender and Joseph Anttila Hall think about this "big hammer" hack. This is what is mentioned in the 4th item [here](https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/276#issuecomment-974141778), I explain why I'd like some big hammer like that (basically, when the leak is in the konn-client, we need to backport to k8s versions and having a fix in the proxy-server is way easier to deploy, besides helping with future regressions too until they are fixed. Currently we had to disable konnectivity until we fix these leaks, with a hammer, we expect that is not needed to disable it but can use the hammer until we fix the leak).

As long as this is an opt in change (which is seems to be) I am pro the approach. I do have a couple of minor concerns. One is this introduces yet another ID for us to track. I would like to see if it corresponds to something we already have. We already have a lot of IDs and things to track. Adding another one just makes debugging the system even harder. 

I also would like to see if we can fix the frontend connection to be mux'd on a single gRPC connection between each KAS * KonnServer pair (when gRPC is configured). I don't think this prevents that but it may complicate it slightly.

cheftako:
# sigs.k8s.io/apiserver-network-proxy/tests [sigs.k8s.io/apiserver-network-proxy/tests.test]
tests/proxy_test.go:427:33: not enough arguments in call to server.NewProxyServer
	have (string, []server.ProxyStrategy, int, *server.AgentTokenAuthenticationOptions, bool)
	want (string, []server.ProxyStrategy, int, *server.AgentTokenAuthenticationOptions, bool, time.Duration)
tests/proxy_test.go:466:28: not enough arguments in call to server.NewProxyServer
	have (string, []server.ProxyStrategy, number, *server.AgentTokenAuthenticationOptions, bool)
	want (string, []server.ProxyStrategy, int, *server.AgentTokenAuthenticationOptions, bool, time.Duration)
# sigs.k8s.io/apiserver-network-proxy/pkg/server [sigs.k8s.io/apiserver-network-proxy/pkg/server.test]
pkg/server/server_test.go:162:23: not enough arguments in call to NewProxyServer
	have (string, []ProxyStrategy, number, *AgentTokenAuthenticationOptions, bool)
	want (string, []ProxyStrategy, int, *AgentTokenAuthenticationOptions, bool, time.Duration)
pkg/server/server_test.go:192:21: not enough arguments in call to NewProxyServer
	have (string, []ProxyStrategy, number, nil, bool)
	want (string, []ProxyStrategy, int, *AgentTokenAuthenticationOptions, bool, time.Duration)
pkg/server/server_test.go:200:20: not enough arguments in call to NewProxyServer
	have (string, []ProxyStrategy, number, nil, bool)
	want (string, []ProxyStrategy, int, *AgentTokenAuthenticationOptions, bool, time.Duration)
FAIL	sigs.k8s.io/apiserver-network-proxy/pkg/server [build failed]

rata:
@cheftako cool, good to know this approach looks fine if it is opt-in (as it is, yes :)). Thanks!

And yeah, I'd like this to be simpler too. We will see with @ipochi if this can be simplified or at least reduce the complexity when this opt-in flag is not used

cmd/server/app/options/options.go
ipochi: Thanks for this feedback. This is about the idle time between frontend and the proxy-server in which no packet type is transferred on the stream.

Initially I wanted to reuse the FrontendKeepaliveTime, but I felt making it separate would be much better to make this approach opt-in.

ipochi:
UPDATE:

Found an issue with this PR which continuously increases memory after 5 or 6 hours. Still trying to debug the cause of this issue.

While the PR terminates idle connections which decrease the `open_fds` and `vendor/google.golang.org/grpc/internal/transport/http2_server.go` metric , surprisingly this only results is ~10% memory usage reduction despite closing all open grpc connections.

This same behaviour or only ~10%  memory usage reduction after the test was noticed by using the out of the box gRPC Keepalive parameters (`MaxConnectionAge` and `MaxConnectionAgeGrace`). Curently outweighing the pros and cons of both the approaches to terminate idle connections:

This PR : terminates frontend->proxy-server connections based on inactivity after a specified duration.
gRPC Params: terminates all frontend->proxy-server connections regardless of activity.

PR #313, opened 2021-12-20
Agent: Add flag for health server listen address

alvaroaleman:
This allows for example to make the health server listen on the
localhost interface only. The default of binding to all interfaces is
unchanged.

### DIFF
--- a/cmd/agent/app/options/options.go
+++ b/cmd/agent/app/options/options.go
@@ -27,6 +27,7 @@ type GrpcProxyAgentOptions struct {
 	AlpnProtos      []string
 
 	// Ports for the health and admin server
+	HealthServerHost string
 	HealthServerPort int
 	AdminServerPort  int
 	// Enables pprof at host:adminPort/debug/pprof.
@@ -80,6 +81,7 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.ProxyServerHost, "proxy-server-host", o.ProxyServerHost, "The hostname to use to connect to the proxy-server.")
 	flags.IntVar(&o.ProxyServerPort, "proxy-server-port", o.ProxyServerPort, "The port the proxy server is listening on.")
 	flags.StringSliceVar(&o.AlpnProtos, "alpn-proto", o.AlpnProtos, "Additional ALPN protocols to be presented when connecting to the server. Useful to distinguish between network proxy and apiserver connections that share the same destination address.")
+	flags.StringVar(&o.HealthServerHost, "health-server-host", o.HealthServerHost, "The host address to listen on, without port.")
 	flags.IntVar(&o.HealthServerPort, "health-server-port", o.HealthServerPort, "The port the health server is listening on.")
 	flags.IntVar(&o.AdminServerPort, "admin-server-port", o.AdminServerPort, "The port the admin server is listening on.")
 	flags.BoolVar(&o.EnableProfiling, "enable-profiling", o.EnableProfiling, "enable pprof at host:admin-port/debug/pprof")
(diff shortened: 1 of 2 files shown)


cmd/agent/app/options/options.go
```diff
@@ -192,6 +194,7 @@ func NewGrpcProxyAgentOptions() *GrpcProxyAgentOptions {
 		CaCert:                    "",
 		ProxyServerHost:           "127.0.0.1",
 		ProxyServerPort:           8091,
+		HealthServerAddr:          "",
```
jkh52, line 197: naming nit: HealthServerHost (to be consistent with ProxyServerHost)

alvaroaleman: The ProxyServerHost is a host to connect to, as opposed to an address to listen on. One doesn't really bind a listener to a `Host` IMHO

jkh52: Yes, but it seems to me that an address is composed of a host + port, and since this flag only covers the former, it is better named "~host".

alvaroaleman: Updated

jkh52:
It would be good to update NewGrpcProxyAgentOptions and Print functions as well, for consistency with other flags. Sorry for not catching before.

cmd/agent/app/options/options.go
```diff
@@ -80,6 +81,7 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.ProxyServerHost, "proxy-server-host", o.ProxyServerHost, "The hostname to use to connect to the proxy-server.")
 	flags.IntVar(&o.ProxyServerPort, "proxy-server-port", o.ProxyServerPort, "The port the proxy server is listening on.")
 	flags.StringSliceVar(&o.AlpnProtos, "alpn-proto", o.AlpnProtos, "Additional ALPN protocols to be presented when connecting to the server. Useful to distinguish between network proxy and apiserver connections that share the same destination address.")
+	flags.StringVar(&o.HealthServerHost, "health-server-host", o.HealthServerHost, "The host address to listen on, without port.")
```
cheftako, line 84: Any new configuration options should be reflected in the Print() method below.

cheftako: Would also be nice if we could validate the value configured, though that seems harder than it sounds (and not required .. yet).

andrewsykim, line 84: I think the general naming convention for this is:
```
--health-server-addr
```

andrewsykim: Ah, I see this comment abbove now https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/313#discussion_r772608611

cheftako: Can we add the HealthServerHost field to the print() method?

alvaroaleman: Added it, sorry, this PR fell off my plate during the end of last year

alvaroaleman:
@cheftako @andrewsykim I think all comments are addressed, please have another look

cmd/agent/app/options/options.go
andrewsykim: > Ah, I see this comment abbove now #313 (comment)

sorry for the confusion, ^ when I said this I meant you can ignore my comment :)

andrewsykim: (I think it's better to be consistent with --proxy-server-host, and while it'd be nice if both ended in `-addr`, I don't think it's worth breaking the existing flag)

alvaroaleman: ok, changed the name back to `health-server-host`

cmd/agent/app/server.go
```diff
@@ -93,7 +93,7 @@ func (a *Agent) runHealthServer(o *options.GrpcProxyAgentOptions) error {
 	muxHandler.HandleFunc("/ready", readinessHandler)
 	muxHandler.HandleFunc("/readyz", readinessHandler)
 	healthServer := &http.Server{
-		Addr:           fmt.Sprintf(":%d", o.HealthServerPort),
+		Addr:           fmt.Sprintf("%s:%d", o.HealthServerHost, o.HealthServerPort),
```
andrewsykim, line 96: This should use `net.JoinHostPort` to future-proof this for IPv6

alvaroaleman: Done!

andrewsykim (commented):
/approve
/lgtm

PR #314, opened 2021-12-20
Update go builder image from 1.17.2 to 1.17.5.

jkh52:
Picks up security fixes:
https://go.dev/doc/devel/release#go1.17.minor


mainred (approved):
Looks good, thanks

PR #317, opened 2022-01-11
chore: reduce annoying error message

mainred:
We have customers bring concerns about this error though the code has comments to show there is no harm.

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -502,7 +502,7 @@ func (a *Client) remoteToProxy(connID int64, ctx *connContext) {
 			return
 		} else if err != nil {
 			// Normal when receive a CLOSE_REQ
-			klog.ErrorS(err, "connection read failure", "connectionID", connID)
+			klog.V(5).InfoS("connection read failure", "connectionID", connID, "err", err)
 			return
 		} else {
 			resp.Payload = &client.Packet_Data{Data: &client.Data{
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -501,8 +501,12 @@ func (a *Client) remoteToProxy(connID int64, ctx *connContext) {
 			klog.V(2).InfoS("connection EOF", "connectionID", connID)
 			return
 		} else if err != nil {
-			// Normal when receive a CLOSE_REQ
-			klog.V(5).InfoS("connection read failure", "connectionID", connID, "err", err)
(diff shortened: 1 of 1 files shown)


pkg/agent/client.go
```diff
@@ -502,7 +502,7 @@ func (a *Client) remoteToProxy(connID int64, ctx *connContext) {
 			return
 		} else if err != nil {
 			// Normal when receive a CLOSE_REQ
-			klog.ErrorS(err, "connection read failure", "connectionID", connID)
```
andrewsykim, line 505: It would be unforuntate to not log this though when there's an error when CLOSE_REQ wasn't received? 

Wondering if there's a more deterministic way to only ignore the log when we know CLOSE_REQ was received.

andrewsykim: I'm still getting familiar with the code here, but is it an option to check the connection ID from `a.connManager` before deciding the log level?

cheftako: If CLOSE_REQ has been called, then the cleanFunc will have been called. In the cleanFunc we remove connID from the connManager prior to closing the conn. So you should be able to use the presence of connID in the a.connManager to determine the log level.

andrewsykim: ^ @mainred can we make this change please?

pkg/agent/client.go
```diff
@@ -501,8 +501,12 @@ func (a *Client) remoteToProxy(connID int64, ctx *connContext) {
 			klog.V(2).InfoS("connection EOF", "connectionID", connID)
 			return
 		} else if err != nil {
-			// Normal when receive a CLOSE_REQ
-			klog.ErrorS(err, "connection read failure", "connectionID", connID)
+			if _, ok := a.connManager.Get(connID); !ok {
```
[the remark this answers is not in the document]
mainred: Thanks for pointing out. no connID means it has been deleted from a normal connection close, otherwise, we don't ignore this error.

mihivagyok:
@mainred @cheftako 
I'm just getting this error on the master branch with basic validation:
```
I0303 13:13:37.367516 2048686 client.go:406] "close connection" connectionID=1
E0303 13:13:37.367624 2048686 client.go:514] "connection read failure" err="read tcp 127.0.0.1:41224->127.0.0.1:8000: use of closed network connection" connectionID=1
```

Thanks!
Adam

Issue #309, opened 2021-12-13
arm docker images missing


ncopa:
The re are no arm images for version 0.0.27 and it breaks the the k0s build.

https://github.com/k0sproject/k0s/runs/4489019305?check_suite_focus=true

It would be nice to be able to update to 0.0.27 for todays k0s bugfix release.

rastislavs:
It seems that the support for ARM was removed by this commit: https://github.com/kubernetes-sigs/apiserver-network-proxy/commit/1c2dcc7aa58de993b79637d943d1878f7de9c000

I second to this request, I think arm / arm64 images should be published by default.

ipochi:
This was done temporary as per the comment in the #304. I wonder if that's still the case.
If not I'd like to pitch in to the needful to get it done

/cc @Jefftree @cheftako

Jefftree:
Yes this was a temporary measure because the original arm64 published was incorrect. The fix should be to just add back arm64 to the RELEASE_ARCH_LIST (https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/Makefile#L17)

rastislavs:
> Yes this was a temporary measure because the original arm64 published was incorrect. The fix should be to just add back arm64 to the RELEASE_ARCH_LIST (https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/Makefile#L17)

Just opened the PR with that: https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/320

PR #320, opened 2022-01-25
Add arm64 into RELEASE_ARCH_LIST

rastislavs:
Fixes https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/309

Signed-off-by: Rastislav Szabo <rastislav@kubermatic.com>


Jefftree:
/lgtm
/assign @cheftako 

The prow jobs built both (amd, arm) images very quickly but I imagine our staging push with cloudbuild will be much slower.

Issue #319, opened 2022-01-24
basic validation with GRPC with error logs


mihivagyok:
Branch:
master

Env:
```
Linux 5.4.0-72-generic #80-Ubuntu SMP Mon Apr 12 17:35:00 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
go version go1.17.6 linux/amd64
```

Problem:
I'm executing the basic test commands for Konnectivity:
https://github.com/kubernetes-sigs/apiserver-network-proxy#grpc-client-using-mtls-proxy-with-dial-back-agent

Client reports success, but my problem is that both server and agent report many errors during this basic GRPC scenario:
- server errors:
```
I0124 15:35:37.559080  532646 server.go:632] "Connect request from agent" agentID="9ff239f2-9603-4af1-92a0-61ff0ebba94d"
I0124 15:35:37.559981  532646 backend_manager.go:191] "Register backend for agent" connection=&{ServerStream:0xc0000ea0c0} agentID="9ff239f2-9603-4af1-92a0-61ff0ebba94d"
I0124 15:35:48.896258  532646 server.go:368] "proxy request from client" userAgent=[grpc-go/1.42.0]
I0124 15:35:48.896331  532646 server.go:407] start serving frontend stream
I0124 15:35:48.897481  532646 backend_manager.go:302] "Pick agent as backend" agentID="9ff239f2-9603-4af1-92a0-61ff0ebba94d"
I0124 15:35:48.900222  532646 server.go:269] "Register frontend for agent" frontend=&{Mode:grpc Grpc:0xc0004ad8c0 HTTP:<nil> CloseHTTP:<nil> connected:0xc000048240 connectID:1 agentID:9ff239f2-9603-4af1-92a0-61ff0ebba94d start:{wall:13867656054798239569 ext:26337495701 loc:0x212b2a0} backend:0xc000120030} agentID="9ff239f2-9603-4af1-92a0-61ff0ebba94d" connectionID=1
[...]

PR #322, opened 2022-02-01
fix: don't send closeReq twice if it is already sent

mihivagyok:
This PR tries to fix https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/319

- don't send closeReq twice if it is already sent: currently we have single use grpc tunnel, that's why connID does/shall not change during the lifecycle of the connection, that's why I think the firstConnID variable is not really necessary - as it is just controls some Info log currently
- although, I agree that we could add connID comparison/validation, if needed
- silent the rpc error Canceled (it comes in normal client-server close case)

@mainred I have just found that you have a similar PR:

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -414,7 +414,7 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer, 
 	klog.V(4).Infoln("start serving frontend stream")
 
 	var connID int64
-	var closeReqSent bool = false
+	var closeReqSent bool
 	// The first packet should be a DIAL_REQ, then we will get a
 	// backend from the configured BackendMangers.
 	var backend Backend
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -470,11 +470,9 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer, 
 				// TODO: retry with other backends connecting to this agent.
 				klog.ErrorS(err, "CLOSE_REQ to Backend failed", "serverID", s.serverID, "connectionID", connID)
 			} else {
-				klog.V(5).InfoS("CLOSE_REQ sent to backend", "serverID", s.serverID, "connectionID", connID)
+				klog.V(5).Infoln("CLOSE_REQ sent to backend", "serverID", s.serverID, "connectionID", connID)
+				closeReqSent = true
 			}
-			closeReqSent = true
-			klog.V(5).Infoln("CLOSE_REQ sent to backend", "serverID", s.serverID, "connectionID", connID)
-
 		case client.PacketType_DIAL_CLS:
 			random := pkt.GetCloseDial().Random
 			klog.V(5).InfoS("Received DIAL_CLOSE", "serverID", s.serverID, "dialID", random)
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -26,7 +26,9 @@ import (
 	"sync"
 	"time"
 
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
 	authv1 "k8s.io/api/authentication/v1"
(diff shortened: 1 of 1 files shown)


tamassoos3 (approved):

SzabolcsSzoke (approved):

pkg/server/server.go
```diff
@@ -406,9 +413,10 @@ func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
 func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer, recvCh <-chan *client.Packet) {
 	klog.V(4).Infoln("start serving frontend stream")
 
-	var firstConnID int64
-	// The first packet should be a DIAL_REQ, we will randomly get a
-	// backend from the BackendManger then.
+	var connID int64
+	var closeReqSent bool = false
```
andrewsykim, line 417: nit: the `= false` in the end is not needed

andrewsykim (commented):
/ok-to-test

pkg/server/server.go
```diff
@@ -387,7 +389,12 @@ func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
 				return
 			}
 			if err != nil {
-				klog.ErrorS(err, "Stream read from frontend failure")
+				rpcCode := status.Convert(err).Code()
+				if rpcCode == codes.Canceled {
+					klog.V(2).InfoS("Stream has been canceled", "rpc_code", rpcCode)
```
jkh52, line 394: we know it's canceled, so maybe "rpc_code" is uncessary here (but could be helpful below in the ErrorS)

jkh52:
Overall LGTM, and I like the two improvements (handle cancel more nicely, and avoid duplicate close). But the "firstConnID" safety check might be valuable to keep around; even if we don't see it happening now, it could help guard against a serious bug in the future.

pkg/server/server.go
mihivagyok: Hi! Do you mean shall I remove the logging when it is canceled? Thanks!

cheftako: I think it was also add the rpcCode to the error message when rpcCode != codes.Cancelled.

mihivagyok: Thank you! It is clear now :)

pkg/server/server.go
```diff
@@ -460,6 +469,7 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
 				// TODO: retry with other backends connecting to this agent.
 				klog.ErrorS(err, "CLOSE_REQ to Backend failed", "serverID", s.serverID, "connectionID", connID)
 			}
+			closeReqSent = true
```
andrewsykim, line 472: Here we are setting `closeReqSent = true` even if `backend.Send()` above errored

mihivagyok: Thanks, good comment! Added a `continue` in the error handling. Thank you!

pkg/server/server.go
```diff
@@ -459,7 +468,9 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
 			if err := backend.Send(pkt); err != nil {
 				// TODO: retry with other backends connecting to this agent.
 				klog.ErrorS(err, "CLOSE_REQ to Backend failed", "serverID", s.serverID, "connectionID", connID)
+				continue
```
andrewsykim, line 471: It might be slightly easier to understand this by using an else block here to set closeReqSent

mihivagyok: Sure, I could do that, I was just following the generic error handling pattern in the function :) Thank you!

mihivagyok:
@cheftako @andrewsykim Please take a look. I have rebased the change, and review for final review. Thanks!

jkh52:
FYI this seems fixed at https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/417

Issue #255, opened 2021-08-10
Konnectivity server leaks memory and free sockets


Tamas-Biro1:
Hello,

We are using ANP version 0.0.21 with a patch https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/179 (I know it is already in 0.0.22, but currently it is not an option to upgrade)

We have a cluster with 1 ANP server and 4 ANP agents. Sometimes we see that the server starts consuming lot of memory and once it reaches the limit then restarts, and the whole thing starts from the beginning. This is true also for the number of free sockets. Here is the prometheus graph for memory and socket numbers:

<img width="980" alt="image" src="https://user-images.githubusercontent.com/60390128/128824608-6210a139-a312-4317-97cc-a0c3afd5e040.png">

<img width="980" alt="image" src="https://user-images.githubusercontent.com/60390128/128824465-82097f80-5ce3-40f0-b456-1fe911f725b7.png">

From the server logs I only see the "usual" things could be seen in previous versions:
```
I0806 18:25:02.385941       1 server.go:736] "Close streaming" agentID="94a18c0b-efdf-476b-a8fa-10e0c0057345" connectionID=57
I0806 18:25:02.388576       1 server.go:723] "Received CLOSE_RSP" connectionID=85
E0806 18:25:02.388662       1 server.go:731] "CLOSE_RSP send to client stream error" err="tls: use of closed connection" connectionID=85
I0806 18:25:02.388764       1 server.go:283] "Remove frontend for agent" frontend=&{Mode:http-connect Grpc:<nil> HTTP:0xc001ff1c00 connected:0xc0000afbc0 connectID:85 agentID:1c6f546a-3e02-46fc-a061-dcb014d7969e start:{wall:13851803065787238258 ext:26239485524435 loc:0x218a780} backend:0xc000952fa0} agentID="1c6f546a-3e02-46fc-a061-dcb014d7969e" connectionID=85
I0806 18:25:02.388843       1 server.go:736] "Close streaming" agentID="1c6f546a-3e02-46fc-a061-dcb014d7969e" connectionID=85
```

and
```
I0806 18:25:02.428722       1 server.go:680] "Received DIAL_RSP" random=2394930260326768039 agentID="1c6f546a-3e02-46fc-a061-dcb014d7969e" connectionID=91
[...]

PR #324, opened 2022-02-10
fix: set timeout on sendmsg to avoid memory leak

mihivagyok:
- PR tries to fix https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/255
- I have found that if the server and agent tunnel hangs, new requests will cause memory leak in Server
- I made the change based on the following discussion: https://github.com/grpc/grpc-go/issues/1229#issuecomment-302755717 
- once this change is there, the memory leak does not happen
- **note**: this change does not solve the issue which make the tunnel hanging

### DIFF
--- a/pkg/server/backend_manager.go
+++ b/pkg/server/backend_manager.go
@@ -26,6 +26,7 @@ import (
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+
 	"k8s.io/klog/v2"
 	client "sigs.k8s.io/apiserver-network-proxy/konnectivity-client/proto/client"
 	pkgagent "sigs.k8s.io/apiserver-network-proxy/pkg/agent"
@@ -87,14 +88,13 @@ type backend struct {
 func (b *backend) Send(p *client.Packet) error {
 	b.mu.Lock()
 	defer b.mu.Unlock()
+	// https://github.com/grpc/grpc-go/issues/1229
+	// wrap a timer for around SendMsg to avoid blocking grpc call
+	// (e.g. stream is full)
 	errChan := make(chan error, 1)
 	go func() {
-		err := b.conn.Send(p)
-		errChan <- err
+		errChan <- b.conn.Send(p)
 		close(errChan)
-		if err != nil {
-			klog.ErrorS(err, "SendMsg has exited abnormally")
-		}
 	}()
 	t := time.NewTimer(10 * time.Second)
 	select {
--- a/pkg/server/backend_manager.go
+++ b/pkg/server/backend_manager.go
@@ -24,6 +24,8 @@ import (
 	"sync"
 	"time"
 
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"k8s.io/klog/v2"
 	client "sigs.k8s.io/apiserver-network-proxy/konnectivity-client/proto/client"
 	pkgagent "sigs.k8s.io/apiserver-network-proxy/pkg/agent"
@@ -85,7 +87,25 @@ type backend struct {
 func (b *backend) Send(p *client.Packet) error {
 	b.mu.Lock()
 	defer b.mu.Unlock()
-	return b.conn.Send(p)
+	errChan := make(chan error, 1)
+	go func() {
+		err := b.conn.Send(p)
+		errChan <- err
+		close(errChan)
+		if err != nil {
+			klog.ErrorS(err, "SendMsg has exited abnormally")
+		}
+	}()
+	t := time.NewTimer(10 * time.Second)
+	select {
+	case <-t.C:
+		return status.Errorf(codes.DeadlineExceeded, "conn.Send has timed out")
+	case err := <-errChan:
+		if !t.Stop() {
+			<-t.C
+		}
+		return err
+	}
 }
 
 func (b *backend) Context() context.Context {


andrewsykim (commented):
/ok-to-test

andrewsykim:
@mihivagyok are you able to share steps you used to reproduce and test the memory leak? I'd like to test/validate your patch against a cluster I have

mihivagyok:
@andrewsykim 

Configuration:
- konnectivity-server is configured with `--mode=http-connect` and `--proxy-strategies=destHost,default`
- there is one konnectivity-agent with `--agent-identifiers=ipv4=<NODE_IP>`
- konnectivity server has memory limit of 1 gb

Test:
- using `kubectl cp` to copy huge files to a test pod (̀for `kubectl cp`, Konnecticity servergets agent based on the destHost strategy - request matches with the agent identifier)
- eventually the `kubectl cp` fails to finish
- after that `kubectl logs` is failing (for `kubectl cp`, Konnecticity server also gets agent based on the destHost strategy - request matches with the agent identifier)
- among `kubectl cp/log`, there is a metrics server API server which generates traffic towards the cluster, but for that Konnectivity server picks random agent (default strategy)
- to make the leak, run thousands of `kubectl log` command - will fail, but it will increase the memory/socket usage heavily
- after some time, it will reach the memory limit, and kubernetes restarts konnectivity server

https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/255#issuecomment-947801403


off topic:
I think we have some idea why `kubectl cp` fails eventually: the issue comes when we have multiple proxy strategies. The backend (agent) connection are guarded by mutex. But if there are multiple strategy, then the same backend with the same connection will be registered to multiple backend managers. To illustrate this:
```
BackendManager - DestHost
    - Backend1 (agent identifier, conn, mutex)
BackedManager - Default
    - Backend1* (conn, mutex)
```

We have two agent instances (one for DestHost, one for Default backend manager), but with the same `conn`. In this case, it is possible that different go routines are trying to write to the same connection as there is no protection between the instances (there is mutex only within the instance), error could happen. I could submit an issue about this and discuss this theory there.

Thanks,
Adam

andrewsykim:
> note: this change does not solve the issue which make the tunnel hanging

Are you able to check the `konnectivity_network_proxy_server_grpc_connections` metric on the proxy server with this patch?

pkg/server/backend_manager.go
```diff
@@ -85,7 +87,25 @@ type backend struct {
 func (b *backend) Send(p *client.Packet) error {
 	b.mu.Lock()
 	defer b.mu.Unlock()
-	return b.conn.Send(p)
+	errChan := make(chan error, 1)
+	go func() {
+		err := b.conn.Send(p)
```
andrewsykim, line 92: Assuming https://github.com/grpc/grpc-go/issues/1229 is the problem we're seeing, should we be considering a similar fix when calling Send from client -> proxy server?

pkg/server/backend_manager.go
```diff
@@ -24,6 +24,8 @@ import (
 	"sync"
 	"time"
 
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
```
andrewsykim, line 28: small nit: add a line break between this and the k8s.io imports below

pkg/server/backend_manager.go
```diff
@@ -85,7 +87,25 @@ type backend struct {
 func (b *backend) Send(p *client.Packet) error {
 	b.mu.Lock()
 	defer b.mu.Unlock()
-	return b.conn.Send(p)
+	errChan := make(chan error, 1)
+	go func() {
```
andrewsykim, line 91: We should add a comment for why we have an explicit timeout with a link to https://github.com/grpc/grpc-go/issues/1229

pkg/server/backend_manager.go
```diff
@@ -85,7 +87,25 @@ type backend struct {
...
+		err := b.conn.Send(p)
+		errChan <- err
+		close(errChan)
+		if err != nil {
+			klog.ErrorS(err, "SendMsg has exited abnormally")
```
andrewsykim, line 96: I think we can remove this log line because the caller of Send (mainly from pkg/server/server.go) is wrapping or loggin the returned error already

andrewsykim: This would also simplify the code here to just:
```
go func() {
    errChan <- b.conn.Send(p)
    close(errChan)
}()

...
...
```

pkg/server/backend_manager.go
mihivagyok: Done. Thanks!

pkg/server/backend_manager.go
mihivagyok: Sure, good idea! Thank you!

andrewsykim:
@mihivagyok I think I'm able to reproduce this issue now, and it was possible while just using the default proxy strategy. From the goroutine stacktrace, this was the semaphore lock that was blocking goroutines:
```
goroutine 192765 [semacquire, 125 minutes]:
sync.runtime_SemacquireMutex(0xc00038c094, 0x1428700, 0x1)
        /usr/local/go/src/runtime/sema.go:71 +0x47
sync.(*Mutex).lockSlow(0xc00038c090)
        /usr/local/go/src/sync/mutex.go:138 +0x105
sync.(*Mutex).Lock(...)
        /usr/local/go/src/sync/mutex.go:81
sigs.k8s.io/apiserver-network-proxy/pkg/server.(*backend).Send(0xc00038c090, 0xc003b6fb00, 0x0, 0x0)
        /go/src/sigs.k8s.io/apiserver-network-proxy/pkg/server/backend_manager.go:86 +0xb9
sigs.k8s.io/apiserver-network-proxy/pkg/server.(*ProxyServer).serveRecvFrontend(0xc000283b00, 0x190d458, 0xc003a5d930, 0xc0072ddc20)
        /go/src/sigs.k8s.io/apiserver-network-proxy/pkg/server/server.go:432 +0xacc
created by sigs.k8s.io/apiserver-network-proxy/pkg/server.(*ProxyServer).Proxy
        /go/src/sigs.k8s.io/apiserver-network-proxy/pkg/server/server.go:367 +0x318
```

mihivagyok:
@andrewsykim That's great news.
I think the semaphore is working according to the design: one send is blocking, but an other request is coming and the mutex is guarding and the go routine gets blocked.

I think my concern regarding multiple proxy strategies / backend managers are still vaild [here](https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/324#issuecomment-1039234171).

To fix this, I think one backend manager is needed which could be configured with multiple strategies - so the connection would be used only in one manager. The one backend manager would select from its agents based on the strategies. I think the code could be changed easily to achieve this.

Do you think that if it is feasible or not?
Thank you!

andrewsykim:
> To fix this, I think one backend manager is needed which could be configured with multiple strategies - so the connection would be used only in one manager. The one backend manager would select from its agents based on the strategies. I think the code could be changed easily to achieve this.

> Do you think that if it is feasible or not?

I'm still getting familiar with the codebase so I'm not 100% sure yet. But if you're willing to open the PR it would be helpful for me to understand your proposal better

andrewsykim:
Btw -- it seems like the goroutine leaking due to backend connection mutex can happen for multiple reasons. In my specific test, it was due to write quota on the  stream, I created a separate issue for that here https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/335

pkg/server/backend_manager.go
```diff
@@ -85,7 +88,24 @@ type backend struct {
 func (b *backend) Send(p *client.Packet) error {
 	b.mu.Lock()
 	defer b.mu.Unlock()
-	return b.conn.Send(p)
+	// https://github.com/grpc/grpc-go/issues/1229
+	// wrap a timer for around SendMsg to avoid blocking grpc call
+	// (e.g. stream is full)
+	errChan := make(chan error, 1)
+	go func() {
```
andrewsykim, line 95: Is it possible for this goroutine to start leaking if b.conn.Send can block forever?

andrewsykim: My thinking is no, because when the stream closes eventually from returning from Send, this will return an io.EOF but I'm not 100% confident about it.

mihivagyok: Although I kinda agree that this change just masks the real problem. Thanks!

mihivagyok: Yes. 
I mention the buffered channel only because that is needed to be able free up the goroutine and the channel. Thanks'

PR #326, opened 2022-02-17
exit client with error when no agents are available

PratikDeoghare:
If there are no agents available `DialContext` fails after 30s with  `dial timeout, backstop` which is not sufficient to tell if the agents are not available or the dial is failing for some other reason. 

This PR makes it fail quickly with the error message received. Like this,

`error: failed to get dialer for client, got failed to dial request 127.0.0.:8000, got No backend available`.

This will also exit the goroutine as soon as we know it is no longer needed and remove the dial from pending dials.

### DIFF
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -27,7 +27,7 @@ import (
 	"time"
 
 	"google.golang.org/grpc"
-	"k8s.io/klog/v2"
+	klog "k8s.io/klog/v2"
 	"sigs.k8s.io/apiserver-network-proxy/konnectivity-client/proto/client"
 )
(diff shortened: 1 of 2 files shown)


mihivagyok:
@andrewsykim This is almost the same:

PR #307 is also having some conflict with my PR #322.

Thanks!
Adam

PratikDeoghare:
@mihivagyok Ah I see. I didn't know. 

The fix is definitely needed then :smile:

andrewsykim:
Thanks for the PR @PratikDeoghare, but I'm going to close this since it's a duplicate of https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/307

/close

PR #327, opened 2022-02-18
rename pkg/agent/client to pkg/agent/agent

PratikDeoghare:
It is easy to confuse  `pkg/agent/client ` with `konnectivity-client/pkg/client`, especially if you are coming back to the code after a long time or in the middle of debugging session. 

This renaming should help to remove the confusion.


mihivagyok:
I think this is a good idea! Thanks!

ipochi:
I like the idea. However I would suggest to be more descriptive of the intention in the commit message. I think the PR description could be the commit message.

PR #328, opened 2022-02-23
konnectivity-client: ensure Close() is called for grpc.ClientConn in CreateSingleUseGrpcTunnel

andrewsykim:
As part of the on-going investigation in https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/276.

If `grpcClient.Proxy(ctx)` from `CreateSingleUseGrpcTunnel` errors, then we never call Close() on the client connection. 

### DIFF
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -76,6 +76,7 @@ func CreateSingleUseGrpcTunnel(ctx context.Context, address string, opts ...grpc
 
 	stream, err := grpcClient.Proxy(ctx)
 	if err != nil {
+		c.Close()
 		return nil, err
 	}
 


konnectivity-client/pkg/client/client_test.go
```diff
@@ -151,6 +152,15 @@ func TestClose(t *testing.T) {
 	}
 }
 
+func TestCreateSingleUseGrpcTunnel_NoLeakOnFailure(t *testing.T) {
```
andrewsykim, line 155: Goroutine leak reproduced without the fix:
```
$ go test ./...
I0223 02:03:46.766980   27663 client.go:104] "stream read failure"
I0223 02:03:46.767325   27663 client.go:104] "stream read failure"
--- FAIL: TestCreateSingleUseGrpcTunnel_NoLeakOnFailure (0.45s)
    leaks.go:78: found unexpected goroutines:
        [Goroutine 25 in state select, with google.golang.org/grpc.(*ccBalancerWrapper).watcher on top of the stack:
        goroutine 25 [select]:
        google.golang.org/grpc.(*ccBalancerWrapper).watcher(0xc0001ae4c0)
                /usr/local/google/home/andrewsy/gke-internal-kubernetes/pkg/mod/google.golang.org/grpc@v1.27.1/balancer_conn_wrappers.go:69 +0x8f
        created by google.golang.org/grpc.newCCBalancerWrapper
                /usr/local/google/home/andrewsy/gke-internal-kubernetes/pkg/mod/google.golang.org/grpc@v1.27.1/balancer_conn_wrappers.go:60 +0x1ca
        
         Goroutine 26 in state select, with google.golang.org/grpc.(*addrConn).resetTransport on top of the stack:
        goroutine 26 [select]:
        google.golang.org/grpc.(*addrConn).resetTransport(0xc0001fe000)
                /usr/local/google/home/andrewsy/gke-internal-kubernetes/pkg/mod/google.golang.org/grpc@v1.27.1/clientconn.go:1149 +0x5f7
        created by google.golang.org/grpc.(*addrConn).connect
                /usr/local/google/home/andrewsy/gke-internal-kubernetes/pkg/mod/google.golang.org/grpc@v1.27.1/clientconn.go:815 +0x145
        ]
FAIL
FAIL    sigs.k8s.io/apiserver-network-proxy/konnectivity-client/pkg/client      0.496s
FAIL

```

andrewsykim: ^ consistent with pprof data collected here https://github.com/kubernetes/kubernetes/pull/108215#discussion_r812506220

konnectivity-client/go.mod
```diff
@@ -11,10 +11,11 @@ require (
 )
 
 require (
+	go.uber.org/goleak v1.1.12
```
andrewsykim, line 14: Might end up removing this dep if it complicates back-porting to k/k

konnectivity-client/pkg/client/client.go
```diff
@@ -76,6 +76,7 @@ func CreateSingleUseGrpcTunnel(ctx context.Context, address string, opts ...grpc
 
 	stream, err := grpcClient.Proxy(ctx)
 	if err != nil {
+		c.Close()
```
andrewsykim, line 79: `c.Close()` was only called from tunnel.serve() below. If grpc.Client.Proxy() errors, which is possible when there are no client connections available, we would never call c.Close()

konnectivity-client/pkg/client/client_test.go
rata: @andrewsykim  IMHO it will be great to add this test back. We had a bunch of fixes in the konnectivity-client and each fix there means we need to backport it to k8s, which is painful and slow to rollout to customers.

IMHO, it seems nice to keep it, the deps didn't seem like so many. What do you think?

andrewsykim: yeah I'll look into adding it back, it'll mostly depend on how it effects deps which I can look into

rata: @andrewsykim it seems it might not be an issue with go 1.17 if we just use it in tests? See the last paragraph in the selected answer: https://stackoverflow.com/questions/64071364/best-way-to-use-test-dependencies-in-go-but-prevent-export-them

Not sure if that will do the trick. Also, kubernetes 1.21 to 1.23 is using go 1.16 so, even if it does help, not right now. The other option might be to move the tests to another pkg, but it is not great either :(

andrewsykim: Created https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/333 to follow-up, I'll open a PR to add these back after I test the module update in k/k

rata: Thanks!

konnectivity-client/pkg/client/client.go
wojtek-t: good catch

PR #329, opened 2022-02-23
Fix misleading log messages in proxy server when sending DIAL_REQ and CLOSE_REQ

andrewsykim:


pkg/server/server.go
```diff
@@ -445,8 +445,9 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
 				})
 			if err := backend.Send(pkt); err != nil {
 				klog.ErrorS(err, "DIAL_REQ to Backend failed", "serverID", s.serverID)
+			} else {
+				klog.V(5).Infoln("DIAL_REQ sent to backend") // got this. but backend didn't receive anything.
```
rata, line 449: This makes me wonder if in case of error, we want to do nothing, as we currently do, or we want to "break" or something. If we want to break, then the else is not needed...

If we break here, we close the conn. I'm not sure if this is what we want, but unsure if keeping it open if the DIAL_REQ failed and is not retried is a state we want to be either... What do you think?

pkg/server/server.go
```diff
@@ -459,8 +460,9 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
 			if err := backend.Send(pkt); err != nil {
 				// TODO: retry with other backends connecting to this agent.
 				klog.ErrorS(err, "CLOSE_REQ to Backend failed", "serverID", s.serverID, "connectionID", connID)
+			} else {
+				klog.V(5).Infoln("CLOSE_REQ sent to backend", "serverID", s.serverID, "connectionID", connID)
```
rata, line 464: Idem, if sending the REQ failed, don't we want to close this anyways? So, add a `break` to the error case, instead of this else?

pkg/server/server.go
rata: @andrewsykim What do you think?

cheftako, line 449: If we're going to add this, it probably makes sense to also pass through the serverID and random. Also makes me think we should consider surfacing a backend/agent id.

andrewsykim: I definitely agree to revisit this logic and make sure we're returning correctly in the right places. I have to give it more thought.

andrewsykim: I don't disagree with this, but this log line already exists (see line below), this is just moving it into an else block so we don't log this on error.

PR #334, opened 2022-02-24
Remove gcloud tagging step for unpushed images

Jefftree:
Fix docker push prow job failure: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/apiserver-network-proxy-push-images/1473458166879490048

We recently removed the proxy-test-client and http-server from being pushed to the container registry.

/assign @cheftako


Jefftree:
ping @cheftako

PR #336, opened 2022-02-25
Add goroutine leak tests

andrewsykim:
Following up from this thread https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/328#discussion_r812506917 to re-add the goroutine leak test.

I sync'd the module version with what's in k8s.io/kubernetes to reduce potential go module issues.


andrewsykim (commented):
/assign @cheftako

konnectivity-client/pkg/client/client_test.go
```diff
@@ -151,6 +152,15 @@ func TestClose(t *testing.T) {
 	}
 }
 
+func TestCreateSingleUseGrpcTunnel_NoLeakOnFailure(t *testing.T) {
+	defer goleak.VerifyNone(t, goleak.IgnoreCurrent())
+
+	_, err := CreateSingleUseGrpcTunnel(context.Background(), "127.0.0.1:12345", grpc.WithInsecure())
+	if err == nil {
```
rata, line 159:
```suggestion
        // Test a connection that fails to establish
	t, err := CreateSingleUseGrpcTunnel(context.Background(), "127.0.0.1:12345", grpc.WithInsecure())
	if err == nil || t != nil {
```

go.mod
```diff
@@ -38,11 +38,14 @@ require (
 	github.com/prometheus/client_model v0.2.0 // indirect
 	github.com/prometheus/common v0.10.0 // indirect
 	github.com/prometheus/procfs v0.2.0 // indirect
+	go.uber.org/goleak v1.1.10 // indirect
```
rata, line 41: Just curious, go mod tidy adds this here? I'm not sure why, but I guess it is ok... Or did you add it manually?

konnectivity-client/pkg/client/client_test.go
rata: Untested, but just looking at the code seems we could test if `t` is nil or not.

My main suggestion here, though, is not that but to have a simple comment saying we expect this connection to fail (another option might be to have a var for the IP/port and use a name that makes it clear we want this to fail)

rata:
Overall, ths LGTM. Just some minor nits

PR #337, opened 2022-03-01
konnectivity-client: fix race condition with pending dials leading to goroutine leaks

andrewsykim:
This fixes a race condition with how konnectivity-client handles pending dials, which led to goroutine leaks in both kube-apiserver and proxy-server. The race condition was possible due to the use of buffered channels for receiving results for pending dials. See https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/276#issuecomment-1055025833 for more details.

### DIFF
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -130,6 +130,7 @@ func (t *grpcTunnel) serve(c clientConn) {
 					//   2. grpcTunnel.DialContext() returned early due to a dial timeout or the client canceling the context
 					//
 					// In either scenario, we should return here as this tunnel is no longer needed.
+					klog.V(1).InfoS("DialResp has no receiver; dropped", "connectionID", resp.ConnectID, "dialID", resp.Random)
 					return
 				}
 			}
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -125,8 +125,11 @@ func (t *grpcTunnel) serve(c clientConn) {
 				select {
 				case ch <- result:
 				default:
-					klog.ErrorS(fmt.Errorf("blocked pending channel"), "Received second dial response for connection request", "connectionID", resp.ConnectID, "dialID", resp.Random)
-					// On multiple dial responses, avoid leaking serve goroutine.
+					// If there are no readers of the pending dial channel above, it means one of two things:
+					//   1. There was a second DIAL_RSP for the connection request (this is very unlikely but possible)
+					//   2. grpcTunnel.DialContext() returned early due to a dial timeout or the client canceling the context
+					//
+					// In either scenario, we should return here as this tunnel is no longer needed.
 					return
 				}
 			}
@@ -183,7 +186,7 @@ func (t *grpcTunnel) DialContext(ctx context.Context, protocol, address string) 
 	}
 
 	random := rand.Int63() /* #nosec G404 */
-	resCh := make(chan dialResult, 1)
+	resCh := make(chan dialResult)
 	t.pendingDialLock.Lock()
 	t.pendingDial[random] = resCh
 	t.pendingDialLock.Unlock()
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -231,8 +231,10 @@ func (t *grpcTunnel) DialContext(ctx context.Context, protocol, address string) 
 		t.conns[res.connid] = c
 		t.connsLock.Unlock()
 	case <-time.After(30 * time.Second):
+		klog.V(5).InfoS("Timed out waiting for DialResp", "dialID", random)
 		return nil, errors.New("dial timeout, backstop")
 	case <-ctx.Done():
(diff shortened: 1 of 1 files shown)


andrewsykim:
/hold 

(pending validation from @rata)

konnectivity-client/pkg/client/client.go
```diff
@@ -183,7 +186,7 @@ func (t *grpcTunnel) DialContext(ctx context.Context, protocol, address string)
 	}
 
 	random := rand.Int63() /* #nosec G404 */
-	resCh := make(chan dialResult, 1)
+	resCh := make(chan dialResult)
```
rata, line 189: I think I'd at least add a comment here. This channel was unbuffered before, but was made buffered by this commit: 9bbe1fc5fac157b660fc8a47e10d57c0ff29573c. This probably would not have happened if we at least had a comment there.

(the logic to return if sending to the channel blocks didn't exist back then, but was introduced here: 9f0e43ed. It probably was buggy before this commit too, as it would block indefinitely in that case)

```suggestion
        // This channel MUST NOT be buffered. The sender needs to know when we are not receiving things, so they can abort.
	resCh := make(chan dialResult)
```

What do you think?

rata (approved):
This indeed seems to fix the remaining leak in #276 , thanks!

This overall LGTM, added a small nit about a comment, but of course no blocker to merge this (we need to merge this and create a new release, so we can update it in k8s)

I'd love if we can add some kind of test for this, though. We can add it in a different PR, though, no need to delay the release for that :)

rata:
/retest

The `pull-apiserver-network-proxy-docker-build-arm64` test fail seems to be unrelated:
```
protoc -I . proto/agent/agent.proto --go_out=plugins=grpc:/home/prow/go/src
make[1]: *** [Makefile:115: proto/agent/agent.pb.go] Error 127
make: *** [Makefile:228: docker-build/proxy-agent-arm64] Error 2
```

It seems just `protoc` is missing there.

rata:
Well, a retest did the trick. Tests are green now!

andrewsykim:
/retest

Prow flake:
```
Init container clonerefs not ready: (state: waiting, reason: "PodInitializing", message: "") Init container initupload not ready: (state: waiting, reason: "PodInitializing", message: "") Init container place-entrypoint not ready: (state: waiting, reason: "PodInitializing", message: "")
```

konnectivity-client/pkg/client/client.go
```diff
@@ -125,8 +125,11 @@ func (t *grpcTunnel) serve(c clientConn) {
 				select {
 				case ch <- result:
 				default:
-					klog.ErrorS(fmt.Errorf("blocked pending channel"), "Received second dial response for connection request", "connectionID", resp.ConnectID, "dialID", resp.Random)
-					// On multiple dial responses, avoid leaking serve goroutine.
+					// If there are no readers of the pending dial channel above, it means one of two things:
```
cheftako, line 128: I don't think we want to lose the logging for exiting dial_rsp here. Happy to talk about what the right log level is but we should be logging this.

andrewsykim: Added a V(1) log, similar to the one on line 118

andrewsykim:
/retest

Prow flake again:
```
Init container clonerefs not ready: (state: waiting, reason: "PodInitializing", message: "") Init container initupload not ready: (state: waiting, reason: "PodInitializing", message: "") Init container place-entrypoint not ready: (state: waiting, reason: "PodInitializing", message: "")
```

konnectivity-client/pkg/client/client.go
```diff
@@ -125,8 +125,12 @@ func (t *grpcTunnel) serve(c clientConn) {
 				select {
 				case ch <- result:
```
liggitt, line 126: at first glance, this looks like it's introducing a race in the opposite direction... 

we're sending from `err := t.stream.Send(req)` before we start receiving from the channel at `case res := <-resCh`

if the dial completes and we try to send the result to the channel before the receiving code reaches `case res := <-resCh`, won't we hit the default case and fail/close?

andrewsykim: Walter had raised a similar concern fwiw.

We _think_ that race condition is possible, although very unlikely, because `case res := <-resCh` happens immediately after `t.stream.Send(req) `. And `t.stream.Send(req) ` is a non-blocking call, in that it doesn't wait until the message is received by the other side. This dial response received here `case ch <- result` also only happens after a full round trip of the dial response from konnectivity agent -> konnectivity server -> kube-apiserver. So given the sending side of the channel requires a full roundtrip on the transport the race condition seems really unlikely. 

However, even if that race condition can happen, the change at least cleans up the gRPC client connection here, instead of hanging forever and waiting til the caller of DialContext cancels the context anyways.

liggitt: hmm, a fix that trades one race for a probably less likely race is hard for me to reason about the safety of... and certainly not one I'd want to backport to 3 existing release branches without soaking/proving it out. did we look for a way to do this in a non-racy way? https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/339 came to mind... we're already giving the serve loop a result channel to send to... can we also give it a cancel channel to unblock itself on that DialContext closes when it has given up on receiving?

PR #339, opened 2022-03-02
Avoid race between serve loop and DialContext receiving dial result

liggitt:
Fixes a possible race between `case res := <-resCh` in `DialContext` and `case ch <- result:` in the `serve` loop.

If that race occurred, no leak would occur, but `DialContext` would wait for the timeout or context cancellation, then return a failure.

This PR changes the send/default select in `serve` to select between sending to the result channel and receiving from a cancel channel. This lets DialContext explicitly indicate when it is done waiting on the result channel.

Added a unit test test demonstrating the race where the `serve` loop processes the dial response before the `DialContext` function makes it to waiting on the result channel. The test fails on master and works on this PR.

### DIFF
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -43,10 +43,17 @@ type dialResult struct {
 	connid int64
 }
 
+type pendingDial struct {
+	// resultCh is the channel to send the dial result to
+	resultCh chan<- dialResult
+	// cancelCh is the channel closed when resultCh no longer has a receiver
+	cancelCh <-chan struct{}
+}
+
 // grpcTunnel implements Tunnel
 type grpcTunnel struct {
 	stream          client.ProxyService_ProxyClient
-	pendingDial     map[int64]chan<- dialResult
+	pendingDial     map[int64]pendingDial
 	conns           map[int64]*conn
 	pendingDialLock sync.RWMutex
 	connsLock       sync.RWMutex
@@ -82,7 +89,7 @@ func CreateSingleUseGrpcTunnel(ctx context.Context, address string, opts ...grpc
 
 	tunnel := &grpcTunnel{
 		stream:             stream,
-		pendingDial:        make(map[int64]chan<- dialResult),
+		pendingDial:        make(map[int64]pendingDial),
 		conns:              make(map[int64]*conn),
 		readTimeoutSeconds: 10,
 	}
@@ -111,7 +118,7 @@ func (t *grpcTunnel) serve(c clientConn) {
 		case client.PacketType_DIAL_RSP:
 			resp := pkt.GetDialResponse()
 			t.pendingDialLock.RLock()
-			ch, ok := t.pendingDial[resp.Random]
+			pendingDial, ok := t.pendingDial[resp.Random]
 			t.pendingDialLock.RUnlock()
 
 			if !ok {
@@ -123,14 +130,16 @@ func (t *grpcTunnel) serve(c clientConn) {
 					connid: resp.ConnectID,
 				}
 				select {
-				case ch <- result:
-				default:
+				// try to send to the result channel
+				case pendingDial.resultCh <- result:
+				// unblock if the cancel channel is closed
+				case <-pendingDial.cancelCh:
 					// If there are no readers of the pending dial channel above, it means one of two things:
 					//   1. There was a second DIAL_RSP for the connection request (this is very unlikely but possible)
 					//   2. grpcTunnel.DialContext() returned early due to a dial timeout or the client canceling the context
 					//
 					// In either scenario, we should return here as this tunnel is no longer needed.
-					klog.V(1).InfoS("DialResp has no receiver; dropped", "connectionID", resp.ConnectID, "dialID", resp.Random)
+					klog.V(1).InfoS("Pending dial has been cancelled; dropped", "connectionID", resp.ConnectID, "dialID", resp.Random)
 					return
 				}
 			}
@@ -187,10 +196,16 @@ func (t *grpcTunnel) DialContext(ctx context.Context, protocol, address string) 
 	}
 
 	random := rand.Int63() /* #nosec G404 */
+
+	// This channel is closed once we're returning and no longer waiting on resultCh
+	cancelCh := make(chan struct{})
+	defer close(cancelCh)
+
 	// This channel MUST NOT be buffered. The sender needs to know when we are not receiving things, so they can abort.
 	resCh := make(chan dialResult)
+
 	t.pendingDialLock.Lock()
-	t.pendingDial[random] = resCh
+	t.pendingDial[random] = pendingDial{resultCh: resCh, cancelCh: cancelCh}
 	t.pendingDialLock.Unlock()
 	defer func() {
 		t.pendingDialLock.Lock()
--- a/konnectivity-client/pkg/client/client_test.go
+++ b/konnectivity-client/pkg/client/client_test.go
@@ -38,7 +38,7 @@ func TestDial(t *testing.T) {
 
 	tunnel := &grpcTunnel{
 		stream:      s,
-		pendingDial: make(map[int64]chan<- dialResult),
+		pendingDial: make(map[int64]pendingDial),
 		conns:       make(map[int64]*conn),
 	}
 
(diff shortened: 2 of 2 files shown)


andrewsykim (commented):
/assign

konnectivity-client/pkg/client/client.go
```diff
@@ -123,8 +130,10 @@ func (t *grpcTunnel) serve(c clientConn) {
 					connid: resp.ConnectID,
 				}
 				select {
-				case ch <- result:
-				default:
+				// try to send to the result channel
+				case pendingDial.resultCh <- result:
```
cheftako, line 134: So this can leak goroutines. cancelCh is only closed if DialContext exits. So if stream.Send below blocks (it is a blocking call) then the read will never be called and DialContext will not exit.

liggitt: is it possible for serve to get a DIAL_RSP and for the `Send` that initiated the dial to still be blocked?

liggitt: sure... but if it is blocked and hasn't sent the PacketType_DIAL_REQ message yet, can't we be 100% sure that the `serve` loop won't be handling a PacketType_DIAL_RESP response for it yet?

liggitt: This still handles the race condition in the two scenarios described below - a second DIAL_RSP and DialContext() returning early because of timeout or context cancellation

I'm not at all familiar with the `Send()` behavior, but it would really surprise me if it were possible for `Send()` to remain blocked after the DIAL_REQ message has actually been sent and received and acted on... I can't tell if you're saying that's possible or not

cheftako: > put another way, does Send() block after sending a DIAL_REQ message completely enough that the stream receiver got it and is responding with a DIAL_RESP message for serve to handle?

Good point. If the DIAL_REQ send blocks, your right we won't get a DIAL_RESP message. So we'll only leak one side of the connection.

rata:
@liggitt it might be worth pointing out in the commit that if that race happens there is no leak, but DialContext() will wait for a timeout (either the time or the context) to finish

liggitt:
> @liggitt it might be worth pointing out in the commit that if that race happens there is no leak, but DialContext() will wait for a timeout (either the time or the context) to finish

updated the description

konnectivity-client/pkg/client/client_test.go
```diff
@@ -59,6 +59,49 @@ func TestDial(t *testing.T) {
...
+	go tunnel.serve(&fakeConn{})
+	go ts.serve()
+
+	_, err := tunnel.DialContext(ctx, "tcp", "127.0.0.1:80")
+	if err != nil {
+		t.Fatalf("expect nil; got %v", err)
+	}
+
+	if ts.packets[0].Type != client.PacketType_DIAL_REQ {
+		t.Fatalf("expect packet.type %v; got %v", client.PacketType_CLOSE_REQ, ts.packets[0].Type)
```
rata, line 86:
```suggestion
	if ts.packets[0].Type != client.PacketType_DIAL_REQ {
		t.Fatalf("expect packet.type %v; got %v", client.PacketType_DIAL_REQ, ts.packets[0].Type)
```

konnectivity-client/pkg/client/client_test.go
```diff
@@ -59,6 +59,49 @@ func TestDial(t *testing.T) {
 	}
 }
 
+func TestDialRace(t *testing.T) {
```
rata, line 62: We need to explain what is the race you are testing here. Now we know, but the code will be fixed and it will be hard to guess. It will be very hard to maintain this test otherwise.

konnectivity-client/pkg/client/client_test.go
```diff
@@ -59,6 +59,49 @@ func TestDial(t *testing.T) {
...
+}
+
+// fakeSlowSend wraps ProxyService_ProxyClient and adds an artificial 1 second delay after calling Send
+type fakeSlowSend struct {
+	client.ProxyService_ProxyClient
+}
+
+func (s fakeSlowSend) Send(p *client.Packet) error {
+	err := s.ProxyService_ProxyClient.Send(p)
+	time.Sleep(time.Second)
```
rata, line 101:
```suggestion
        // To make the race more likely, we send the pkg so the other side can process it
        // We delay the send operation, so hopefully the other side process it before we start to receive on the channel
        // And therefore, excercise the case where the other side process the pkt before we receive in the channel.
	err := s.ProxyService_ProxyClient.Send(p)
	time.Sleep(time.Second)
```

konnectivity-client/pkg/client/client_test.go
rata, line 62: Maybe we want to detect go routines leaks in this test case? You can do something like this PR is doing with the uber pkg: https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/336

Note the use of `VerifyNone()` as that ignores the goroutine that is running due to klog.

rata (approved):
LGTM

Left some comments, mostly on the tests. I'm fine if we want to address them in other PRs and get the fixed to users ASAP.

konnectivity-client/pkg/client/client_test.go
liggitt: ha, this was a copy/paste of TestDial... should be fixed in both

konnectivity-client/pkg/client/client_test.go
```diff
@@ -51,14 +51,64 @@ func TestDial(t *testing.T) {
...
 	}
 
 	if ts.packets[0].GetDialRequest().Address != "127.0.0.1:80" {
 		t.Errorf("expect packet.address %v; got %v", "127.0.0.1:80", ts.packets[0].GetDialRequest().Address)
 	}
 }
 
+// TestDialRace exercises the scenario where serve() observes and handles DIAL_RSP
+// before DialContext() does any work after sending the DIAL_REQ, and ensures
+// the handoff of the established connection doesn't error and doesn't time out.
```
rata, line 64: I don't think that last part is super clear. What if we don't say that and just say we had bugs if this happens in the past?

```suggestion
// before DialContext() does any work after sending the DIAL_REQ. We had bugs
// handling this case in the past, make sure we don't regress.
```

rata (approved):
LGTM, one small clarification in the race explanation

(I think having a way longer explanation, with ASCII explaining what is happening on each side, might be fine too. For race conditions, it is never too clear. But I guess our future selves might be able to get it with this smaller text)

andrewsykim (commented):
/lgtm

I did some additional testing on this PR, following the same steps when testing https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/337 to double-check that the original goroutine leak is still not possible. Deferring approval to @cheftako

andrewsykim:
/retest

(prow flake)

konnectivity-client/pkg/client/client.go
cheftako: > I assume by block you mean "it is a blocking call that gets stucked", instead of "it is a blocking call that might take a few ms to execute" (as most blocking send can do, when they don't sync with the receiver).

Correct. In this instance I'm not worried about it blocking for a few ms or even seconds. My concern is we know from painful experience that on a quota imbalance that the stream.Send() call can block forever.

cheftako, line 134: Nit, the log in this case says ""DialResp has no receiver;" can we change that to something like "Pending dial has been canceled;" which seems more accurate? Having no receiver seems like a necessary but not sufficient condition now.

konnectivity-client/pkg/client/client.go
```diff
@@ -123,8 +130,10 @@ func (t *grpcTunnel) serve(c clientConn) {
...
+				// unblock if the cancel channel is closed
+				case <-pendingDial.cancelCh:
```
cheftako, line 136: Nit, the log in this case says ""DialResp has no receiver;" can we change that to something like "Pending dial has been canceled;" which seems more accurate? Having no receiver seems like a necessary but not sufficient condition now.

liggitt: that was based on the documented contract of `cancelCh`:

> cancelCh is the channel closed when resultCh no longer has a receiver

happy to reword to whatever makes sense

konnectivity-client/pkg/client/client.go
```diff
@@ -187,10 +196,16 @@ func (t *grpcTunnel) DialContext(ctx context.Context, protocol, address string)
 	}
 
 	random := rand.Int63() /* #nosec G404 */
+
+	// This channel is closed once we're returning and no longer waiting on resultCh
+	cancelCh := make(chan struct{})
+	defer close(cancelCh)
```
cheftako, line 202: I think it would be safer to do the close(cancelCh) under the t.pendingDialLock. I think it would make sense to combine the 2 defers here. 
Eg.
defer func() {
		t.pendingDialLock.Lock()
        close(cancelCh)
		delete(t.pendingDial, random) // Safe to delete missing key.
		t.pendingDialLock.Unlock()
	}()

konnectivity-client/pkg/client/client_test.go
cheftako: Given that the original problem being solved was a goroutine leak, would be nice to add this. Not sure this test is likely to catch such a leak but the wrapper seems low cost.

konnectivity-client/pkg/client/client.go
liggitt: There's no need to lock the close since we're the only closer. I think the close should stay where it is for a couple reasons:
1. an independent `defer` ensures no other panic (even inside the other defer function) keeps us from closing this channel
2. putting `defer close(cancelCh)` immediately after channel creation makes it really easy to see that this channel is always closed on return via all codepaths

liggitt:
updated comments and log message, commented on where the `close` is done. will defer leak tests to https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/336

liggitt:
weird, `pull-apiserver-network-proxy-docker-build-arm64` failed with a `protoc` not found error

/retest

konnectivity-client/pkg/client/client.go
cheftako: Sorry should have been more clear. I mean bring the rest of the defer logic here (Line 201), not push this logic down (To line 211). 

Having said that random should be deleted from pendingDial prior to cancelCh being closed which I believe negates my concern.

liggitt:
CI is happy, anything else before merge?

cheftako:
> CI is happy, anything else before merge?
Sorry Jordan. Keep wrestling with the fact we are moving the "pendingDial.resultCh <- result" write from non-blocking to blocking. In the konnectivity server having the routine blocking away from the stream.Recv() has been a real problem. However given that this is single use tunnel, I can't see how it will cause any problems.

konnectivity-client/pkg/client/client_test.go
```diff
@@ -51,14 +51,63 @@ func TestDial(t *testing.T) {
...
+	client.ProxyService_ProxyClient
+}
+
+func (s fakeSlowSend) Send(p *client.Packet) error {
+	// send the request so it can start being processed immediately
+	err := s.ProxyService_ProxyClient.Send(p)
+	// delay returning to simulate slowness on the client side,
+	// to exercise serve() observing/handling the DIAL_RSP before
+	// the client does any post-Send() work
+	time.Sleep(time.Second)
```
rata, line 107: @liggitt now that I think about it, this is not very nice. Sleep is not a good way to sync, maybe we can use the channles or a proper sync mechanism to make sure the other side is where we expect before we continue.

konnectivity-client/pkg/client/client.go
rata: @cheftako oohh, that quota thing... good point. I'm not familiar with that, nor I have observed it myself. But yeah, that is something I missed, thanks

liggitt: I agree flowcontrol / quota could make `Send()` block for all sorts of reasons *before* completing sending the request, but it doesn't block *after*, right?

PR #344, opened 2022-03-11
Desthost: Add support for wildcard host entries

alvaroaleman:
This change adds support for wildcard host entries to the desthost
backend manager. Useful if all domains below a certain domain should use
a specific backend.

### DIFF
--- a/cmd/agent/app/options/options.go
+++ b/cmd/agent/app/options/options.go
@@ -92,7 +92,7 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.DurationVar(&o.SyncIntervalCap, "sync-interval-cap", o.SyncIntervalCap, "The maximum interval for the SyncInterval to back off to when unable to connect to the proxy server")
 	flags.DurationVar(&o.KeepaliveTime, "keepalive-time", o.KeepaliveTime, "Time for gRPC agent server keepalive.")
 	flags.StringVar(&o.ServiceAccountTokenPath, "service-account-token-path", o.ServiceAccountTokenPath, "If non-empty proxy agent uses this token to prove its identity to the proxy server.")
-	flags.StringVar(&o.AgentIdentifiers, "agent-identifiers", o.AgentIdentifiers, "Identifiers of the agent that will be used by the server when choosing agent. N.B. the list of identifiers must be in URL encoded format. e.g.,host=localhost&host=node1.mydomain.com&cidr=127.0.0.1/16&ipv4=1.2.3.4&ipv4=5.6.7.8&ipv6=:::::&default-route=true")
+	flags.StringVar(&o.AgentIdentifiers, "agent-identifiers", o.AgentIdentifiers, "Identifiers of the agent that will be used by the server when choosing agent. N.B. the list of identifiers must be in URL encoded format. e.g.,host=localhost&host=node1.mydomain.com&cidr=127.0.0.1/16&ipv4=1.2.3.4&ipv4=5.6.7.8&ipv6=:::::&default-route=true. If a host starts with . it will be treated as wildcard")
 	flags.BoolVar(&o.WarnOnChannelLimit, "warn-on-channel-limit", o.WarnOnChannelLimit, "Turns on a warning if the system is going to push to a full channel. The check involves an unsafe read.")
 	flags.BoolVar(&o.SyncForever, "sync-forever", o.SyncForever, "If true, the agent continues syncing, in order to support server count changes.")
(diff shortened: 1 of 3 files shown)


andrewsykim:
Just a FYI, currently holding off on features until we sort out https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/346. This is so we can sustainably back-port bug fixes only into the current release without introducing more features.

pkg/server/desthost_backend_manager_test.go
```diff
@@ -0,0 +1,102 @@
...
+			backends: map[string][]*backend{
+				"sts.amazonaws.com": {newNamedBackend("sts.amazonaws.com")},
+			},
+			expectedBackendName: "sts.amazonaws.com",
+		},
+		{
+			name:     "Wildcard match",
+			destHost: "sts.amazonaws.com",
+			backends: map[string][]*backend{
+				".amazonaws.com": {newNamedBackend(".amazonaws.com")},
```
jkh52, line 49: why support both .amazonaws.com and *.amazonaws.com?

If we only support the latter, it seems more clear and simpler.

jkh52: Any reason you chose leading dot instead of leading asterisk? The asterisk seems closer to DNS spec (https://datatracker.ietf.org/doc/html/rfc4592#section-2.1.1). I don't see a lot of precedence for leading dots indicating wildcard.

cmd/agent/app/options/options.go
```diff
@@ -92,7 +92,7 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
...
 	flags.DurationVar(&o.KeepaliveTime, "keepalive-time", o.KeepaliveTime, "Time for gRPC agent server keepalive.")
 	flags.StringVar(&o.ServiceAccountTokenPath, "service-account-token-path", o.ServiceAccountTokenPath, "If non-empty proxy agent uses this token to prove its identity to the proxy server.")
-	flags.StringVar(&o.AgentIdentifiers, "agent-identifiers", o.AgentIdentifiers, "Identifiers of the agent that will be used by the server when choosing agent. N.B. the list of identifiers must be in URL encoded format. e.g.,host=localhost&host=node1.mydomain.com&cidr=127.0.0.1/16&ipv4=1.2.3.4&ipv4=5.6.7.8&ipv6=:::::&default-route=true")
+	flags.StringVar(&o.AgentIdentifiers, "agent-identifiers", o.AgentIdentifiers, "Identifiers of the agent that will be used by the server when choosing agent. N.B. the list of identifiers must be in URL encoded format. e.g.,host=localhost&host=node1.mydomain.com&cidr=127.0.0.1/16&ipv4=1.2.3.4&ipv4=5.6.7.8&ipv6=:::::&default-route=true. If a host starts with *. or . it will be treated as wildcard")
```
jkh52, line 95: Make sure to update (remove the 'or').

pkg/server/desthost_backend_manager_test.go
alvaroaleman: @jkh52 These vars are generally not standardized, so the effective standard that exist is curls behavior and curl only works with dot and not with an asterisk:
```bash
# Set proxy
$ export HTTPS_PROXY=http://undef.com

# proxy is invalid, so by default this doesn't work
$ curl https://google.com -v
* Uses proxy env variable HTTPS_PROXY == 'http://undef.com'
*   Trying 209.59.192.38:1080...
* connect to 209.59.192.38 port 1080 failed: Connection refused
* Failed to connect to undef.com port 1080 after 261 ms: Connection refused
* Closing connection 0
curl: (7) Failed to connect to undef.com port 1080 after 261 ms: Connection refused

# Set no proxy for tld with leading dot, works
$ NO_PROXY=.com curl https://google.com -v
* Uses proxy env variable NO_PROXY == '.com'
*   Trying 2a00:1450:4005:80b::200e:443...
* Connected to google.com (2a00:1450:4005:80b::200e) port 443 (#0)

# Set no proxy for tld with leading asterisk, doesn't work
NO_PROXY='*.com' curl https://google.com -v
* Uses proxy env variable NO_PROXY == '*.com'
* Uses proxy env variable HTTPS_PROXY == 'http://undef.com'
*   Trying 209.59.192.38:1080...
* connect to 209.59.192.38 port 1080 failed: Connection refused
* Failed to connect to undef.com port 1080 after 190 ms: Connection refused
* Closing connection 0
curl: (7) Failed to connect to undef.com port 1080 after 190 ms: Connection refused
```

PR #341, opened 2022-03-03
konnectivity-client: close channels for connections that never received CLOSE_RSP

andrewsykim:
This PR is a follow-up to https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/325 and fixes a goroutine leak that was reproducible in `TestProxyHandle_SlowContext_GRPC` when a client closes a transport before the CLOSE_RSP is received:
```
--- FAIL: TestProxyHandle_SlowContext_GRPC (4.47s)                                                                                                                                                 
    leaks.go:78: found unexpected goroutines:                                                                                                                                                      
        [Goroutine 2057 in state chan receive, with sigs.k8s.io/apiserver-network-proxy/konnectivity-client/pkg/client.(*conn).Read on top of the stack:                                           
        goroutine 2057 [chan receive]:                                                                                                                                                             
        sigs.k8s.io/apiserver-network-proxy/konnectivity-client/pkg/client.(*conn).Read(0xc001673860, {0xc0022fe000, 0x1000, 0x5f3513?})                                                           
                /usr/local/google/home/andrewsy/go/src/sigs.k8s.io/apiserver-network-proxy/konnectivity-client/pkg/client/conn.go:74 +0xc5                                                         
        net/http.(*persistConn).Read(0xc0022fa120, {0xc0022fe000, 0x1000, 0x1000})                                                                                                                 
                /usr/lib/google-golang/src/net/http/transport.go:1931 +0x110                                                                                                                       
        bufio.(*Reader).fill(0xc0005afec0)                                                                                                                                                         
                /usr/lib/google-golang/src/bufio/bufio.go:106 +0x294                                                                                                                               
        bufio.(*Reader).Peek(0xc0005afec0, 0x1)                                                                                                                                                    
                /usr/lib/google-golang/src/bufio/bufio.go:144 +0xcc                                                                                                                                
        net/http.(*persistConn).readLoop(0xc0022fa120)                                                                                                                                             
                /usr/lib/google-golang/src/net/http/transport.go:2092 +0x2bc                                                                                                                       
        created by net/http.(*Transport).dialConn                                                                                                                                                  
                /usr/lib/google-golang/src/net/http/transport.go:1752 +0x25c5   
```

Fixes https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/340

PR #351, opened 2022-04-12
Fixes connection leakage due to close timeout

yue9944882:
we noticed some connection leakage happening when the connections are not gracefully closed. the `conn#Read(..)` function will not return infinitely even if the `conn#Close(..)` called due to client-side timeout. this will cause memory leakage when ANP is working under an unstable-ish network environment. this PR notifies the `Read(..)` to exit w/ an error when the conn is aborted from client-side.

### DIFF
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -242,6 +242,7 @@ func (t *grpcTunnel) DialContext(ctx context.Context, protocol, address string) 
 		c.connID = res.connid
 		c.readCh = make(chan []byte, 10)
 		c.closeCh = make(chan string, 1)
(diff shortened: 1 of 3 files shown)


andrewsykim:
@yue9944882 do you have steps to reproduce the leak by any chance? Or is it only reproducible on live cluster with traffic?

yue9944882:
@andrewsykim the unit test in this pull can reproduce the leakage, w/o the fix `Read(..)` call will block infinitely

andrewsykim:
Thanks, I added ported your unit test over to https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/341. Hope you don't mind I added you as a co-author there

yue9944882:
@andrewsykim my pleasure

PR #353, opened 2022-04-14
agent: reduce noisy error logs when trying to write to a closed network connection

andrewsykim:
This is a follow-up PR to https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/317, where we reduced error logs whenever reading from a closed network connection. This PR introduces a similar fix but also from the writer.

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -507,9 +507,10 @@ func (a *Client) remoteToProxy(connID int64, ctx *connContext) {
 			klog.V(2).InfoS("connection EOF", "connectionID", connID)
 			return
 		} else if err != nil {
+			// "use of closed network connection" errors are expected upon receiving CLOSE_REQ
+			// If connID doesn't exist in connManager, we assume the connection was meant to be closed.
 			if _, ok := a.connManager.Get(connID); !ok {
-				// Normal when receive a CLOSE_REQ
-				klog.V(5).InfoS("read from a normally closed connection", "connectionID", connID, "err", err)
+				klog.V(5).InfoS("reading to a closed connection", "connectionID", connID, "err", err)
 			} else {
 				klog.ErrorS(err, "connection read failure", "connectionID", connID)
 			}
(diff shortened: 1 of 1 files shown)


pkg/agent/client.go
```diff
@@ -507,9 +507,10 @@ func (a *Client) remoteToProxy(connID int64, ctx *connContext) {
 			klog.V(2).InfoS("connection EOF", "connectionID", connID)
 			return
 		} else if err != nil {
+			// "use of closed network connection" errors are expected upon receiving CLOSE_REQ
+			// If connID doesn't exist in connManager, we assume the connection was meant to be closed.
 			if _, ok := a.connManager.Get(connID); !ok {
-				// Normal when receive a CLOSE_REQ
-				klog.V(5).InfoS("read from a normally closed connection", "connectionID", connID, "err", err)
+				klog.V(5).InfoS("reading to a closed connection", "connectionID", connID, "err", err)
```
andrewsykim, line 513: cc @mainred re: https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/317

pkg/agent/client.go
```diff
@@ -507,9 +507,10 @@ func (a *Client) remoteToProxy(connID int64, ctx *connContext) {
 			klog.V(2).InfoS("connection EOF", "connectionID", connID)
 			return
 		} else if err != nil {
+			// "use of closed network connection" errors are expected upon receiving CLOSE_REQ
+			// If connID doesn't exist in connManager, we assume the connection was meant to be closed.
 			if _, ok := a.connManager.Get(connID); !ok {
```
andrewsykim, line 512: Seems like we close the connection on the agent side before we actually delete connID from connManager, which makes me think this check is maybe not doing what we want:

https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/6c2477cfa8a6e11752a27bc99b4a06cc9499557c/pkg/agent/client.go#L413-L421

I'm not sure if deleting connID from connManager before ctx.conn.Close() is safe though.

pkg/agent/client.go
jkh52, line 513: Any reason this was changed to "reading to"? I normally think "reading from" or "writing to".

jkh52:
/unhold

I believe @andrewsykim is unavailable and the "to" vs. "from" is not worth holding on.

PR #356, opened 2022-05-13
Bugfix: send missing dial identifier when no agents are connected.

jkh52:
This allows the client to notice immediately, instead of timing out. I
forked this from a slightly larger
and will try to annotate PR so that mainred@'s contribution is recognized.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -416,25 +416,30 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer, 
 		switch pkt.Type {
 		case client.PacketType_DIAL_REQ:
 			klog.V(5).Infoln("Received DIAL_REQ")
+			random := pkt.GetDialRequest().Random
 			// TODO: if we track what agent has historically served
 			// the address, then we can send the Dial_REQ to the
 			// same agent. That way we save the agent from creating
 			// a new connection to the address.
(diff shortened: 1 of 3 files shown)


pkg/server/server.go
```diff
@@ -425,8 +425,13 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
 				klog.ErrorS(err, "Failed to get a backend", "serverID", s.serverID)
```
cheftako, line 425: Should probably add 
  "dialID", random 
to this error message.

pkg/server/server.go
```diff
@@ -425,8 +425,13 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
...
+					Type: client.PacketType_DIAL_RSP,
+					Payload: &client.Packet_DialResponse{
+						DialResponse: &client.DialResponse{
+							Random: pkt.GetDialRequest().Random,
+							Error:  err.Error(),
+						},
+					},
 				}
 				if err := stream.Send(resp); err != nil {
 					klog.V(5).InfoS("Failed to send DIAL_RSP for no backend", "error", err, "serverID", s.serverID)
```
cheftako, line 437: Ditto

pkg/server/server.go
jkh52: Good idea, done.

Issue #357, opened 2022-05-24
Issues with CreateSingleUseGrpcTunnel


ash2k:
1. `CreateSingleUseGrpcTunnel()` does not allow to use different contexts for dialing and connection lifetime control. There seem to be no way to shut the tunnel down.
2. Context cancellation/timeout in `grpcTunnel.DialContext()` doesn't interrupt any I/O, doesn't signal to the `serve()` goroutine in any way. This is suspicious and is possibly a source of connection leaks.

See this thread and PR for background https://github.com/kubernetes/kubernetes/pull/110079#discussion_r874800421.

jkh52:
This is marked rotten/stale, but I would argue that at this point it is fixed.

> CreateSingleUseGrpcTunnel() does not allow to use different contexts for dialing and connection lifetime control. There seem to be no way to shut the tunnel down.

This concern is still valid, but because of kube-apiserver connection re-use, there isn't an obvious way to close the single-use tunnel until the client calls `Close` on the `net.Conn` (actually a konnectivity `client.conn`).

> Context cancellation/timeout in grpcTunnel.DialContext() doesn't interrupt any I/O, doesn't signal to the serve() goroutine in any way. This is suspicious and is possibly a source of connection leaks.

This was fixed by https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/360.

PR #360, opened 2022-06-07
Introduce CreateSingleUseGrpcTunnelWithContext to replace

jkh52:
CreateSingleUseGrpcTunnel.

This API change is needed to properly fix
https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/357

### DIFF
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -35,7 +35,7 @@ import (
 type Tunnel interface {
 	// Dial connects to the address on the named network, similar to
 	// what net.Dial does. The only supported protocol is tcp.
-	DialContext(ctx context.Context, protocol, address string) (net.Conn, error)
+	DialContext(requestCtx context.Context, protocol, address string) (net.Conn, error)
 }
 
 type dialResult struct {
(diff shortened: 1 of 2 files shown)


konnectivity-client/pkg/client/client.go
```diff
@@ -74,14 +74,25 @@ var _ clientConn = &grpc.ClientConn{}
 // Currently, a single tunnel supports a single connection, and the tunnel is closed when the connection is terminated
 // The Dial() method of the returned tunnel should only be called once
 func CreateSingleUseGrpcTunnel(ctx context.Context, address string, opts ...grpc.DialOption) (Tunnel, error) {
```
jkh52, line 76: mark deprecated

konnectivity-client/pkg/client/client.go
```diff
@@ -175,6 +190,8 @@ func (t *grpcTunnel) serve(c clientConn) {
 				case <-timer.C:
 					klog.ErrorS(fmt.Errorf("timeout"), "readTimeout has been reached, the grpc connection to the proxy server will be closed", "connectionID", conn.connID, "readTimeoutSeconds", t.readTimeoutSeconds)
 					return
+				case <-tunnelCtx.Done():
+					klog.V(1).InfoS("Tunnel has been closed, the grpc connection to the proxy server will be closed", "connectionID", conn.connID)
```
andrewsykim, line 194: (coming back from a long break and just going through PRs I missed)

Wondering why we don't return early in this specific case when the tunnel context gets cancelled.

andrewsykim: Was playing around with a unit test to see if a potential leak was reproducible: https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/399

PR #361, opened 2022-06-08
Reduce proxy-server log noise.

jkh52:
Note that agentID and connectionID are already present in structured
logs; including the frontend doesn't add much value.

Example logs:
```
I1111 18:45:04.309310       1 server.go:262] "Register frontend for agent" frontend=&{Mode:grpc Grpc:0xc002874760 HTTP:<nil> connected:0xc002640f60 connectID:96 agentID:df752e37-615c-4ce9-a973-99136e1a8de5 start:{wall:13860803171892147580 ext:58170314102322 loc:0x2198760} backend:0xc0001b3fe0} agentID="df752e37-615c-4ce9-a973-99136e1a8de5" connectionID=96
I1111 18:51:25.308754       1 server.go:283] "Remove frontend for agent" frontend=&{Mode:grpc Grpc:0xc002874760 HTTP:<nil> connected:0xc002640f60 connectID:96 agentID:df752e37-615c-4ce9-a973-99136e1a8de5 start:{wall:13860803171892147580 ext:58170314102322 loc:0x2198760} backend:0xc0001b3fe0} agentID="df752e37-615c-4ce9-a973-99136e1a8de5" connectionID=96
```

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -266,7 +266,7 @@ func (s *ProxyServer) removeBackend(agentID string, conn agent.AgentService_Conn
 }
 
 func (s *ProxyServer) addFrontend(agentID string, connID int64, p *ProxyClientConnection) {
-	klog.V(2).InfoS("Register frontend for agent", "frontend", p, "agentID", agentID, "connectionID", connID)
+	klog.V(2).InfoS("Register frontend for agent", "agentID", agentID, "connectionID", connID)
 	s.fmu.Lock()
 	defer s.fmu.Unlock()
 	if _, ok := s.frontends[agentID]; !ok {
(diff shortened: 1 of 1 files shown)


pkg/server/server.go
```diff
@@ -266,7 +266,7 @@ func (s *ProxyServer) removeBackend(agentID string, conn agent.AgentService_Conn
 }
 
 func (s *ProxyServer) addFrontend(agentID string, connID int64, p *ProxyClientConnection) {
-	klog.V(2).InfoS("Register frontend for agent", "frontend", p, "agentID", agentID, "connectionID", connID)
```
cheftako, line 269: Why are we dropping tracking of the frontend (KAS) from these logs? 

I agree we should probably use an id and not the whole structure. However the agentID is the ID for the backend. The connection ID is the ID of the individual connection. Not sure we have such an ID today but we could add one when we get the initial DIAL REQ (https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/server/server.go#L417)

cheftako: Alternately we could enhance the client library to generate a UUID and send it but thats a bit trickier to do in a backward compatible way.

jkh52: Where kube-apiserver and proxy-server are 1:1 it doesn't seem like much value, but I suppose other configurations would like a frontend identifier here.

I like your second suggestion. Let me modify this PR to move in that direction.

/hold

PR #363, opened 2022-06-15
Make proxy server bind addresses configurable

### LINKED ISSUE #362: Allow to configure bind address for server
On multi homed machines it would be nice to be able to define the bind address for server. As there's multiple listeners (admin, metrics, agents) I'd guess we want to allow different addresses to be used for all of these and thus follow the same pattern as with agent. 

So what if we add following options to server:
- `agent-bind-address`: The address we listen for agent connections on.
- `admin-bind-address`: The address we listen for admin connections on.
- `health-bind-address`: The address we listen for health connections on.

I know this is bit out-of-line compared to agents `health-server-host` flag, but IMO `*-bind-address` is much more explicit that it's really the bind address we're talkign about. Also much more inline with other kube components like api-server,kubelet etc. as they all use `--bind-address`.

I can wrap up a PR for this, just wanted to get some feedback on the idea and naming of the flags.

jnummelin:
This way we can easily control on multi-homed machines which interfaces different endpoint listen on.

All but admin endpoint are listening by default on all interfaces. Admin is only binding to 127.0.0.1 by default.

Fixes #362 

Signed-off-by: Jussi Nummelin <jussi.nummelin@gmail.com>

### DIFF
--- a/cmd/server/app/server.go
+++ b/cmd/server/app/server.go
@@ -268,7 +268,7 @@ func (p *Proxy) runMTLSFrontendServer(ctx context.Context, o *options.ProxyRunOp
 		return nil, err
 	}
 
-	addr := fmt.Sprintf(":%d", o.ServerPort)
+	addr := fmt.Sprintf("%s:%d", o.ServerBindAddress, o.ServerPort)
 
 	if o.Mode == "grpc" {
 		frontendServerOptions := []grpc.ServerOption{
@@ -317,7 +317,7 @@ func (p *Proxy) runAgentServer(o *options.ProxyRunOptions, server *server.ProxyS
 		return err
 	}
 
-	addr := fmt.Sprintf(":%d", o.AgentPort)
+	addr := fmt.Sprintf("%s:%d", o.AgentBindAddress, o.AgentPort)
 	agentServerOptions := []grpc.ServerOption{
 		grpc.Creds(credentials.NewTLS(tlsConfig)),
 		grpc.KeepaliveParams(keepalive.ServerParameters{Time: o.KeepaliveTime}),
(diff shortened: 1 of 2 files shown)


jkh52:
It would be nice to use https://pkg.go.dev/net#JoinHostPort throughout for ipv6 compatibility.

jkh52:
It would be nice to keep things consistent with existing flags (below). But I like your flag names better, since they are more similar with other kubernetes binaries. If you're willing, this PR could introduce "proxy-server-bind-address" and "health-server-bind-address" (and mark the old ones deprecated).

```
./cmd/agent/app/options/options.go:	flags.StringVar(&o.ProxyServerHost, "proxy-server-host", o.ProxyServerHost, "The hostname to use to connect to the proxy-server.")
./cmd/agent/app/options/options.go:	flags.StringVar(&o.HealthServerHost, "health-server-host", o.HealthServerHost, "The host address to listen on, without port.")
```

This LGTM after using JoinHostPort (previous comment).

jnummelin:
I think I'll leave the agent side flags untouched in this PR as they are bit out of scope, if that's ok. (I can have a look later once I get back from holidays if no one else has done this yet.)

I'll push changes to use `[JoinHostPort](https://pkg.go.dev/net#JoinHostPort)` in an hour or so

cmd/server/app/server.go
```diff
@@ -344,7 +345,7 @@ func (p *Proxy) runAdminServer(o *options.ProxyRunOptions, server *server.ProxyS
 		}
 	}
 	adminServer := &http.Server{
-		Addr:           fmt.Sprintf("127.0.0.1:%d", o.AdminPort),
```
jkh52, line 347: To avoid weakening the default security posture, I would populate the new admin bind address flag to 127.0.0.1 (and document). Sorry for not noticing this before.

LGTM otherwise

jnummelin: hmm, not sure if I follow. Aren't we already doing this: https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/363/files/c8843948648d9e50274587a26a88bd2a797defbf#diff-803f4f0a42a022f49b840e5df162f6111f3d7b4e0790de682ef58263478f7140R330

jnummelin:
Any way we could get this to move forward?

jnummelin:
@cheftako @jkh52 What we need to do to get this forward?

cheftako:
Sorry not sure how this dropped off of the radar.
Nice change thank you.

jnummelin:
> Sorry not sure how this dropped off of the radar.

That happens time-to-time for any project, no worries 😄 Thanks for merging this. 🎉

PR #364, opened 2022-06-16
Fix use of closed network connection errors in agent

rastislavs:
This PR finally resolves the agent's part of the issue mentioned in https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/319

During normal agent operation, error logs about `use of closed network connection` are appearing in the agent's log for each connection, e.g.:

```
E0616 08:55:20.722680       1 client.go:557] "conn write failure" err="write tcp 172.25.0.9:41300->10.240.29.126:443: use of closed network connection" connectionID=2                                                                                                    
E0616 08:55:20.722679       1 client.go:515] "connection read failure" err="read tcp 172.25.0.9:41300->10.240.29.126:443: use of closed network connection" connectionID=2                                                                                                 
```

That makes the production operation hard, as it is not easy to spot / alert on real errors then.

There were 2 PRs in the past attempting to fix it (https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/317 + https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/353), but they were not sufficient and the issue still appears.

The reason is that before this PR, the underlaying connection is closed (`connCtx.conn.Close()`) and blocked Read or Write operations are unblocked and return errors, before the connection is actually removed from the connection manager. This PR makes sure that the underlying connection is closed as the last step of the cleanup, after the connection is removed from the connection manager.

The PR also fixes one misleading log message and adds a new log message about a new connection, so that with the log level verbosity 3 we can now track each connection (now without errors), e.g.:

```
I0616 09:15:15.136798       1 client.go:443] "Proxying new connection" connectionID=7                                                                                                                                                                                
I0616 09:15:48.313330       1 client.go:534] "Exiting proxyToRemote" connectionID=7                                                                                                                                                                                        
I0616 09:15:48.313371       1 client.go:491] "Exiting remoteToProxy" connectionID=7
```

Signed-off-by: Rastislav Szabo <rastislav@kubermatic.com>

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -408,17 +408,15 @@ func (a *Client) Serve() {
 						Type:    client.PacketType_CLOSE_RSP,
 						Payload: &client.Packet_CloseResponse{CloseResponse: &client.CloseResponse{}},
 					}
-
 					closeResp.GetCloseResponse().ConnectID = connID
-					err := connCtx.conn.Close()
-					if err != nil {
-						closeResp.GetCloseResponse().Error = err.Error()
-					}
 					if err := a.Send(closeResp); err != nil {
 						klog.ErrorS(err, "close response failure")
 					}
 					close(dataCh)
 					a.connManager.Delete(connID)
+					if err := connCtx.conn.Close(); err != nil {
+						klog.ErrorS(err, "failed to close connection")
+					}
 				} else {
 					klog.ErrorS(fmt.Errorf("connection is nil"), "cannot send CLOSE_RESP to nil connection")
 				}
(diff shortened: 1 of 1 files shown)


rata (commented):
On a cursory look, LGTM

pkg/agent/client.go
```diff
@@ -442,6 +440,7 @@ func (a *Client) Serve() {
 					klog.ErrorS(err, "could not send dialResp")
 					return
 				}
+				klog.V(2).InfoS("Proxying new connection", "connectionID", connID)
```
jkh52, line 443: looks like debugging (not a v2), I would revert or lower to 3.

Issue #367, opened 2022-07-05
Unsafe early return from ProxyToRemote in Agent


cheftako:
https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/v0.0.32/pkg/agent/client.go#L560 is an unsafe return. If the dataCh channel is full it means that the remoteToProxy method could block at https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/v0.0.32/pkg/agent/client.go#L71. We should ensure that the channel is empty and closed before returning.

PR #368, opened 2022-07-05
Fix proxyToRemote early return to be safe.

cheftako:
Removed unneeded and problematic cleanup call.
Also ensured that dataCh would be consumed before proxyToRemote returned
and goroutine went away. This ensures that serve() is not blocked
forever.

Fixes: https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/367


pkg/agent/client.go
```diff
@@ -534,7 +534,18 @@ func (a *Client) proxyToRemote(connID int64, ctx *connContext) {
 			klog.V(3).InfoS("Exiting proxyToRemote", "connectionID", connID)
 		}
 	}()
-	defer ctx.cleanup()
```
tallclair, line 537: where is cleanup getting called now? Or is it not needed?

cheftako: It is still needed. The normal path where I would expect this to be called is in [remoteToProxy](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/v0.0.32/pkg/agent/client.go#L495). There is also a batch cleanup when the grpc stream ends in [Serve](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/v0.0.32/pkg/agent/client.go#L351). If the other end of the connection closes it we should attempt to close based on the [close request](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/v0.0.32/pkg/agent/client.go#L466).

tallclair: Got it. So those other loops, presumably in other go-routines, are sharing the same connContext and will call `cleanup`, so it's not necessary to call cleanup in this method?

cheftako: Yes.

As it stand we spin up at least two goroutines per connContext, one to run remoteToProxy and one to run proxyToRemote. Today they both attempt to do the cleanup. So it should no be necessary for cleanup to be called here. Given this is the consumer side of the channel being cleaned up, it is arguably wrong for the cleanup to be called here.

PR #369, opened 2022-07-07
WarnOnChannelLimit improvements

tallclair:
1. Use a thread-safe approach for detecting full channels, always enable the warning log, and deprecate the associated flag.
2. Add a `full_receive_channels` gauge metric. The metric is incremented when a send is blocked on a full channel, and decremented when the send goes through successfully.

Any guidance on how to test this would be appreciated.

### DIFF
--- a/cmd/server/app/options/options.go
+++ b/cmd/server/app/options/options.go
@@ -72,13 +72,6 @@ type ProxyRunOptions struct {
 	// it will use the default backend manager to choose a random backend.
 	ProxyStrategies string
 
-	// This controls if we attempt to push onto a "full" transfer channel.
-	// However checking that the transfer channel is full is not safe.
-	// It violates our race condition checking. Adding locks around a potentially
-	// blocking call has its own problems, so it cannot easily be made race condition safe.
-	// The check is an "unlocked" read but is still use at your own peril.
-	WarnOnChannelLimit bool
-
 	// Cipher suites used by the server.
 	// If empty, the default suite will be used from tls.CipherSuites(),
 	// also checks if given comma separated list contains cipher from tls.InsecureCipherSuites().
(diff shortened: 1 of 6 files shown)


pkg/server/server.go
```diff
@@ -686,10 +687,15 @@ func (s *ProxyServer) Connect(stream agent.AgentService_ConnectServer) error {
 				return
 			}
 
-			if s.warnOnChannelLimit && len(recvCh) >= xfrChannelSize {
+			select {
+			case recvCh <- in: // Send didn't block, carry on.
+			default: // Send blocked; record it and try again.
 				klog.V(2).InfoS("Receive channel on Connect is full", "agentID", agentID, "serverID", s.serverID)
```
tallclair, line 693: Note to reviewers: This log looks like it could get pretty spammy. I think there are a few options here:

1. It's not too bad, just leave it.
2. Keep the `--warn-on-channel-limit` flag, to limit the log line
3. Increase the verbosity level on this (e.g. `V(4)`)
4. Add some aggregation / rate limiting on the log line

I don't like the idea of having a flag just to control a spammy log line, so I'm leaning towards one of the other options.

cmd/server/app/options/options.go
```diff
@@ -115,8 +108,11 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.IntVar(&o.KubeconfigBurst, "kubeconfig-burst", o.KubeconfigBurst, "Maximum client burst (proxy server uses this client to authenticate agent tokens).")
 	flags.StringVar(&o.AuthenticationAudience, "authentication-audience", o.AuthenticationAudience, "Expected agent's token authentication audience (used with agent-namespace, agent-service-account, kubeconfig).")
 	flags.StringVar(&o.ProxyStrategies, "proxy-strategies", o.ProxyStrategies, "The list of proxy strategies used by the server to pick a backend/tunnel, available strategies are: default, destHost.")
-	flags.BoolVar(&o.WarnOnChannelLimit, "warn-on-channel-limit", o.WarnOnChannelLimit, "Turns on a warning if the system is going to push to a full channel. The check involves an unsafe read.")
 	flags.StringVar(&o.CipherSuites, "cipher-suites", o.CipherSuites, "The comma separated list of allowed cipher suites. Has no effect on TLS1.3. Empty means allow default list.")
+
+	flags.Bool("warn-on-channel-limit", true, "This behavior is now thread safe and always on. This flag will be removed in a future release.")
```
cheftako, line 113: I'm fine as is. We *could* leave it deprecated but leave it hooked in just to turn off the log message if you are still concerned about the message being spammy.

pkg/server/server_test.go
```diff
@@ -335,32 +337,144 @@ func TestServerProxyNormalClose(t *testing.T) {
...
+			agentConn.EXPECT().Send(closePacket(1)).Return(nil).Times(1),
+			// This extra close is unwanted and should be removed; see
+			// https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/307
+			agentConn.EXPECT().Send(closePacket(0)).Return(nil).Times(1),
+		)
+	}
+	baseServerProxyTestWithBackend(t, validate)
+}
+
+func TestServerProxyRecvChanFull(t *testing.T) {
```
cheftako, line 356: Very nice :D

pkg/server/server.go
cheftako, line 693: I think "on Connect" here is a bit obscure. Might be more clear if we referred to is as "forwarding traffic from the agent"?

pkg/server/server.go
```diff
@@ -399,10 +395,15 @@ func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
 				return
 			}
 
-			if s.warnOnChannelLimit && len(recvCh) >= xfrChannelSize {
+			select {
+			case recvCh <- in: // Send didn't block, carry on.
+			default: // Send blocked; record it and try again.
 				klog.V(2).InfoS("Receive channel on Proxy is full", "userAgent", userAgent, "serverID", s.serverID)
```
cheftako, line 401: I think "on Proxy" here is a bit obscure. Might be more clear if we referred to is as "forwarding traffic from the frontend"?

tallclair:
I update the log messages you comment on, along with similar log messages near by.

Issue #372, opened 2022-08-04
New Tag v0.0.33


ipochi:
Hi @cheftako 

It seems there are important PRs merged since the last tag to be considered for a new tag:

The following PR's have been merged 

#371 from ncopa/go-1.17.12
#368 from cheftako/earlyReturn
#369 from tallclair/warn-on-channel-limit
#364 from rastislavs/fix-agent-conn-close
#365 from rastislavs/fix-server-error-logs

ipochi:
It seems there are few HIGH CVE's in the latest master

```
proxy-agent (gobinary)

Total: 6 (LOW: 0, MEDIUM: 2, HIGH: 4, CRITICAL: 0)

┌─────────────────────────────────────┬────────────────┬──────────┬────────────────────────────────────┬───────────────────────────────────┬──────────────────────────────────────────────────────────────┐
│               Library               │ Vulnerability  │ Severity │         Installed Version          │           Fixed Version           │         Title                             │
├─────────────────────────────────────┼────────────────┼──────────┼────────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────────────────────────┤
│ github.com/prometheus/client_golang │ CVE-2022-21698 │ HIGH     │ v1.7.1                             │ 1.11.1                            │ prometheus/client_golang: Denial of service using            │
│                                     │                │          │                                    │                                   │ InstrumentHandlerCounter                                     │
│                                     │                │          │                                    │                                   │ https://avd.aquasec.com/nvd/cve-2022-21698                   │
├─────────────────────────────────────┼────────────────┼──────────┼────────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────────────────────────┤
│ golang.org/x/net                    │ CVE-2021-33194 │ HIGH     │ v0.0.0-20201110031124-69a78807bb2b │ 0.0.0-20210520170846-37e1c6afe023 │ golang: x/net/html: infinite loop in ParseFragment           │
│                                     │                │          │                                    │                                   │ https://avd.aquasec.com/nvd/cve-2021-33194                   │
├─────────────────────────────────────┼────────────────┼──────────┼────────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────────────────────────┤
[...]

PR #381, opened 2022-08-08
update go modules

ipochi:
Updates the builder image version to 1.18.5 and updates the go modules
to mitigate CVEs. The list of vulnerabilities found in the binaries are
listed in [comment](https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/372#issuecomment-1208124322).

Signed-off-by: Imran Pochi <imranpochi@microsoft.com>


go.mod
```diff
@@ -1,60 +1,74 @@
...
-	google.golang.org/grpc v1.42.0
-	k8s.io/api v0.20.10
-	k8s.io/apimachinery v0.20.10
-	k8s.io/client-go v0.20.10
-	k8s.io/component-base v0.20.10
-	k8s.io/klog/v2 v2.4.0
-	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.0
+	golang.org/x/net v0.0.0-20220805013720-a33c5aa5df48
+	google.golang.org/grpc v1.48.0
+	google.golang.org/protobuf v1.28.1
```
cheftako, line 14: I realize protobug moved. Need to check that there are no additional steps needed when we switch package.

go.mod
```diff
@@ -1,60 +1,74 @@
...
+	k8s.io/api v0.24.3
+	k8s.io/apimachinery v0.24.3
+	k8s.io/client-go v0.24.3
+	k8s.io/component-base v0.24.3
+	k8s.io/klog/v2 v2.70.1
+	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.32
```
tallclair, line 20: Leave the konnectivity-client at `v0.0.0`, since it's using a replace directive.

ipochi:
test pull-apiserver-network-proxy-make-lint

test pull-apiserver-network-proxy-make-test

ipochi:
@tallclair @cheftako 

Finally the lint job succeeds. moved all the components to go v1.18.5. Please review :)

go.mod
```diff
@@ -1,60 +1,74 @@
...
+	google.golang.org/protobuf v1.28.1
+	k8s.io/api v0.24.3
+	k8s.io/apimachinery v0.24.3
+	k8s.io/client-go v0.24.3
+	k8s.io/component-base v0.24.3
+	k8s.io/klog/v2 v2.70.1
 	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.0
 )
 
 require (
```
liggitt, line 23: this added a *lot* of new dependencies, many of which we explicitly don't want to depend on transitively in Kubernetes:

* github.com/go-openapi/...
* new incompatible version of github.com/imdario/mergo (xref https://github.com/kubernetes/kubernetes/issues/107499)

Also, this only updated the root go.mod file, not the client konnectivity-client/go.mod file, so now this repo has two disparate dependency trees. If the client go.mod file was updated the same way this was, we would be unable to pull new versions into kubernetes.

How were the versions of these updated modules selected?

cmd/server/app/server.go
```diff
@@ -350,7 +354,7 @@ func (p *Proxy) runAdminServer(o *options.ProxyRunOptions, server *server.ProxyS
 		}
 	}
 	adminServer := &http.Server{
-		Addr:           net.JoinHostPort(o.AdminBindAddress, strconv.Itoa(o.AdminPort)),
+		Addr:           fmt.Sprintf("127.0.0.1:%d", o.AdminPort),
```
tallclair, line 357: Why was this changed? Looks like maybe a bad rebase?

cheftako: +1. However this was merged Sept 2, over 3 months ago.

go.mod
tallclair: The two that you mentioned are both transitive dependencies through client-go. I don't know how that specific client-go version was selected, but since this PR sat open for a long time I'm guessing it was the latest version when this PR was opened.

We haven't been keeping the konnectivity-client go.mod in sync with this since we're tracking the module versions from the oldest version of k8s we support in the client. We should maybe consider adding a presubmit to ensure that konnectivity-client doesn't take any dependencies on kubernetes staging repos, but hopefully our PR review process would catch that.

https://github.com/kubernetes/client-go/blob/v0.24.3/go.mod#L26

openapi

liggitt: oops, I misread the go-openapi additions, these are the ones we want to avoid:
```
      "github.com/go-openapi/analysis": "use k8s.io/kube-openapi/pkg/validation/spec",
      "github.com/go-openapi/spec": "use k8s.io/kube-openapi/pkg/validation/spec instead",
      "github.com/go-openapi/strfmt": "use k8s.io/kube-openapi/pkg/validation/strfmt instead",
      "github.com/go-openapi/validate": "use k8s.io/kube-openapi/pkg/validation/validate instead",
```

the github.com/imdario/mergo one is a known dependency but the version selected here is known to be problematic (see the linked issue above)

Issue #379, opened 2022-08-05
Flaky Test: TestProxyHandle_SlowContext_GRPC


tallclair:
completed 10000 runs in 1h23m51s with 36 failures (0.36%)

Ran with `stress ./tests.test -test.cpu=4 -test.run=TestProxyHandle_SlowContext_GRPC`

All failures look the same:
```
E0805 02:39:20.068886 3728293 server.go:391] "Stream read from frontend failure" err="rpc error: code = DeadlineExceeded desc = context deadline exceeded" userAgent=[grpc-go/1.42.0] serverID="4ffef7ef-73a0-4538-bb9a-41b7030596ae"
E0805 02:39:20.069106 3728293 client.go:134] "stream read failure" err="rpc error: code = DeadlineExceeded desc = context deadline exceeded"
E0805 02:39:20.069400 3728293 server.go:787] "CLOSE_RSP send to client stream error" err="rpc error: code = DeadlineExceeded desc = context deadline exceeded" serverID="4ffef7ef-73a0-4538-bb9a-41b7030596ae" agentID="17ec1aa3-3eff-4c98-8a84-38c86a481522" connectionID=1
E0805 02:39:20.069438 3728293 server.go:684] "Receive stream from agent read failure" err="rpc error: code = Canceled desc = context canceled"
E0805 02:39:20.069519 3728293 client.go:372] "could not read stream" err="rpc error: code = Unavailable desc = error reading from server: EOF"
--- FAIL: TestProxyHandle_SlowContext_GRPC (4.00s)
    proxy_test.go:234: Expected error when context is cancelled, did not receive error
```

Issue #377, opened 2022-08-05
Flaky Test: TestBasicProxy_GRPC


tallclair:
completed 10000 runs in 22m52s with 123 failures (1.23%)

Ran with `stress ./tests.test -test.cpu=4 -test.run=TestBasicProxy_GRPC`

```
--- FAIL: TestBasicProxy_GRPC (1.46s)
    leaks.go:78: found unexpected goroutines:
        [Goroutine 62 in state select, with sigs.k8s.io/apiserver-network-proxy/konnectivity-client/pkg/client.(*conn).Close on top of the stack:
        goroutine 62 [select]:
        sigs.k8s.io/apiserver-network-proxy/konnectivity-client/pkg/client.(*conn).Close(0xc000205040)
        	/home/stclair/src/github.com/kubernetes-sigs/apiserver-network-proxy/konnectivity-client/pkg/client/conn.go:147 +0x40f
        net/http.(*persistConn).closeLocked(0xc0001165a0, {0x1550020, 0xc000088c50})
        	/home/stclair/sdk/go1.18.1/src/net/http/transport.go:2714 +0x130
        net/http.(*persistConn).close(0xc00004e598?, {0x1550020?, 0xc000088c50?})
        	/home/stclair/sdk/go1.18.1/src/net/http/transport.go:2699 +0xb0
        net/http.(*persistConn).readLoop.func1()
        	/home/stclair/sdk/go1.18.1/src/net/http/transport.go:2061 +0x2d
        net/http.(*persistConn).readLoop(0xc0001165a0)
        	/home/stclair/sdk/go1.18.1/src/net/http/transport.go:2236 +0xacc
        created by net/http.(*Transport).dialConn
        	/home/stclair/sdk/go1.18.1/src/net/http/transport.go:1750 +0x173e
        
         Goroutine 63 in state select, with net/http.(*persistConn).writeLoop on top of the stack:
        goroutine 63 [select]:
        net/http.(*persistConn).writeLoop(0xc0001165a0)
        	/home/stclair/sdk/go1.18.1/src/net/http/transport.go:2392 +0xf5
        created by net/http.(*Transport).dialConn
        	/home/stclair/sdk/go1.18.1/src/net/http/transport.go:1751 +0x1791
        ]
```
```
--- FAIL: TestBasicProxy_GRPC (1.45s)
    leaks.go:78: found unexpected goroutines:
[...]

PR #385, opened 2022-08-16
Dial Timeout testing improvements

tallclair:
- Add a dial timeout test
- Fixes https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/379
- Fixes https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/393
- Introduce a `Done()` function & channel to the gRPC tunnel to indicate when the tunnel has been shutdown (currently only used for testing)

/assign @cheftako

### DIFF
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -36,6 +36,9 @@ type Tunnel interface {
 	// Dial connects to the address on the named network, similar to
 	// what net.Dial does. The only supported protocol is tcp.
 	DialContext(requestCtx context.Context, protocol, address string) (net.Conn, error)
+	// Done returns a channel that is closed when the tunnel is no longer serving any connections,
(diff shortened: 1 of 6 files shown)


tallclair:
Race condition due to setting global `klog` options in a test, when running tests in parallel. Moving the klog settings to `TestMain` fixed it.

tallclair:
Rebased.

tallclair:
Failures are known flakes:
https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/393
https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/377

tallclair:
On closer inspection, https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/393 is just a case of the http.Client timeout being too short, when running with the race detector & a constrained number of CPU threads. I upped the timeout to 30s, and the flakes went away.

Issue #373, opened 2022-08-04
Logs observed in proxy server with connID=0


tallclair:
Examples:

- [server.go:721](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/v0.0.27/pkg/server/server.go#L721) `"DIAL_RSP contains failure" err="dial tcp 10.12.1.4:443: i/o timeout" dialID=8897718490787564857 agentID="dbfdab43-278f-4258-836a-22cbacca1764" connectionID=0`
- [server.go:761](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/v0.0.27/pkg/server/server.go#L761) `"could not get frontend client" err="can't find connID 0 in the frontends[24d90262-6980-476e-9649-88f86404327b]" serverID="70d94c44-f904-466c-b51d-1482564f528f" agentID="24d90262-6980-476e-9649-88f86404327b" connectionID=0`

These logs are from `v0.0.27`.

/assign @cheftako

PR #386, opened 2022-08-17
Handle broken dial in agent.

cheftako:
When the ANP server gets a request broken in dial, it sends a DIAL_CLS.
This allows the random to be sent back and proper close logic invoked.
CLOSE_REQ is a problem because it uses connection id.
During dial the connection is is zero.
Now extending the ANP agent to also send DIAL_CLS for breakage
during a dial request.
Also added some clean up code for related cases.

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -431,7 +431,8 @@ func (a *Client) Serve() {
 				}
 			}
 			go func() {
-				defer close(dialDone)
+				closeOnce := sync.Once{}
+				defer closeOnce.Do(func() { close(dialDone) })
 				start := time.Now()
 				conn, err := net.DialTimeout(dialReq.Protocol, dialReq.Address, dialTimeout)
 				if err != nil {
@@ -440,7 +441,7 @@ func (a *Client) Serve() {
 					if err := a.Send(dialResp); err != nil {
 						klog.ErrorS(err, "could not send dialResp")
 					}
-					// Cannot invoke clean up as we have no function yet.
+					// Cannot invoke clean up as we have no conn yet.
 					return
 				}
 				metrics.Metrics.ObserveDialLatency(time.Since(start))
@@ -451,6 +452,7 @@ func (a *Client) Serve() {
 					klog.ErrorS(err, "could not send dialResp")
 					// clean-up is normally called from remoteToProxy which we will never invoke.
 					// So we are invoking it here to force the clean-up to occur.
+					closeOnce.Do(func() { close(dialDone) })
 					connCtx.cleanup()
 					return
 				}
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -404,13 +404,22 @@ func (a *Client) Serve() {
 				<-dialDone
 				if connCtx.conn != nil {
 					klog.V(4).InfoS("close connection", "connectionID", connID)
-					closeResp := &client.Packet{
-						Type:    client.PacketType_CLOSE_RSP,
-						Payload: &client.Packet_CloseResponse{CloseResponse: &client.CloseResponse{}},
+					var closePkt *client.Packet
+					if connID == 0 {
+						closePkt = &client.Packet{
+							Type:    client.PacketType_DIAL_CLS,
+							Payload: &client.Packet_CloseDial{CloseDial: &client.CloseDial{}},
+						}
+						closePkt.GetCloseDial().Random = dialReq.Random
+					} else {
+						closePkt = &client.Packet{
+							Type:    client.PacketType_CLOSE_RSP,
+							Payload: &client.Packet_CloseResponse{CloseResponse: &client.CloseResponse{}},
+						}
+						closePkt.GetCloseResponse().ConnectID = connID
 					}
-					closeResp.GetCloseResponse().ConnectID = connID
-					if err := a.Send(closeResp); err != nil {
-						klog.ErrorS(err, "close response failure")
+					if err := a.Send(closePkt); err != nil {
+						klog.ErrorS(err, "close response failure", "")
 					}
 					close(dataCh)
 					a.connManager.Delete(connID)
@@ -431,6 +440,7 @@ func (a *Client) Serve() {
 					if err := a.Send(dialResp); err != nil {
 						klog.ErrorS(err, "could not send dialResp")
 					}
+					// Cannot invoke clean up as we have no function yet.
 					return
 				}
 				metrics.Metrics.ObserveDialLatency(time.Since(start))
@@ -439,6 +449,9 @@ func (a *Client) Serve() {
 				dialResp.GetDialResponse().ConnectID = connID
 				if err := a.Send(dialResp); err != nil {
 					klog.ErrorS(err, "could not send dialResp")
+					// clean-up is normally called from remoteToProxy which we will never invoke.
+					// So we are invoking it here to force the clean-up to occur.
+					connCtx.cleanup()
 					return
 				}
 				klog.V(3).InfoS("Proxying new connection", "connectionID", connID)
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -431,8 +431,7 @@ func (a *Client) Serve() {
 				}
 			}
 			go func() {
-				closeOnce := sync.Once{}
-				defer closeOnce.Do(func() { close(dialDone) })
+				defer close(dialDone)
 				start := time.Now()
 				conn, err := net.DialTimeout(dialReq.Protocol, dialReq.Address, dialTimeout)
 				if err != nil {
@@ -452,8 +451,9 @@ func (a *Client) Serve() {
 					klog.ErrorS(err, "could not send dialResp")
 					// clean-up is normally called from remoteToProxy which we will never invoke.
 					// So we are invoking it here to force the clean-up to occur.
-					closeOnce.Do(func() { close(dialDone) })
-					connCtx.cleanup()
+					// However, cleanup will block until dialDone is closed.
+					// So placing cleanup on its own goroutine to wait for the deferred close(dialDone) to kick in.
+					go connCtx.cleanup()
 					return
 				}
 				klog.V(3).InfoS("Proxying new connection", "connectionID", connID)
--- a/pkg/agent/client_test.go
+++ b/pkg/agent/client_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 	"time"
 
+	"go.uber.org/goleak"
 	"google.golang.org/grpc"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/apiserver-network-proxy/konnectivity-client/proto/client"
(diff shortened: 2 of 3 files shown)


cheftako:
Fixes https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/373

pkg/agent/client.go
```diff
@@ -431,6 +440,7 @@ func (a *Client) Serve() {
 					if err := a.Send(dialResp); err != nil {
 						klog.ErrorS(err, "could not send dialResp")
 					}
+					// Cannot invoke clean up as we have no function yet.
```
cheftako, line 443: Cannot invoke clean up as we have no conn yet.

pkg/agent/client.go
```diff
@@ -439,6 +449,9 @@ func (a *Client) Serve() {
 				dialResp.GetDialResponse().ConnectID = connID
 				if err := a.Send(dialResp); err != nil {
 					klog.ErrorS(err, "could not send dialResp")
+					// clean-up is normally called from remoteToProxy which we will never invoke.
+					// So we are invoking it here to force the clean-up to occur.
+					connCtx.cleanup()
```
cheftako, line 454: Need to close dialDone *before* calling cleanup, right now we're calling it after.

pkg/agent/client.go
```diff
@@ -439,6 +450,10 @@ func (a *Client) Serve() {
 				dialResp.GetDialResponse().ConnectID = connID
 				if err := a.Send(dialResp); err != nil {
 					klog.ErrorS(err, "could not send dialResp")
+					// clean-up is normally called from remoteToProxy which we will never invoke.
+					// So we are invoking it here to force the clean-up to occur.
+					closeOnce.Do(func() { close(dialDone) })
```
tallclair, line 455: What is the point of `dialDone`? Can we just get rid of it?

It looks like cleanup can be called in 3 places:
1. Here, in the same go routine (doesn't need the channel)
2. `remoteToProxy` - likely started after the channel is closed
3. From a CLOSE_REQ, but a close_req for the connectionID could only be received after the DIAL_RSP was already sent?

pkg/server/server.go
```diff
@@ -738,14 +740,27 @@ func (s *ProxyServer) serveRecvBackend(backend Backend, stream agent.AgentServic
...
+					// We already attempted to tell the frontend that. We should ensure we tell the backend.
+					pkt := &client.Packet{
+						Type: client.PacketType_CLOSE_REQ,
+						Payload: &client.Packet_CloseRequest{
+							CloseRequest: &client.CloseRequest{
+								ConnectID: resp.ConnectID,
+							},
+						},
+					}
+					_ = stream.Send(pkt)
```
tallclair, line 763: I think it's worth logging the error here if it happens?

pkg/server/server.go
```diff
@@ -738,14 +740,27 @@ func (s *ProxyServer) serveRecvBackend(backend Backend, stream agent.AgentServic
...
 					dialErr = true
 				}
 				// Avoid adding the frontend if there was an error dialing the destination
```
tallclair, line 766: Doesn't the break below break out of the whole Receive loop, and drop all connections to the agent? That doesn't seem right....

tallclair: I think it should be a continue? That seems like a pretty major bug...

tallclair (commented):
Still reviewing, but wanted to send comments so far.

pkg/server/server.go
tallclair: Silly me... I forgot that `break` exits a switch statement, not just loops :facepalm:

pkg/agent/client.go
cheftako: cleanup is also called from https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/agent/client.go#L351.

pkg/server/server.go
cheftako: Added a error log a v==5. 
I think its probably worth having for debugging.
However I don't want to exacerbate the alarm.
The earlier "DIAL_RSP send to frontend stream failure" is the more useful and pertinent error.

pkg/agent/client_test.go
```diff
@@ -195,6 +206,86 @@ func TestClose_Client(t *testing.T) {
...
+		t.Log(">>>> Request Received") // FIXME
+		fmt.Fprint(w, "hello, world")
+	}))
+	defer ts.Close()
+
+	func() {
+		defer goleakVerifyNone(t, goleak.IgnoreCurrent())
+
+		// Stimulate sending KAS DIAL_REQ to (Agent) Client
+		dialPacket := newDialPacket("tcp", ts.URL[len("http://"):], 111)
```
tallclair, line 271: nit: prefer
```suggestion
		dialPacket := newDialPacket("tcp", strings.TrimPrefix(ts.URL, "http://"), 111)
```

pkg/server/server.go
```diff
@@ -738,14 +740,27 @@ func (s *ProxyServer) serveRecvBackend(backend Backend, stream agent.AgentServic
 			} else {
```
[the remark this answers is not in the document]
cheftako: Currently the server does not timeout pending dials. It assumes either the frontend or backend will take care of that. So for this to happen, I think you probably have to have a race between the backend sending the DIAL_RSP (which is must have lost) and the frontend sending a DIAL_CLS (which one). The DIAL_RSP could not send a CLOSE_REQ as it did not yet know the connection ID. 

So in short, yes we probably should.

pkg/server/server.go
```diff
@@ -759,6 +774,22 @@ func (s *ProxyServer) serveRecvBackend(backend Backend, stream agent.AgentServic
...
 
+		case client.PacketType_DIAL_CLS:
+			resp := pkt.GetCloseDial()
+			klog.V(5).InfoS("Received DIAL_CLS", "serverID", s.serverID, "agentID", agentID, "dialID", resp.Random)
+			if frontend, ok := s.PendingDial.Get(resp.Random); !ok {
+				klog.V(2).InfoS("DIAL_CLS not recognized; dropped", "dialID", resp.Random, "agentID", agentID)
+			} else {
+				if err := frontend.send(pkt); err != nil {
+					// Normal when frontend closes it.
+					klog.ErrorS(err, "DIAL_CLS send to client stream error", "serverID", s.serverID, "agentID", agentID, "dialID", resp.Random)
```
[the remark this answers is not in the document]
cheftako: So this is a DIAL_CLS we received from the backend during a dial request. In most cases I do not thing the DIAL_CLS has anything to do with the FE. So my thought was that this was an indication of some additional error, so log normally. We could change this to V(2) or V(3) if you think that would be better. I don't think this should be V(5).

pkg/server/server.go
```diff
@@ -759,6 +774,22 @@ func (s *ProxyServer) serveRecvBackend(backend Backend, stream agent.AgentServic
...
+				} else {
+					klog.V(5).InfoS("DIAL_CLS sent to frontend", "dialID", resp.Random)
+				}
+				s.PendingDial.Remove(resp.Random)
+				klog.V(5).InfoS("Close streaming", "agentID", agentID, "dialID", resp.Random)
+			}
+
 		case client.PacketType_DATA:
 			resp := pkt.GetData()
 			klog.V(5).InfoS("Received data from agent", "bytes", len(resp.Data), "agentID", agentID, "connectionID", resp.ConnectID)
```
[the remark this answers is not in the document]
cheftako: For us to be unable to get the frontend from connID then one of the following must have happened.
1) the defer at https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/server/server.go#L716 mus have been called. However we can't be in this bit of code if that were true.
2) We got a CLOSE_RSP from the backend at https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/server/server.go#L791. That suggests packets are out of order but also that there is no need to send a CLOSE_REQ.
3) The agent sent an invalid Connection ID. Not sure what we should do here. Not sure what sending a CLOSE_REQ would do...
4) We somehow "lost" the frontend reference due to a ?bug. Arguably sending the CLOSE_REQ might help here. However I am uncomfortable reacting to theoretical bug we haven't seen yet and shouldn't happen. If we want to do something here I think I would rather add a metric which counted and exported unrecognized packets.

pkg/server/server.go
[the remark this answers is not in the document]
cheftako: Done

pkg/agent/client_test.go
```diff
@@ -195,6 +206,86 @@ func TestClose_Client(t *testing.T) {
...
+		AgentService_ConnectClient: clientStream,
+		sendErr:                    errors.New("expected send error"),
+	}
+
+	// Start agent
+	go testClient.Serve()
+
+	// Start test http server as remote service
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		t.Log(">>>> Request Received") // FIXME
```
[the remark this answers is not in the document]
cheftako: Done

pkg/server/server.go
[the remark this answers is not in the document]
tallclair: If that's the case, is the comment here (line 779) accurate? Or was that juts copied from elsewhere?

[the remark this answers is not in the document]
cheftako: Copied. Sorry will clean up.

pkg/agent/client.go
```diff
@@ -439,6 +450,10 @@ func (a *Client) Serve() {
...
+					connCtx.cleanup()
```
tallclair, line 456: Another way we could handle this is to just call `go connCtx.cleanup()` (and go back to the regular deferred close) in a go-routine. I'm not sure that's better, I just find the `closeOnce.Do(func() { close(dialDone) })` aversive, but that might just be aesthetics...

tallclair (commented):
/lgtm
/hold

For last comment. Feel free to unhold if you don't want to do that.

pkg/agent/client.go
cheftako: Made the switch but also added a comment explaining why we have the extra goroutine as  don't think its obvious on first read through.

cheftako:
A little confused why our arm64 build image seems to be missing protoc.

```
Docker in Docker enabled, initializing...
================================================================================
Starting Docker: docker.
Waiting for docker to be ready, sleeping for 1 seconds.
================================================================================
Done setting up docker in docker.
+ WRAPPED_COMMAND_PID=160
+ wait 160
+ make docker-build/proxy-agent-arm64
make ARCH=arm64 docker-build/proxy-agent
make[1]: Entering directory '/home/prow/go/src/sigs.k8s.io/apiserver-network-proxy'
mkdir -p /home/prow/go/src
protoc -I . proto/agent/agent.proto --go_out=plugins=grpc:/home/prow/go/src
make[1]: *** [Makefile:115: proto/agent/agent.pb.go] Error 1[2](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_apiserver-network-proxy/386/pull-apiserver-network-proxy-docker-build-arm64/1561756513230393344#1:build-log.txt%3A2)7
make: *** [Makefile:228: docker-build/proxy-agent-arm6[4](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_apiserver-network-proxy/386/pull-apiserver-network-proxy-docker-build-arm64/1561756513230393344#1:build-log.txt%3A4)] Error 2
```

pkg/agent/client.go
```diff
@@ -404,13 +404,22 @@ func (a *Client) Serve() {
 				<-dialDone
 				if connCtx.conn != nil {
 					klog.V(4).InfoS("close connection", "connectionID", connID)
-					closeResp := &client.Packet{
-						Type:    client.PacketType_CLOSE_RSP,
-						Payload: &client.Packet_CloseResponse{CloseResponse: &client.CloseResponse{}},
+					var closePkt *client.Packet
+					if connID == 0 {
```
jkh52, line 408: isn't this impossible? Given line 394 I would expect connID to always be at least 1.

tallclair: Good point. The intent here is that if the dial response hasn't been sent, then the server isn't aware of the connection ID. In general, our close semantics are leaky. We really need something like a `DIAL_ACK` to confirm that the connection ID was "registered". Better yet, if we got rid of the dialID / connectionID separation and just had a single connection ID initiated at the get go than we could simplify a lot of this... Unfortunately no one seems to remember why they were separated in the first place.

PR #389, opened 2022-08-23
[konnectivity-client] Expose dial failure reasons

tallclair:
Expose a machine-readable `DialFailureReason` through a `GetDialFailureReason` function, to surface different sources of dial failures to the caller of `DialContext`.

The motivation for this change is to separate out `proxy` dial failure reasons in the kube-apiserver.

/assign @cheftako


tallclair:
This isn't the most elegant solution, so let me know if you have a better idea...

Issue #388, opened 2022-08-18
proxy-server KeepAlive EnforcementPolicy mismatch


tallclair:
https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/279/files added a konnectivity-agent use of KeepAlive [1], meaning if it doesn't see any activity for in the configured window (default 1 hour) it will send a KeepAlive ping to the server. The agent also sets PermitWithoutStream to true, meaning it will send a KeepAlive even if there aren't any active rpc streams.

On the proxy-server side, we don't set a KeepAlive EnforcementPolicy [2], meaning it gets the default values: don't permit without stream, and a minimum period of 5 minutes.

The result of this mismatch is that if an agent isn't receiving any activity, it will ping the server more frequently than allowed, and after 2 "violations" it will forcibly shut down the client connection, resulting in logs like:

```
"could not read stream" err="rpc error: code = Unavailable desc = closing transport due to: connection error: desc = \"error reading from server: EOF\", received prior goaway: code: ENHANCE_YOUR_CALM, debug data: \"too_many_pings\""
```

At a minimum, we should update the policy to set `PermitWithoutStream=true`. We may also want to consider lowering the lowering the minimum period and/or expose it as a configuration parameter.

[1]: https://pkg.go.dev/google.golang.org/grpc@v1.48.0/keepalive#ClientParameters
[2]: https://pkg.go.dev/google.golang.org/grpc@v1.48.0/keepalive#EnforcementPolicy

PR #390, opened 2022-08-24
Adjust proxy-server agent KeepaliveEnforcementPolicy

tallclair:
The agent currently sets `PermitWithoutStream` on its KeepAlive strategy, but the serve enforcement policy uses the default (`PermitWithoutStream=false`). This mismatch could cause the server to forcibly disconnect the agent when it starts sending keepalive pings.

Also, lower the minimum KeepAlive ping time from the default 5 minutes to 30 seconds, to allow more aggressive agent KeepAlive strategies.

Fixes https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/388

/assign @cheftako

### DIFF
--- a/cmd/server/app/server.go
+++ b/cmd/server/app/server.go
@@ -16,6 +16,7 @@ import (
 	"strings"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


cmd/server/app/server.go
```diff
@@ -321,6 +322,10 @@ func (p *Proxy) runAgentServer(o *options.ProxyRunOptions, server *server.ProxyS
 	agentServerOptions := []grpc.ServerOption{
 		grpc.Creds(credentials.NewTLS(tlsConfig)),
 		grpc.KeepaliveParams(keepalive.ServerParameters{Time: o.KeepaliveTime}),
+		grpc.KeepaliveEnforcementPolicy(keepalive.EnforcementPolicy{
```
cheftako, line 325: I realize that we don't tend to keep frontend connections open as long, but that doesn't mean it doesn't happen. Should have a similar fix as https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/cmd/server/app/server.go#L188?

cheftako: From tallclair "I don't think PermitWithoutStream makes sense as long as it's a single use tunnel, so I'm inclined to just keep the defaults..." also that the current client frontend does not set a client keep alive, so it should not be needed (yet).

PR #395, opened 2022-09-02
[konnectivity-client] Fix GetDialFailureReason

tallclair:
The way I called `errors.As` was incorrect, leading to a nil-pointer panic. I fixed that, and added unit tests.


tallclair:
Dropped the second commit, since it's a bit more complicated, and I want to get the nil-pointer fix in first.

PR #396, opened 2022-09-02
Improve agentID logging

tallclair:
Include the agent ID when fetching a backend, and log the agent ID in relevant locations.

### DIFF
--- a/pkg/server/desthost_backend_manager.go
+++ b/pkg/server/desthost_backend_manager.go
@@ -36,19 +36,19 @@ func NewDestHostBackendManager() *DestHostBackendManager {
 }
 
 // Backend tries to get a backend associating to the request destination host.
-func (dibm *DestHostBackendManager) Backend(ctx context.Context) (Backend, error) {
+func (dibm *DestHostBackendManager) Backend(ctx context.Context) (Backend, string, error) {
 	dibm.mu.RLock()
 	defer dibm.mu.RUnlock()
 	if len(dibm.backends) == 0 {
-		return nil, &ErrNotFound{}
+		return nil, "", &ErrNotFound{}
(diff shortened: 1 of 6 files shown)


pkg/server/desthost_backend_manager.go
```diff
@@ -36,19 +36,19 @@ func NewDestHostBackendManager() *DestHostBackendManager {
...
-		return nil, &ErrNotFound{}
+		return nil, "", &ErrNotFound{}
 	}
 	destHost := ctx.Value(destHost).(string)
 	if destHost != "" {
 		bes, exist := dibm.backends[destHost]
 		if exist && len(bes) > 0 {
 			klog.V(5).InfoS("Get the backend through the DestHostBackendManager", "destHost", destHost)
-			return dibm.backends[destHost][0], nil
+			return dibm.backends[destHost][0], destHost, nil
```
cheftako, line 50: Using the agentID as the backend identifier seems reasonable. If we do that I think we should always make it what we return. agentID and destHost are not easily comparable but either the system or anyone examining the logs.

pkg/server/server.go
```diff
@@ -430,7 +433,7 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
 			// the address, then we can send the Dial_REQ to the
 			// same agent. That way we save the agent from creating
 			// a new connection to the address.
-			backend, err = s.getBackend(pkt.GetDialRequest().Address)
+			backend, agentID, err = s.getBackend(pkt.GetDialRequest().Address)
```
cheftako, line 436: Here is an example of why I do not think we should return a destHost as a backed identifier.

PR #398, opened 2022-09-03
[konnectivity-client] Ensure grpc tunnel is closed on dial failure

tallclair:
3 commits:
1. Handle DIAL_CLS packets in the client - treat it similarly to a DIAL_RSP with a failure
2. Send a best-effort DIAL_CLS to the server if the client times out (or the request is canceled)
3. Explicitly close the client connection and stop serving the tunnel on dial failures

**Note to reviewers:** I'm hoping to backport this to older Kubernetes versions, so consider backwards compatibility (both with older k8s, as well as older proxy-servers). I think it's OK, but would appreciate extra attention on this detail.

/assign @cheftako

### DIFF
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -320,7 +320,11 @@ func (t *grpcTunnel) DialContext(requestCtx context.Context, protocol, address s
 
 	klog.V(5).Infoln("DIAL_REQ sent to proxy server")
 
-	c := &conn{stream: t.stream, random: random}
+	c := &conn{
+		stream:      t.stream,
+		random:      random,
+		closeTunnel: t.closeTunnel,
+	}
 
 	select {
 	case res := <-resCh:
@@ -366,7 +370,10 @@ func (t *grpcTunnel) closeDial(dialID int64) {
 	if err := t.stream.Send(req); err != nil {
 		klog.V(5).InfoS("Failed to send DIAL_CLS", "err", err, "dialID", dialID)
 	}
+	t.closeTunnel()
+}
 
+func (t *grpcTunnel) closeTunnel() {
 	atomic.StoreUint32(&t.closing, 1)
 	t.clientConn.Close()
 }
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -327,9 +327,11 @@ func (t *grpcTunnel) DialContext(requestCtx context.Context, protocol, address s
 		t.connsLock.Unlock()
 	case <-time.After(30 * time.Second):
 		klog.V(5).InfoS("Timed out waiting for DialResp", "dialID", random)
+		go t.closeDial(random)
 		return nil, &dialFailure{"dial timeout, backstop", DialFailureTimeout}
 	case <-requestCtx.Done():
 		klog.V(5).InfoS("Context canceled waiting for DialResp", "ctxErr", requestCtx.Err(), "dialID", random)
+		go t.closeDial(random)
 		return nil, &dialFailure{"dial timeout, context", DialFailureContext}
 	}
 
@@ -338,6 +340,21 @@ func (t *grpcTunnel) DialContext(requestCtx context.Context, protocol, address s
 
 func (t *grpcTunnel) Done() <-chan struct{} {
 	return t.done
+}
+
+// Send a best-effort DIAL_CLS request for the given dial ID.
+func (t *grpcTunnel) closeDial(dialID int64) {
+	req := &client.Packet{
+		Type: client.PacketType_DIAL_CLS,
+		Payload: &client.Packet_CloseDial{
+			CloseDial: &client.CloseDial{
+				Random: dialID,
+			},
+		},
+	}
+	if err := t.stream.Send(req); err != nil {
+		klog.V(5).InfoS("Failed to send DIAL_CLS", "err", err, "dialID", dialID)
+	}
 }
 
 func GetDialFailureReason(err error) (isDialFailure bool, reason DialFailureReason) {
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -42,7 +42,7 @@ type Tunnel interface {
 }
 
 type dialResult struct {
-	err    string
+	err    *dialFailure
 	connid int64
 }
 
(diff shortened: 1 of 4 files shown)


konnectivity-client/pkg/client/client.go
```diff
@@ -189,6 +192,34 @@ func (t *grpcTunnel) serve(tunnelCtx context.Context, c clientConn) {
 				return
 			}
 
+		case client.PacketType_DIAL_CLS:
+			resp := pkt.GetCloseDial()
+			t.pendingDialLock.RLock()
```
cheftako, line 205: Please use defer and anonymous even you want to limit scope.

konnectivity-client/pkg/client/client_test.go
```diff
@@ -479,7 +529,9 @@ func (s *proxyServer) serve() {
 			return
 		}
 
+		s.packetsLock.Lock()
```
cheftako, line 569: Less important but still defer on unlock

konnectivity-client/pkg/client/client_test.go
```diff
@@ -540,3 +595,13 @@ func (s *proxyServer) handleData(pkt *client.Packet) *client.Packet {
 		},
 	}
 }
+
+// Override goleakVerifyNone to set t.Helper.
+// TODO: delete this once goleak has been updated to include
+// https://github.com/uber-go/goleak/commit/2dfebe88ddf19de216c4ab15a1189fc640b1ea9f
+func goleakVerifyNone(t *testing.T, options ...goleak.Option) {
```
cheftako, line 639: Consider making this part of a util in konn-client.

konnectivity-client/pkg/client/client.go
```diff
@@ -340,6 +342,21 @@ func (t *grpcTunnel) Done() <-chan struct{} {
 	return t.done
 }
 
+// Send a best-effort DIAL_CLS request for the given dial ID.
+func (t *grpcTunnel) closeDial(dialID int64) {
```
cheftako, line 357: sendCloseDial?

cheftako: Also make sure we do not multi send based on our impl of net.Conn.Close()

konnectivity-client/pkg/client/client.go
```diff
@@ -141,7 +149,7 @@ func (t *grpcTunnel) serve(tunnelCtx context.Context, c clientConn) {
 
 	for {
 		pkt, err := t.stream.Recv()
-		if err == io.EOF {
+		if err == io.EOF || t.isClosing() {
```
cheftako, line 152: I've had problems with not draining connections before exiting. Can we ensure we do not need to drain the stream before exiting?

konnectivity-client/pkg/client/client.go
```diff
@@ -333,6 +341,9 @@ func (t *grpcTunnel) DialContext(requestCtx context.Context, protocol, address s
 		klog.V(5).InfoS("Context canceled waiting for DialResp", "ctxErr", requestCtx.Err(), "dialID", random)
 		go t.closeDial(random)
 		return nil, &dialFailure{"dial timeout, context", DialFailureContext}
+	case <-t.done:
```
cheftako, line 344: Do we really need 3 paths for we're shutting down?

konnectivity-client/pkg/client/client.go
```diff
@@ -355,6 +366,13 @@ func (t *grpcTunnel) closeDial(dialID int64) {
 	if err := t.stream.Send(req); err != nil {
 		klog.V(5).InfoS("Failed to send DIAL_CLS", "err", err, "dialID", dialID)
 	}
+
+	atomic.StoreUint32(&t.closing, 1)
```
cheftako, line 370: Also should handle this in our impl of net.Conn.Close()

konnectivity-client/pkg/client/client.go
tallclair: > Also make sure we do not multi send based on our impl of net.Conn.Close()

We only create a `conn{}` if the dial is successful, and this function is only ever called if the dial fails, so this should never happen. That said, if we did send 2 `DIAL_CLS` packets, the second one should be a no-op on the server.

konnectivity-client/pkg/client/client.go
tallclair: Unfortunately yes, I don't see a clean way to consolidate these. The timeout & context canceled cases could easily be collapsed, but then we wouldn't be able to distinguish between them in the logs. This `<-t.done` case is not strictly necessary (the timeout case would be hit eventually), but it makes for a cleaner shutdown if the underlying connection gets closed prematurely (outside of konnectivity-client).

konnectivity-client/pkg/client/client.go
tallclair: From the documentation on the underlying gRPC method:

> 	// RecvMsg blocks until it receives a message into m or the stream is
	// done. It returns io.EOF when the stream completes successfully. On
	// any other error, the stream is aborted and the error contains the RPC
	// status.

I interpret that as EOF will only be returned if it doesn't return a different error (i.e. doesn't need to be drained). I also don't see us draining the stream anywhere else, nor do the gRPC docs mention "drain" anywhere.

konnectivity-client/pkg/client/client_test.go
tallclair: Ack. Will follow up in a separate PR.

tallclair:
Addressed feedback. Also added a commit to to ensure the tunnel is always closed when the `net.Conn` is closed (fixing another potential leak)

konnectivity-client/pkg/client/client.go
```diff
@@ -53,13 +54,70 @@ type pendingDial struct {
...
+}
+
+func (p *pendingDialManager) get(dialID int64) (pendingDial, bool) {
+	p.mutex.RLock()
+	defer p.mutex.RUnlock()
+	pd, ok := p.pendingDials[dialID]
+	return pd, ok
+}
+
+// TODO: Replace with a generic implementation once it is safe to assume the client is built with go1.18+
```
cheftako, line 82: Given this is in konnectivity-client, I believe we need to wait for K/K 1.27.

konnectivity-client/pkg/client/client.go
```diff
@@ -154,28 +214,29 @@ func (t *grpcTunnel) serve(tunnelCtx context.Context, c clientConn) {
...
 		case client.PacketType_DIAL_RSP:
 			resp := pkt.GetDialResponse()
-			t.pendingDialLock.RLock()
-			pendingDial, ok := t.pendingDial[resp.Random]
-			t.pendingDialLock.RUnlock()
+			pendingDial, ok := t.pendingDial.get(resp.Random)
 
 			if !ok {
+				// If the DIAL_RSP does not match a pending dial, it means one of two things:
+				//   1. There was a second DIAL_RSP for the connection request (this is very unlikely but possible)
```
cheftako, line 221: I don't think we should ever get two DIAL_RSP for a pending dial. Collisions shouldn't be possible given the client sets req.Random. Maybe this could happen if the server had a race between sending the DIAL_RESP and a DIAL_CLS on something like a timeout?

cheftako: Almost the inverse of #1 below.

konnectivity-client/pkg/client/client.go
tallclair: Yep :+1:

konnectivity-client/pkg/client/client.go
tallclair: Yeah, I was also confused by this comment. The comment was there before (moved to here), but I can't think of when this would occur. I'll update it.

PR #400, opened 2022-09-15
Dial metrics

tallclair:
Introduce a new server-side metric to track dial failures, as observed from the proxy-server. Also introduce a new metrics-testing library (which I spent an embarrassing amount of time on).

### DIFF
--- a/pkg/server/metrics/metrics.go
+++ b/pkg/server/metrics/metrics.go
@@ -50,6 +50,7 @@ type ServerMetrics struct {
 	backend           *prometheus.GaugeVec
 	pendingDials      *prometheus.GaugeVec
 	fullRecvChannels  *prometheus.GaugeVec
+	dialFailures      *prometheus.CounterVec
 }
 
 // newServerMetrics create a new ServerMetrics, configured with default metric names.
@@ -122,6 +123,17 @@ func newServerMetrics() *ServerMetrics {
 			"service_method",
 		},
 	)
+	dialFailures := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "dial_failure_count",
+			Help:      "Number of dial failures observed. Multiple failures can occur for a single dial request.",
+		},
+		[]string{
+			"reason",
+		},
+	)
 
 	prometheus.MustRegister(latencies)
 	prometheus.MustRegister(frontendLatencies)
@@ -130,6 +142,7 @@ func newServerMetrics() *ServerMetrics {
 	prometheus.MustRegister(backend)
 	prometheus.MustRegister(pendingDials)
 	prometheus.MustRegister(fullRecvChannels)
+	prometheus.MustRegister(dialFailures)
 	return &ServerMetrics{
 		latencies:         latencies,
 		frontendLatencies: frontendLatencies,
@@ -138,6 +151,7 @@ func newServerMetrics() *ServerMetrics {
 		backend:           backend,
 		pendingDials:      pendingDials,
 		fullRecvChannels:  fullRecvChannels,
+		dialFailures:      dialFailures,
 	}
 }
 
@@ -186,4 +200,18 @@ func (a *ServerMetrics) SetPendingDialCount(count int) {
 // FullRecvChannel retrieves the metric for counting full receive channels.
 func (a *ServerMetrics) FullRecvChannel(serviceMethod string) prometheus.Gauge {
 	return a.fullRecvChannels.With(prometheus.Labels{"service_method": serviceMethod})
+}
+
+type DialFailureReason string
+
+const (
+	DialFailureEndpoint             DialFailureReason = "endpoint"              // Dial failure reported by the agent back to the server.
+	DialFailureUnrecognizedResponse DialFailureReason = "unrecognized_response" // Dial repsonse received for unrecognozide dial ID.
+	DialFailureSend                 DialFailureReason = "send"                  // Successful dial response from agent, but failed to send to frontend.
+	DialFailureBackendClose         DialFailureReason = "backend_close"         // Received a DIAL_CLS from the backend before the dial completed.
+	DialFailureFrontendClose        DialFailureReason = "frontend_close"        // Received a DIAL_CLS from the frontend before the dial completed.
+)
+
+func (a *ServerMetrics) ObserveDialFailure(reason DialFailureReason) {
+	a.dialFailures.With(prometheus.Labels{"reason": string(reason)}).Inc()
 }
--- a/pkg/server/metrics/metrics.go
+++ b/pkg/server/metrics/metrics.go
@@ -168,7 +168,6 @@ func (a *ServerMetrics) Reset() {
 	a.latencies.Reset()
 	a.frontendLatencies.Reset()
 	a.connections.Reset()
-	a.httpConnections.Set(0)
 	a.backend.Reset()
 	a.pendingDials.Reset()
 	a.fullRecvChannels.Reset()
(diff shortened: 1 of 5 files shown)


pkg/server/metrics/metrics.go
```diff
@@ -187,3 +215,17 @@ func (a *ServerMetrics) SetPendingDialCount(count int) {
 func (a *ServerMetrics) FullRecvChannel(serviceMethod string) prometheus.Gauge {
 	return a.fullRecvChannels.With(prometheus.Labels{"service_method": serviceMethod})
 }
+
+type DialFailureReason string
+
+const (
+	DialFailureEndpoint             DialFailureReason = "endpoint"              // Dial failure reported by the agent back to the server.
```
andrewsykim, line 222: nit: `DialFailureAgent` or `DialFailureBackend` might be slightly clearer IMO

pkg/server/metrics/metrics.go
```diff
@@ -187,3 +215,17 @@ func (a *ServerMetrics) SetPendingDialCount(count int) {
...
+	DialFailureUnrecognizedResponse DialFailureReason = "unrecognized_response" // Dial repsonse received for unrecognozide dial ID.
+	DialFailureSend                 DialFailureReason = "send"                  // Successful dial response from agent, but failed to send to frontend.
```
andrewsykim, line 224: Thinking about how specific we want to be with the reasons. For e.g. would `DialFailureSendResponseToFrontend` make things clearer or just adding too much verbosity?

pkg/server/metrics/metrics.go
andrewsykim: Or `DialFailureResponseFromAgent`?

pkg/server/metrics/metrics.go
```diff
@@ -187,3 +215,17 @@ func (a *ServerMetrics) SetPendingDialCount(count int) {
...
+	DialFailureBackendClose         DialFailureReason = "backend_close"         // Received a DIAL_CLS from the backend before the dial completed.
+	DialFailureFrontendClose        DialFailureReason = "frontend_close"        // Received a DIAL_CLS from the frontend before the dial completed.
```
andrewsykim, line 226: Looking at how `DialFailureFrontendClose` and `DialFailureBackendClose` is used, it seems like handling DIAL_CLS while a dial is pending is expected in some cases if either ends of the stream want to close the connection. Wondering if labeling this as a "failure" in the metrics could be misleading?

pkg/server/metrics/metrics.go
```diff
@@ -87,41 +96,52 @@ func newServerMetrics() *ServerMetrics {
...
 	)
+	dialFailures := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: Namespace,
+			Subsystem: Subsystem,
+			Name:      DialFailuresMetric,
+			Help:      "Number of dial failures observed. Multiple failures can occur for a single dial request.",
+		},
+		[]string{
+			"reason",
```
andrewsykim, line 142: Thoughts on adding `serverID` and `agentID` here?

Maybe `serverID` is redundant, because it always refers to the current proxy server publishing the metrics anyways. 

I guess with `agentID` we run the risk of increasing metrics cardinality because there's no limit on # of agents you could end up with, but generally it's not a huge number. But it would really help identify failure modes where there's a single bad agent.

andrewsykim (commented):
Overall LGTM, just left some minor comments around naming and labels.

(deferring approval to @cheftako since I'm still catching up from a long break)

pkg/server/metrics/metrics.go
cheftako: For a cloud provider the cardinality issue is huge. For a cluster the agentID and serverID may only be a handful. Across all clusters we expect no repeats, so this number gets very large, very quickly. In addition over time these IDs change (updates, restarts, ...) so again it grows without bound.

pkg/server/metrics/metrics.go
```diff
@@ -138,13 +159,20 @@ func newServerMetrics() *ServerMetrics {
...
+		dialFailures:      dialFailures,
 	}
 }
 
 // Reset resets the metrics.
 func (a *ServerMetrics) Reset() {
 	a.latencies.Reset()
 	a.frontendLatencies.Reset()
+	a.connections.Reset()
+	a.httpConnections.Set(0)
```
cheftako, line 171: This is a gauge, why are we setting this to 0?

pkg/server/server.go
```diff
@@ -26,14 +26,15 @@ import (
 	"sync"
 	"time"
 
+	runpprof "runtime/pprof"
```
cheftako, line 29: Seems like this should actually be between "io" and "strconv".

pkg/server/metrics/metrics.go
tallclair: Yeah, I agree with @cheftako. I think this is better tackled through logging.

pkg/server/metrics/metrics.go
tallclair: Eh. I set it to 0 since it doesn't have a `Reset()` method. But I think I agree that doesn't really make sense. Removed.

pkg/server/metrics/metrics.go
tallclair: How about `DialFailureSendResponse = "send_rsp"` as a compromise? The server shouldn't ever be sending a DIAL_RSP anywhere else, so I think frontend is redundant here?

pkg/server/metrics/metrics.go
tallclair: I agree it's a little funky. However, from the frontend perspective this is a dial failure either way (https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/konnectivity-client/pkg/client/client.go#L266).

Now, should this be a dial failure from the server side? The way I'm interpreting it is the DIAL_REQ wasn't completed with a successful DIAL_RSP, so it's a failure.

I think that's all semantic details though. In practice, we want to know how many times a dial request failed to complete successfully.

Make sense?

tallclair: Also, if you have a suggestion for a better word than `failure`, I'm open to renaming these.

pkg/server/server.go
tallclair: oh yeah, I think my editor did this automatically...

pkg/server/metrics/metrics.go
tallclair: How about `ErrorResponse`? Since it's literally a `DIAL_RSP` with an error in it? I can include `...FromAgent` if you prefer.

andrewsykim: +1 to `DialResponseError` or similar.

pkg/server/metrics/metrics.go
andrewsykim: sgtm

pkg/server/metrics/metrics.go
andrewsykim: > Also, if you have a suggestion for a better word than failure, I'm open to renaming these.

Maybe `DialStatus` and the status can indicate if the status is a failure state or not? E.g:
```
DialStatusSendResponseFailure  DialStatus = "send_rsp_failure"
DialStatusBackendClose              DialStatus = "backend_closed"
...
```

No need to change them though, I think the current values achieve what we need. Mainly trying to avoid situations where logs or metrics are interpreted as cause for concern when they're not.

pkg/server/server.go
```diff
@@ -493,6 +493,7 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
 			// Currently not worrying about backend as we do not have an established connection,
 			s.PendingDial.Remove(random)
 			klog.V(5).InfoS("Removing pending dial request", "serverID", s.serverID, "dialID", random)
+			metrics.Metrics.ObserveDialFailure(metrics.DialFailureFrontendClose)
```
tallclair, line 496: Only count if there was a pending dial to remove.

cheftako: +1.

tallclair:
synced with logging changes. This should be good to go.

pkg/server/server.go
```diff
@@ -776,13 +778,17 @@ func (s *ProxyServer) serveRecvBackend(backend Backend, stream agent.AgentServic
...
+					metrics.Metrics.ObserveDialFailure(metrics.DialFailureErrorResponse)
 					dialErr = true
 				}
 				err := frontend.send(pkt)
 				s.PendingDial.Remove(resp.Random)
 				if err != nil {
 					klog.ErrorS(err, "DIAL_RSP send to frontend stream failure",
 						"dialID", resp.Random, "serverID", s.serverID, "agentID", agentID, "connectionID", resp.ConnectID)
+					if !dialErr { // Avoid double-counting.
+						metrics.Metrics.ObserveDialFailure(metrics.DialFailureSendResponse)
```
cheftako, line 790: I don't want to double count. However this is a somewhat interesting case as both the Frontend and Backend connections are both broken. During error handling I would generally expect one of them to be broken, Both seems more interesting.

tallclair: I agree that's an interesting case, but maybe better left to logs? I'm not sure dropping the conditional here would help surface this - or are you suggesting we add a different failure reason to capture this case?

PR #401, opened 2022-09-15
Adding basic test for server options.

cheftako:
Adding a test for our default values.
Adding a test for port validation.


cmd/server/app/options/options_test.go
```diff
@@ -0,0 +1,111 @@
...
+	expectedType := reflect.TypeOf(expected)
+	if actualType != expectedType {
+		t.Fatalf("For field %s, expected a value of type %s but got %s.", fieldName, actualType, expectedType)
+	}
+	if actual != expected {
+		t.Fatalf("For field %s, expected a value %v but got %v.", fieldName, actual, expected)
+	}
+}
+
+func TestValidateServerPortRange(t *testing.T) {
```
tallclair, line 56: (optional) consider combining these 3 port validation tests into a single table-driven test.

pkg/agent/client_test.go
```diff
@@ -354,6 +357,18 @@ func newClosePacket(connID int64) *client.Packet {
 	}
 }
 
+func waitForConnectionDeletion(t *testing.T, testClient *Client, connID int64) {
+	t.Helper()
+	var ok bool
+	wait.PollImmediate(100*time.Millisecond, 3*time.Second, func() (bool, error) {
```
tallclair, line 363: 3 seconds is a pretty short timeout when running on over-scheduled CI infrastructure. Consider using a higher timeout.

pkg/agent/client_test.go
```diff
@@ -354,6 +357,18 @@ func newClosePacket(connID int64) *client.Packet {
...
+		_, ok = testClient.connManager.Get(connID)
+		return !ok, nil
+	})
+	if ok {
```
tallclair, line 367: nit: wait.PollImmediat returns a timeout error if it times out, so I think this is usually done as
```suggestion
	if wait.PollImmediate(100*time.Millisecond, 3*time.Second, func() (bool, error) {
		_, ok = testClient.connManager.Get(connID)
		return !ok, nil
	}) != nil {
```

or if you prefer:

```suggestion
	err := wait.PollImmediate(100*time.Millisecond, 3*time.Second, func() (bool, error) {
		_, ok = testClient.connManager.Get(connID)
		return !ok, nil
	})
	if err != nil {
```

pkg/agent/client_test.go
```diff
@@ -266,6 +266,9 @@ func TestFailedSend_DialResp_GRPC(t *testing.T) {
 	defer ts.Close()
 
 	func() {
+		// goleak.IgnoreCurrent() needs to include all the goroutines which were started by testClient.Serve() above.
+		// That includes things like pkg/agent/client.go:614 +0x88. Sleep does not guarantee they will have started.
+		time.Sleep(time.Second)
```
tallclair, line 271: Blech. I don't like this, but I don't have a simple alternative. I wonder if we could add pprof labeling to agent go routines, and then filter based on the pprof labels? Sounds cleaner than trying to parse stack traces, but not something I expect you to implement in this PR.

tallclair (commented):
Just a bunch of nits, do with them as you please.

cmd/server/app/options/options_test.go
```diff
@@ -0,0 +1,111 @@
+package options
+
+import (
+	"testing"
+	"reflect"
+	"time"
+	"fmt"
+)
+
+func TestDefaultServerOptions(t *testing.T) {
```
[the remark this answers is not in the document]
cheftako: A little confused. These are the default flag values.

cmd/server/app/options/options_test.go
```diff
@@ -0,0 +1,111 @@
...
+	assertDefaultValue(t, "AgentServiceAccount", defaultServerOptions.AgentServiceAccount, "")
+	assertDefaultValue(t, "KubeconfigPath", defaultServerOptions.KubeconfigPath, "")
+	assertDefaultValue(t, "KubeconfigQPS", defaultServerOptions.KubeconfigQPS, float32(0))
+	assertDefaultValue(t, "KubeconfigBurst", defaultServerOptions.KubeconfigBurst, 0)
+	assertDefaultValue(t, "AuthenticationAudience", defaultServerOptions.AuthenticationAudience, "")
+	assertDefaultValue(t, "ProxyStrategies", defaultServerOptions.ProxyStrategies, "default")
+	assertDefaultValue(t, "CipherSuites", defaultServerOptions.CipherSuites, "")
+}
+
+func assertDefaultValue(t *testing.T, fieldName string, actual, expected interface{}) {
```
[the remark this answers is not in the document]
cheftako: Good suggestion. Done

cmd/server/app/options/options_test.go
cheftako: Switched to a generic table driven test.
Not convinced its an improvement but ....

cmd/server/app/options/options_test.go
```diff
@@ -0,0 +1,111 @@
...
+	assertValidationError(t, "HealthPort", testServerOptions.Validate(), fmt.Errorf("please do not try to use reserved port 1023 for the health port"))
+	testServerOptions.HealthPort = 1024
+	assertValidationError(t, "HealthPort", testServerOptions.Validate(), nil)
+	testServerOptions.HealthPort = 49151
+	assertValidationError(t, "HealthPort", testServerOptions.Validate(), nil)
+	testServerOptions.HealthPort = 49152
+	assertValidationError(t, "HealthPort", testServerOptions.Validate(), fmt.Errorf("please do not try to use ephemeral port 49152 for the health port"))
+}
+
+func assertValidationError(t *testing.T, fieldName string, actual, expected error) {
```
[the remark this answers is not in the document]
cheftako: Done

pkg/agent/client_test.go
cheftako: Not a huge fan either. Right now I would like to get this race condition to mostly not happen. Its not representative of a real failure AFAICT. If the issue persists lets investigate a more complicated fix.

pkg/agent/client_test.go
cheftako: Done

PR #402, opened 2022-09-16
Improve server logging, especially around initial dial

tallclair:
These changes ensure that we get the following information at `V(2)`:

1. Dial requests, along with destination address: logging the address makes it much easier to correlate dial & connection IDs with specific apiserver requests.
2. "Connection established": once the `DIAL_RSP` is received, make sure to log the transition with both connection & dial ID, to make it easier to connect those when tracing logs
3. On `DIAL_CLS`, log that the dial was canceled. These are logged at the error level, since it represents a "dial failure", even though it's not technically an error from the server perspective. This is inline with how we log agent errors in the dial response.

Also include several miscellaneous log cleanups.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -47,16 +47,16 @@ const xfrChannelSize = 10
 type key int
 
 type ProxyClientConnection struct {
-	Mode      string
-	Grpc      client.ProxyService_ProxyServer
-	HTTP      io.ReadWriter
-	CloseHTTP func() error
-	connected chan struct{}
-	connectID int64
-	agentID   string
-	start     time.Time
-	backend   Backend
-	address   string // cached for logging
+	Mode        string
+	Grpc        client.ProxyService_ProxyServer
+	HTTP        io.ReadWriter
+	CloseHTTP   func() error
+	connected   chan struct{}
+	connectID   int64
+	agentID     string
+	start       time.Time
+	backend     Backend
+	dialAddress string // cached for logging
 }
 
 const (
@@ -342,7 +342,7 @@ func NewProxyServer(serverID string, proxyStrategies []ProxyStrategy, serverCoun
 		case ProxyStrategyDefaultRoute:
 			bms = append(bms, NewDefaultRouteBackendManager())
 		default:
-			klog.V(4).InfoS("Unknonw proxy strategy", "strategy", ps)
+			klog.ErrorS(nil, "Unknown proxy strategy", "strategy", ps)
 		}
 	}
 
@@ -437,7 +437,7 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer, 
 		case client.PacketType_DIAL_REQ:
 			random := pkt.GetDialRequest().Random
 			address := pkt.GetDialRequest().Address
-			klog.V(3).InfoS("Received DIAL_REQ", "dialID", random, "address", address)
+			klog.V(3).InfoS("Received DIAL_REQ", "dialID", random, "dialAddress", address)
 			// TODO: if we track what agent has historically served
 			// the address, then we can send the Dial_REQ to the
 			// same agent. That way we save the agent from creating
@@ -464,12 +464,12 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer, 
 			s.PendingDial.Add(
 				random,
 				&ProxyClientConnection{
-					Mode:      "grpc",
-					Grpc:      stream,
-					connected: make(chan struct{}),
-					start:     time.Now(),
-					backend:   backend,
-					address:   address,
+					Mode:        "grpc",
+					Grpc:        stream,
+					connected:   make(chan struct{}),
+					start:       time.Now(),
+					backend:     backend,
+					dialAddress: address,
 				})
 			if err := backend.Send(pkt); err != nil {
 				klog.ErrorS(err, "DIAL_REQ to Backend failed", "serverID", s.serverID, "dialID", random)
@@ -497,10 +497,10 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer, 
 			klog.V(5).InfoS("Received DIAL_CLOSE", "serverID", s.serverID, "dialID", random)
 			// Currently not worrying about backend as we do not have an established connection,
 			if pd := s.PendingDial.Remove(random); pd != nil {
-				klog.ErrorS(nil, "Dial closed prematurely",
-					"source", "frontend",
+				klog.ErrorS(nil, "Dial cancelled (DIAL_CLS) by frontend",
 					"dialID", random,
-					"address", pd.address,
+					"dialAddress", pd.dialAddress,
+					"dialDuration", time.Since(pd.start),
 				)
 			}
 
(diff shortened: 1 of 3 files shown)


pkg/server/server.go
```diff
@@ -491,8 +496,13 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
 			random := pkt.GetCloseDial().Random
 			klog.V(5).InfoS("Received DIAL_CLOSE", "serverID", s.serverID, "dialID", random)
 			// Currently not worrying about backend as we do not have an established connection,
-			s.PendingDial.Remove(random)
-			klog.V(5).InfoS("Removing pending dial request", "serverID", s.serverID, "dialID", random)
+			if pd := s.PendingDial.Remove(random); pd != nil {
+				klog.ErrorS(nil, "Dial closed prematurely",
```
andrewsykim, line 500: Could it be misleading to consider this an error or should it be V(2) (or even V=0)? My current understanding is that sometimes `DIAL_CLS` could be received before the initial dial completes (e.g. short context deadline, etc)

pkg/server/server.go
```diff
@@ -797,8 +812,14 @@ func (s *ProxyServer) serveRecvBackend(backend Backend, stream agent.AgentServic
 				} else {
 					klog.V(5).InfoS("DIAL_CLS sent to frontend", "dialID", resp.Random)
 				}
-				s.PendingDial.Remove(resp.Random)
-				klog.V(5).InfoS("Close streaming", "agentID", agentID, "dialID", resp.Random)
+				if s.PendingDial.Remove(resp.Random) != nil {
+					klog.ErrorS(nil, "Dial closed prematurely",
```
andrewsykim, line 816: Same question here about whether logging error here could be misleading

pkg/server/server.go
```diff
@@ -432,13 +435,14 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
 	for pkt := range recvCh {
 		switch pkt.Type {
 		case client.PacketType_DIAL_REQ:
-			klog.V(5).Infoln("Received DIAL_REQ")
 			random := pkt.GetDialRequest().Random
+			address := pkt.GetDialRequest().Address
+			klog.V(3).InfoS("Received DIAL_REQ", "dialID", random, "address", address)
```
cheftako, line 440: I think this is the same value as "dialAddress" we have in labels (Eg. https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/agent/client.go#L440). It would be good to keep the naming consistent.

pkg/server/server.go
cheftako: +1. We know pending dials can be removed on one of several paths. This is not necessarily an error.

pkg/server/server.go
```diff
@@ -516,28 +526,25 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
 			klog.V(5).Infoln("DATA sent to Backend")
 
 		default:
-			klog.V(5).InfoS("Ignore packet coming from frontend",
+			klog.V(2).InfoS("Unrecognized packet from frontend",
```
cheftako, line 529: At some point we are going to want to extend our protocol. Since we cannot update all agents and servers simultaneously this will happen. Not sure we want this as V2 and littering our logs?

pkg/server/server.go
```diff
@@ -516,28 +526,25 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
...
 	}
 
 	klog.V(5).InfoS("Close streaming", "serverID", s.serverID, "connectionID", firstConnID)
 
-	pkt := &client.Packet{
-		Type: client.PacketType_CLOSE_REQ,
-		Payload: &client.Packet_CloseRequest{
-			CloseRequest: &client.CloseRequest{
-				ConnectID: firstConnID,
+	if backend != nil && firstConnID != 0 {
```
cheftako, line 536: if backend == nil something has gone fairly wrong. Would be good to understand that use case. Would prefer not to lose the log statement for that case.

pkg/server/server.go
tallclair: Restating from the PR description:

> On DIAL_CLS, log that the dial was canceled. These are logged at the error level, since it represents a "dial failure", even though it's not technically an error from the server perspective. This is inline with how we log agent errors in the dial response.

IMO logging this at error level is consistent with logging the error in the response at error level (https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/server/server.go#L764). This feels like a good fit for `Warning` level, but it seems like warning level is being deprecated? https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#change-log-functions

I could drop these to an info level, but the way I forsee using these logs is to debug dial failures, and logging them at error level makes the failures standout much more easily (at least in GCP logs).

WDYT?

pkg/server/server.go
tallclair: Yeah, that makes sense. This seems like a good usecase for the packet counts metric...

pkg/server/server.go
tallclair: Makes sense. The old log message wasn't really accurate though, so I changed to `Streaming closed before backend intialized`.

pkg/server/server.go
tallclair: Message changed to `Dial cancelled (DIAL_CLS) by frontend`

pkg/server/server.go
tallclair: Changed message to `Dial terminated (DIAL_CLS) by backend`

tallclair:
OK, I think I addressed everything we discussed on this. PTAL.

pkg/server/server.go
```diff
@@ -491,8 +496,13 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
 			random := pkt.GetCloseDial().Random
 			klog.V(5).InfoS("Received DIAL_CLOSE", "serverID", s.serverID, "dialID", random)
 			// Currently not worrying about backend as we do not have an established connection,
-			s.PendingDial.Remove(random)
-			klog.V(5).InfoS("Removing pending dial request", "serverID", s.serverID, "dialID", random)
+			if pd := s.PendingDial.Remove(random); pd != nil {
+				klog.ErrorS(nil, "Dial cancelled (DIAL_CLS) by frontend",
+					"dialID", random,
+					"dialAddress", pd.dialAddress,
+					"dialDuration", time.Since(pd.start),
```
andrewsykim, line 503: Noticed most log lines in this function include serverID as a key, wondering if it's worthwhile keeping that here too

pkg/server/server.go
```diff
@@ -797,8 +818,14 @@ func (s *ProxyServer) serveRecvBackend(backend Backend, stream agent.AgentServic
...
 					klog.V(5).InfoS("DIAL_CLS sent to frontend", "dialID", resp.Random)
 				}
-				s.PendingDial.Remove(resp.Random)
-				klog.V(5).InfoS("Close streaming", "agentID", agentID, "dialID", resp.Random)
+				if s.PendingDial.Remove(resp.Random) != nil {
+					klog.ErrorS(nil, "Dial terminated (DIAL_CLS) by backend",
+						"dialID", resp.Random,
+						"agentID", agentID,
+						"dialAddress", frontend.dialAddress,
+						"dialDuration", time.Since(frontend.start),
```
andrewsykim, line 826: Same comment  here about whether we should include serverID just for consistency

andrewsykim (commented):
/approve
/lgtm
/hold

Minor nit about logging serverID, feel free to ignore and cancel hold

tallclair:
/hold cancel

Discussed offline, will open a separate PR to clean up `serverID` logging.

tallclair:
Opened https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/409 for `serverID` logging.

PR #408, opened 2022-09-20
Adding tests for Agent options and validation.

cheftako:


cmd/agent/app/options/options_test.go
```diff
@@ -0,0 +1,151 @@
+package options
```
andrewsykim, line 1: Missing license header here.

I think `cmd/server/app/options/options_test.go` might be missing license header too.

cheftako: Sadly more than just these two. Did a quick scan and added license headers where I found them to be missing.

tallclair:
We might want to copy https://github.com/kubernetes/kubernetes/blob/master/hack/verify-boilerplate.sh into this repo.

cmd/agent/app/options/options_test.go
```diff
@@ -0,0 +1,167 @@
...
+import (
+	"testing"
+	"reflect"
+	"time"
+	"github.com/stretchr/testify/assert"
+	"fmt"
+)
+
+func TestDefaultServerOptions(t *testing.T) {
+	defaultAgentOptions := NewGrpcProxyAgentOptions()
```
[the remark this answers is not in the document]
cheftako: Done

PR #409, opened 2022-09-20
ServerID & AgentID logging cleanup

tallclair:
1. Allow serverID and agentID to be set by an environment variable. This makes it easier to set them using the [kubernetes downward API](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/) when running as a pod. For example, in a container spec:
```yaml
      env:
        - name: PROXY_SERVER_ID
          valueFrom:
            fieldRef:
              fieldPath: metadata.uid
```
2. Exclude the serverID from proxy-server logs (except startup flag logging), and likewise agentID for agent logs. These values are set at startup and never change, and are redundant with the log source.
3. (RFC) Also remove the redundant serverID / agentID from their respective go-routine labeling. I don't feel strongly about whether to include them here.


cmd/agent/app/options/options.go
```diff
@@ -212,3 +212,12 @@ func NewGrpcProxyAgentOptions() *GrpcProxyAgentOptions {
 	}
 	return &o
 }
+
+func defaultAgentID() string {
+	// Default to the value set by the PROXY_AGENT_ID environment variable. If both the flag &
+	// environment variable are set, the flag always wins.
+	if id := os.Getenv("PROXY_AGENT_ID"); id != "" {
```
cheftako, line 219: Not that your trying but its worth noting that we cannot detect multiple agents using the same agent id. We can see that we get repeat connection handshakes from the same agent but that's expected.

andrewsykim (commented):
/approve
/lgtm
/hold

PR #413, opened 2022-11-08
Fix agent shutdown dial race

tallclair:
If a dial is in progress while the agent client is shutting down, there is a race condition where the connection is registered after the connections were already cleaned up, which could lead to the connection being leaked.

This PR moves the closing responsibilities to the connection manager, and refuses to add new connections while the connection manager is shutting down. It also performs the connection cleanup in parallel.

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -75,17 +75,25 @@ func (c *connContext) send(msg []byte) {
 type connectionManager struct {
 	mu          sync.RWMutex
 	connections map[int64]*connContext
+	closed      bool
 }
 
-func (cm *connectionManager) Add(connID int64, ctx *connContext) {
+func (cm *connectionManager) Add(connID int64, ctx *connContext) error {
 	cm.mu.Lock()
 	defer cm.mu.Unlock()
+	if cm.closed {
+		return fmt.Errorf("connection manager is shutting down")
+	}
 	cm.connections[connID] = ctx
+	return nil
 }
 
 func (cm *connectionManager) Get(connID int64) (*connContext, bool) {
 	cm.mu.RLock()
 	defer cm.mu.RUnlock()
+	if cm.closed {
+		return nil, false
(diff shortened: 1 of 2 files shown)


pkg/agent/client.go
```diff
@@ -75,17 +75,27 @@ func (c *connContext) send(msg []byte) {
 type connectionManager struct {
 	mu          sync.RWMutex
 	connections map[int64]*connContext
+	closed      bool
 }
 
-func (cm *connectionManager) Add(connID int64, ctx *connContext) {
+// Add registers a new connection, and returns true if the connection was registered correctly.
+// If the connectionManager is shutting down, no new connections may be added.
+func (cm *connectionManager) Add(connID int64, ctx *connContext) bool {
```
jkh52, line 83: returning an err seems more self-documenting

pkg/agent/client.go
```diff
@@ -96,14 +106,21 @@ func (cm *connectionManager) Delete(connID int64) {
...
+	if cm.closed {
+		return // already closed
+	}
+	cm.closed = true
+
+	// Clean up connections in parallel.
 	for _, connCtx := range cm.connections {
-		connContexts = append(connContexts, connCtx)
+		connCtx := connCtx // Don't reuse loop variable.
+		runpprof.Do(ctx, runpprof.Labels(
```
jkh52, line 120: missing "go" here

pkg/agent/client.go
tallclair: Done.

pkg/agent/client.go
tallclair: Fixed. Thanks!

jkh52:
/approve

/hold
in case the test race is interesting

jkh52:
Friendly ping (needs rebase).

I think this is an attractive PR to prioritize (low risk and helps make other agent improvements).

Issue #410, opened 2022-10-05
Improve post-dial instrumentation


tallclair:
We've added a lot of instrumentation around dials, but we have very little instrumentation once the connection is established. Some ideas for metrics to add:

- [x] Total data packet counts in each direction (rate is the metric we care about)
- [ ] Per-connection packet counts (histogram)
- [ ] Connection lifespan (histogram)
- [x] Packet send failure count
- [ ] Connection close instrumentation, similar to the "dial failure reason" instrumenting we did (who closed the connection, why)

xoxonut:
Hi! I’d like to contribute to this issue. Which of the three remaining instrumentation would be good for a first-time contributor?

PR #415, opened 2022-11-14
Agent: add several new metrics.

jkh52:
Add several new metrics:

* dial_failure_count
* open_server_connections
* open_endpoint_connections
* packets
* stream_errors

Note that server_connection_failure_count metric (konnectivity-server dials) was easy to mistake for dial_failure_count (endpoint dials).

Includes some of the ideas in https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/410

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -80,6 +80,7 @@ type connectionManager struct {
 func (cm *connectionManager) Add(connID int64, ctx *connContext) {
 	cm.mu.Lock()
 	defer cm.mu.Unlock()
+	metrics.Metrics.EndpointConnectionInc()
 	cm.connections[connID] = ctx
 }
 
@@ -93,6 +94,9 @@ func (cm *connectionManager) Get(connID int64) (*connContext, bool) {
 func (cm *connectionManager) Delete(connID int64) {
 	cm.mu.Lock()
 	defer cm.mu.Unlock()
+	// Delete for a connID is called from cleanFunc, which is
+	// protected by cleanOnce.
+	metrics.Metrics.EndpointConnectionDec()
 	delete(cm.connections, connID)
 }
 
@@ -272,9 +276,11 @@ func (a *Client) Send(pkt *client.Packet) error {
 	a.sendLock.Lock()
 	defer a.sendLock.Unlock()
 
+	metrics.Metrics.ObservePacket(metrics.DirectionToServer, pkt.Type)
 	err := a.stream.Send(pkt)
 	if err != nil && err != io.EOF {
-		metrics.Metrics.ObserveFailure(metrics.DirectionToServer)
+		metrics.Metrics.ObserveServerFailure(metrics.DirectionToServer)
+		metrics.Metrics.ObserveStreamSendError(err, pkt.Type)
 		a.cs.RemoveClient(a.serverID)
 	}
 	return err
@@ -286,7 +292,8 @@ func (a *Client) Recv() (*client.Packet, error) {
 
 	pkt, err := a.stream.Recv()
 	if err != nil && err != io.EOF {
-		metrics.Metrics.ObserveFailure(metrics.DirectionFromServer)
+		metrics.Metrics.ObserveServerFailure(metrics.DirectionFromServer)
+		metrics.Metrics.ObserveStreamReceiveError(err)
 	}
 	return pkt, err
 }
@@ -381,6 +388,7 @@ func (a *Client) Serve() {
 			continue
 		}
 
+		metrics.Metrics.ObservePacket(metrics.DirectionFromServer, pkt.Type)
 		switch pkt.Type {
 		case client.PacketType_DIAL_REQ:
 			klog.V(4).InfoS("received DIAL_REQ", "serverID", a.serverID, "agentID", a.agentID)
@@ -444,6 +452,11 @@ func (a *Client) Serve() {
 				start := time.Now()
 				conn, err := net.DialTimeout(dialReq.Protocol, dialReq.Address, dialTimeout)
 				if err != nil {
+					reason := metrics.DialFailureUnknown
+					if neterr, ok := err.(net.Error); ok && neterr.Timeout() {
+						reason = metrics.DialFailureTimeout
+					}
+					metrics.Metrics.ObserveDialFailure(reason)
 					klog.ErrorS(err, "error dialing backend", "dialID", dialReq.Random)
 					dialResp.GetDialResponse().Error = err.Error()
 					if err := a.Send(dialResp); err != nil {
--- a/pkg/agent/metrics/metrics.go
+++ b/pkg/agent/metrics/metrics.go
@@ -20,13 +20,23 @@ import (
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"google.golang.org/grpc/status"
+	"sigs.k8s.io/apiserver-network-proxy/konnectivity-client/proto/client"
 )
 
 type Direction string
 
 const (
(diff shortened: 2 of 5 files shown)


pkg/agent/client.go
```diff
@@ -93,7 +94,10 @@ func (cm *connectionManager) Get(connID int64) (*connContext, bool) {
 func (cm *connectionManager) Delete(connID int64) {
 	cm.mu.Lock()
 	defer cm.mu.Unlock()
-	delete(cm.connections, connID)
+	if _, ok := cm.connections[connID]; ok {
+		metrics.Metrics.EndpointConnectionDec()
```
jkh52, line 98: Instead of this .Inc() + .Dec() approach, I would have much rather .Set() the gauge based on the len() of cm.connections. But I think that doesn't work because it would be per-client, and we want an overall count metric that does not have server-id cardinality. clientset.go doesn't take any read locks across clients and I didn't want to add that just for metrics.

Any suggestions are appreciated.

tallclair: Hmm. You could maintain a map of `serverID --> connection count`, and have `metrics.Metrics.SetEndpointConnectionCount(cm.serverID, len(cm.connections))`, which would update the map, take the sum across the map, and set the gauge based on that. Just need to make sure to clean up when clients are deleted.

I'm not sure if that's worth the complexity though...

tallclair (commented):
It would be nice to have tests for these. See https://github.com/kubernetes-sigs/apiserver-network-proxy/commit/075b152adaceac0995a8e3566c45fae463e57763#diff-2f0061389a0b73c21eb27c5a7de0b16fe1b8987638a0ac65a27b6be20fd100c4 for some examples of metrics testing.

pkg/agent/client.go
jkh52: That's nice in terms of avoiding new locks, and I think data accuracy. I am most afraid of the "duplicate client" deleted case.

/hold

pkg/agent/metrics/metrics.go
```diff
@@ -46,13 +46,43 @@ var (
...
+		[]string{},
+	)
+	dialFailures := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "dial_failure_total",
+			Help:      "Number of dial failures observed.",
+		},
+		[]string{},
```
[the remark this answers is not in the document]
jkh52: Agreed, added a reason label.

pkg/agent/client.go
```diff
@@ -442,8 +446,10 @@ func (a *Client) Serve() {
 			go runpprof.Do(context.Background(), labels, func(context.Context) {
 				defer close(dialDone)
 				start := time.Now()
+				metrics.Metrics.ObserveDialStart()
 				conn, err := net.DialTimeout(dialReq.Protocol, dialReq.Address, dialTimeout)
 				if err != nil {
+					metrics.Metrics.ObserveDialFailure()
```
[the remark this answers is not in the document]
jkh52: I want to have a separate stream errors metric. That way we don't have multiple dial failures for a given single-use tunnel.

pkg/agent/metrics/metrics.go
```diff
@@ -62,33 +92,59 @@ func newAgentMetrics() *AgentMetrics {
...
-	failures := prometheus.NewCounterVec(
-		prometheus.CounterOpts{
+	connections := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
 			Namespace: namespace,
 			Subsystem: subsystem,
-			Name:      "server_connection_failure_count",
-			Help:      "Count of failures to send to or receive from the proxy server, labeled by the direction (from_server or to_server)",
+			Name:      "endpoint_connections",
+			Help:      "Number of current endpoint connections.",
```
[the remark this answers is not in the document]
jkh52: Agreed, applied.

jkh52:
@tallclair Please see latest snapshot. I replaced dial_starts with a generic packets count, and I think the pattern will translate nicely to the other two binaries. Once we like the metric definitions, I'll add some unit tests.

pkg/agent/client.go
jkh52: I added a comment with explanation why I think as-is will be accurate.

pkg/agent/metrics/metrics.go
```diff
@@ -46,13 +47,34 @@ var (
...
 func newAgentMetrics() *AgentMetrics {
-	latencies := prometheus.NewHistogramVec(
+	serverFailures := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "server_connection_failure_count",
+			Help:      "Count of failures to send to or receive from the proxy server, labeled by the direction (from_server or to_server)",
+		},
+		[]string{"direction"},
```
tallclair, line 66: These should all be gRPC errors, right? It might be useful to add the error code as a label: https://pkg.go.dev/google.golang.org/grpc@v1.51.0/status#Code

pkg/agent/client.go
```diff
@@ -444,6 +451,11 @@ func (a *Client) Serve() {
 				start := time.Now()
 				conn, err := net.DialTimeout(dialReq.Protocol, dialReq.Address, dialTimeout)
 				if err != nil {
+					reason := metrics.DialFailureUnknown
+					if strings.Contains(err.Error(), "i/o timeout") {
```
tallclair, line 455: Better way to test this:
```suggestion
					if neterr, ok := err.(net.Error); ok && neterr.Timeout() {
```

pkg/agent/metrics/metrics.go
jkh52: This was not a real diff, just a move.

We could deprecate this metric with a stream errors (receive + sent) count, let me try adding here.

A recent Recv error log I found does appear to have code: `"could not read stream" err="rpc error: code = Unavailable desc = error reading from server: read tcp 10.76.0.7:55908->10.128.0.66:8132: read: connection timed out"`

tallclair: It might be sufficient to just look up the logs in that case. I don't think it's worth adding a new metric for it.

jkh52: Latest snapshot has stream errors, PTAL

pkg/agent/client.go
```diff
@@ -80,6 +80,7 @@ type connectionManager struct {
 func (cm *connectionManager) Add(connID int64, ctx *connContext) {
 	cm.mu.Lock()
 	defer cm.mu.Unlock()
+	metrics.Metrics.EndpointConnectionInc()
```
cheftako, line 83: Inc/dec everytime on Metrics *should* get you the correct answer, as long as no one changes the map outside of these methods. However we can also just use the  map length.

jkh52: There isn't a clientset map length, only one per client. There is some discussion here on this, maybe buried.

jkh52:
Prior snapshot got a `I1122 21:32:31.542236    7790 backend_manager.go:234] "This should not happen. Removed connection that is not the first connection" connection=&{ServerStream:0xc000028750} remainingConnections=[0xc002ac2708]
190`


https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_apiserver-network-proxy/415/pull-apiserver-network-proxy-test/1595168077585059840

jkh52:
I'll close this in favor of https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/423 to make sure all the names land consistency (there is rename advice there).

PR #417, opened 2022-11-17
Close backend connection when frontend is not found

tallclair:
I don't actually understand how we're getting into this state, but this log line shows up with some frequency:

```
"could not get frontend client" err="can't find connID 2 in the frontends[a19ff692-407c-4326-89e2-c53d8e4d2617]" serverID="1768996f-7935-40e8-9e24-51de50e81663" agentID="a19ff692-407c-4326-89e2-c53d8e4d2617" connectionID=2
```

This is a non-recoverable condition, indicating that somehow the frontend was disconnected but left the backend connection open. When this happens, reply to the backend with a close request.

### DIFF
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -339,19 +339,14 @@ func (t *grpcTunnel) serve(tunnelCtx context.Context) {
 
 			if !ok {
 				klog.ErrorS(nil, "Connection not recognized", "connectionID", resp.ConnectID, "packetType", "DATA")
-				req := &client.Packet{
+				t.Send(&client.Packet{
 					Type: client.PacketType_CLOSE_REQ,
 					Payload: &client.Packet_CloseRequest{
 						CloseRequest: &client.CloseRequest{
 							ConnectID: resp.ConnectID,
 						},
 					},
-				}
-				const segment = commonmetrics.SegmentFromClient
-				metrics.Metrics.ObservePacket(segment, req.Type)
-				if err := t.stream.Send(req); err != nil {
-					metrics.Metrics.ObserveStreamError(segment, err, req.Type)
-				}
+				})
 				continue
 			}
 			timer := time.NewTimer((time.Duration)(t.readTimeoutSeconds) * time.Second)
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -338,7 +338,15 @@ func (t *grpcTunnel) serve(tunnelCtx context.Context) {
 			conn, ok := t.conns.get(resp.ConnectID)
 
 			if !ok {
-				klog.V(1).InfoS("Connection not recognized", "connectionID", resp.ConnectID)
+				klog.ErrorS(nil, "Connection not recognized", "connectionID", resp.ConnectID)
+				t.stream.Send(&client.Packet{
+					Type: client.PacketType_CLOSE_REQ,
+					Payload: &client.Packet_CloseRequest{
+						CloseRequest: &client.CloseRequest{
+							ConnectID: resp.ConnectID,
+						},
+					},
+				})
 				continue
 			}
 			timer := time.NewTimer((time.Duration)(t.readTimeoutSeconds) * time.Second)
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -334,6 +334,10 @@ func (t *grpcTunnel) serve(tunnelCtx context.Context) {
 
 		case client.PacketType_DATA:
 			resp := pkt.GetData()
+			if resp.ConnectID == 0 {
+				klog.ErrorS(nil, "Received packet missing ConnectID", "packetType", "DATA")
+				continue
+			}
 			// TODO: flow control
 			conn, ok := t.conns.get(resp.ConnectID)
 
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -338,15 +338,20 @@ func (t *grpcTunnel) serve(tunnelCtx context.Context) {
 			conn, ok := t.conns.get(resp.ConnectID)
 
 			if !ok {
-				klog.ErrorS(nil, "Connection not recognized", "connectionID", resp.ConnectID)
-				t.stream.Send(&client.Packet{
+				klog.ErrorS(nil, "Connection not recognized", "connectionID", resp.ConnectID, "packetType", "DATA")
+				req := &client.Packet{
 					Type: client.PacketType_CLOSE_REQ,
 					Payload: &client.Packet_CloseRequest{
 						CloseRequest: &client.CloseRequest{
 							ConnectID: resp.ConnectID,
 						},
 					},
-				})
+				}
+				const segment = commonmetrics.SegmentFromClient
+				metrics.Metrics.ObservePacket(segment, req.Type)
+				if err := t.stream.Send(req); err != nil {
+					metrics.Metrics.ObserveStreamError(segment, err, req.Type)
+				}
 				continue
 			}
 			timer := time.NewTimer((time.Duration)(t.readTimeoutSeconds) * time.Second)
@@ -365,7 +370,7 @@ func (t *grpcTunnel) serve(tunnelCtx context.Context) {
 			conn, ok := t.conns.get(resp.ConnectID)
 
 			if !ok {
-				klog.V(1).InfoS("Connection not recognized", "connectionID", resp.ConnectID)
+				klog.V(1).InfoS("Connection not recognized", "connectionID", resp.ConnectID, "packetType", "CLOSE_RSP")
 				continue
 			}
 			close(conn.readCh)
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -497,6 +497,10 @@ func (a *Client) Serve() {
 		case client.PacketType_DATA:
 			data := pkt.GetData()
 			klog.V(4).InfoS("received DATA", "connectionID", data.ConnectID)
+			if data.ConnectID == 0 {
+				klog.ErrorS(nil, "Received packet missing ConnectID from frontend", "packetType", "DATA")
+				continue
+			}
 
 			ctx, ok := a.connManager.Get(data.ConnectID)
 			if ok {
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -602,9 +602,14 @@ func (a *Client) proxyToRemote(connID int64, ctx *connContext) {
 		// As the read side of the dataCh channel, we cannot close it.
 		// However serve() may be blocked writing to the channel,
 		// so we need to consume the channel until it is closed.
+		discardedPktCount := 0
 		for range ctx.dataCh {
 			// Ignore values as this indicates there was a problem
 			// with the remote connection.
+			discardedPktCount++
+		}
+		if discardedPktCount > 0 {
+			klog.V(2).InfoS("Discard packets while exiting proxyToRemote", "pktCount", discardedPktCount, "connectionID", connID)
 		}
 	}()
 
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -501,6 +501,18 @@ func (a *Client) Serve() {
 			ctx, ok := a.connManager.Get(data.ConnectID)
 			if ok {
 				ctx.send(data.Data)
+			} else {
+				klog.ErrorS(nil, "received DATA for unrecognized connection", "connectionID", data.ConnectID)
+				a.Send(&client.Packet{
+					Type: client.PacketType_CLOSE_RSP,
+					Payload: &client.Packet_CloseResponse{
+						CloseResponse: &client.CloseResponse{
+							ConnectID: data.ConnectID,
+							Error:     "unrecognized connectID",
+						},
+					},
+				})
+				continue
 			}
 
 		case client.PacketType_CLOSE_REQ:
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -502,7 +502,7 @@ func (a *Client) Serve() {
 			if ok {
 				ctx.send(data.Data)
 			} else {
-				klog.ErrorS(nil, "received DATA for unrecognized connection", "connectionID", data.ConnectID)
+				klog.V(2).InfoS("received DATA for unrecognized connection", "connectionID", data.ConnectID)
 				a.Send(&client.Packet{
 					Type: client.PacketType_CLOSE_RSP,
 					Payload: &client.Packet_CloseResponse{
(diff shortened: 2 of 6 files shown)


tallclair:
Actually, it's possible this was already fixed by https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/386. Even if that's the case, I think this is the correct way of handling this error, should it ever occur.

tallclair:
It occured to me that we should probably apply this same change everywhere a data packet is handled (client & agent too).

tests/concurrent_test.go
```diff
@@ -47,16 +47,16 @@ func TestProxy_ConcurrencyGRPC(t *testing.T) {
...
-	tunnel, err := client.CreateSingleUseGrpcTunnel(ctx, proxy.front, grpc.WithInsecure())
-	if err != nil {
-		t.Fatal(err)
-	}
-
 	var wg sync.WaitGroup
 	verify := func() {
 		defer wg.Done()
 
+		// run test client
```
tallclair, line 54: Note to reviewers: This test was reusing the tunnel across multiple requests, which we explicitly say not to do. It worked before because the test only sets up a single backend, and the FE request fits in a single data packet. This PR makes the failure more explicit when the connection is reused, which broke this test.

tallclair:
Rebased.

/assign @jkh52

jkh52:
`TestServerProxyConnectionMismatch` test failure looks deterministic.

tallclair:
Fixed. I'm not sure why it wasn't failing before. The order isn't supposed to matter on those mocked calls (they're not in an `InOrder(...)` block), but for some reason it does... I don't have time to dig deeper right now, but reordering fixes it.

pkg/server/server.go
```diff
@@ -433,6 +433,16 @@ func (s *ProxyServer) readFrontendToChannel(stream client.ProxyService_ProxyServ
 func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer, recvCh <-chan *client.Packet) {
 	klog.V(5).Infoln("start serving frontend stream")
 
+	defer func() {
+		// As the read side of the recvCh channel, we cannot close it.
+		// However readFrontendToChannel() may be blocked writing to the channel,
+		// so we need to consume the channel until it is closed.
+		for range recvCh {
```
jkh52, line 440: Is it possible to add unit test coverage?

What is the bad thing this prevents - leaked goroutine? If so I suggest explicit in this comment.

konnectivity-client/pkg/client/client.go
```diff
@@ -339,7 +339,15 @@ func (t *grpcTunnel) serve(tunnelCtx context.Context) {
 			conn, ok := t.conns.get(resp.ConnectID)
 
 			if !ok {
-				klog.V(1).InfoS("Connection not recognized", "connectionID", resp.ConnectID)
+				klog.ErrorS(nil, "Connection not recognized", "connectionID", resp.ConnectID)
```
cheftako, line 342: This could be an error but it can also be a race condition. We used to have this as an error and it tended to cause undue concern as it tends to happen when a connection is being shutdown. t.conns missing a reference to connectID usually just indicates that the connection is being shut down but that the other end sent a data packet before it realized the connection was going away.

pkg/agent/client.go
```diff
@@ -501,6 +501,18 @@ func (a *Client) Serve() {
 			ctx, ok := a.connManager.Get(data.ConnectID)
 			if ok {
 				ctx.send(data.Data)
+			} else {
+				klog.ErrorS(nil, "received DATA for unrecognized connection", "connectionID", data.ConnectID)
```
cheftako, line 505: Again, a little concerned about having this as an error.

pkg/server/server.go
```diff
@@ -433,6 +433,16 @@ func (s *ProxyServer) readFrontendToChannel(stream client.ProxyService_ProxyServ
...
+			// Ignore values as this indicates there was a problem
+			// with the remote connection.
+		}
```
cheftako, line 443: Can we add a debug message to let us know how many packets we dropped? Might even be worth adding a metric.

pkg/server/server.go
```diff
@@ -520,16 +530,24 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
...
+				klog.V(2).InfoS("Backend has not been initialized for the connection. Client should send a Dial Request first", "connectionID", connID)
+				s.sendFrontendClose(stream, connID, "backend not initialized")
+				return
+			}
+
 			if firstConnID == 0 {
 				firstConnID = connID
 			} else if firstConnID != connID {
-				klog.V(5).InfoS("Data does not match first connection id", "fistConnectionID", firstConnID, "connectionID", connID)
+				klog.ErrorS(nil, "Data does not match first connection id", "fistConnectionID", firstConnID, "connectionID", connID)
```
cheftako, line 542: fist?

konnectivity-client/pkg/client/client.go
```diff
@@ -339,7 +339,15 @@ func (t *grpcTunnel) serve(tunnelCtx context.Context) {
...
+				t.stream.Send(&client.Packet{
```
jkh52, line 343: ideally we would add ObservePacket() and conditional ObserveStreamError() here.

(See the other 2 stream send in this file, and note that t.stream is the raw stream.)

pkg/server/server.go
cheftako: Do we believe we have seen this? If so I would very much like to see a metric added.

pkg/server/server.go
```diff
@@ -520,16 +530,24 @@ func (s *ProxyServer) serveRecvFrontend(stream client.ProxyService_ProxyServer,
 			connID := pkt.GetData().ConnectID
 			data := pkt.GetData().Data
 			klog.V(5).InfoS("Received data from connection", "bytes", len(data), "connectionID", connID)
+			if backend == nil {
+				klog.V(2).InfoS("Backend has not been initialized for the connection. Client should send a Dial Request first", "connectionID", connID)
```
cheftako, line 534: Seems more likely the client sent the data packet on the wrong stream than that they attempted to send a data packet before a dial request.

konnectivity-client/pkg/client/client.go
tallclair: Hmm, good point. Especially with the fixes in this PR, it's pretty unlikely this would keep happening in another scenario. I'll change it back.

Not in this PR, but we may want to consider keeping a cache of recently closed connections. I've thought of a few cases where it would be useful.

pkg/agent/client.go
tallclair: Done.

pkg/server/server.go
tallclair: Without this, the go-routine that actually receives the packets and queues them up in the channel could be blocked here: https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/682e779546f9ebe97952d2ae892e7fd8027a5521/pkg/server/server.go#L422, which would prevent it from reading an EOF from the `Recv()` call: https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/682e779546f9ebe97952d2ae892e7fd8027a5521/pkg/server/server.go#L400-L401, which is what actually triggers the channel to be closed.

So yes, without this the go-routines could be deadlocked.

Note, this was copied from the agent client, which does the same thing: https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/agent/client.go#L586-L597

pkg/server/server.go
tallclair: Done. Also added to the agent version of this.

pkg/server/server.go
tallclair: I realized that `serverRecvBackend` also needs the same treatment, and added it there.

pkg/server/server.go
tallclair: Fixed. This could happen if a frontend client was attempting to reuse a tunnel. Agree on the metric, but I'd add it as part of a generic `connection_closed` metric with a `close_reason`. Basically the post-dial equivalent of the `dial_failures` metric: https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/410

pkg/server/server.go
```diff
@@ -856,7 +856,8 @@ func (s *ProxyServer) serveRecvBackend(backend Backend, stream agent.AgentServic
 			klog.V(5).InfoS("Received data from agent", "bytes", len(resp.Data), "agentID", agentID, "connectionID", resp.ConnectID)
 			frontend, err := s.getFrontend(agentID, resp.ConnectID)
 			if err != nil {
-				klog.ErrorS(err, "could not get frontend client", "agentID", agentID, "connectionID", resp.ConnectID)
+				klog.ErrorS(err, "could not get frontend client; closing conenction", "agentID", agentID, "connectionID", resp.ConnectID)
```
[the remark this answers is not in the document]
tallclair: Fixed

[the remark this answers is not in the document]
tallclair: The close case just means the frontend terminated the connection before the backend responded with the CLOSE_RSP (it could also mean the server initiated the close for some reason). In either case, the backend will have already shut down any connection state after sending a CLOSE_RSP, so no need to respond with a CLOSE_REQ in that case.

That said, it's a good call that in the frontend CLOSE_REQ error cases, we should return a CLOSE_RSP from the server so the client can fail fast. I'll add that as a separate commit.

konnectivity-client/pkg/client/client.go
tallclair: Done. We may want to consider switching to a wrapper around the ProxyStream client that records the metrics automatically.

pkg/server/server.go
tallclair: Agreed. This was copied from the log message under the `DATA` case, but it's weird in both places. I just removed the second sentence.

konnectivity-client/pkg/client/client.go
cheftako: +1 to keep recent connections to understand this case better.

konnectivity-client/pkg/client/client.go
```diff
@@ -339,7 +339,20 @@ func (t *grpcTunnel) serve(tunnelCtx context.Context) {
 			conn, ok := t.conns.get(resp.ConnectID)
 
 			if !ok {
-				klog.V(1).InfoS("Connection not recognized", "connectionID", resp.ConnectID)
+				klog.ErrorS(nil, "Connection not recognized", "connectionID", resp.ConnectID, "packetType", "DATA")
```
cheftako, line 342: I thought we were taking this back to Info?

pkg/server/server.go
cheftako: I think its fine but if we're worried we could adding a draining connection metric. Increment when we enter the defer and decrement when we exit. If this gets stuck calling the recvCh then the metric will stay up.

pkg/server/server.go
cheftako: Not sure even a reuse will cause a new connID. the tunnel sets the connID while the tunnel is being set up. 
The connection ID is set by the agent at https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/agent/client.go#L404.
The means it will only be created in response to a dial request which is created by the client at https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/cec42080fe55c2d21a2dfde35314f5f05bc65b85/konnectivity-client/pkg/client/client.go#L410.
Within the client that is only called from CreateSingleUseGrpcTunnelWithContext (https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/cec42080fe55c2d21a2dfde35314f5f05bc65b85/konnectivity-client/pkg/client/client.go#L169).
So reusing a connection on a tunnel should not create a new connection id.

konnectivity-client/pkg/client/client.go
```diff
@@ -339,7 +339,20 @@ func (t *grpcTunnel) serve(tunnelCtx context.Context) {
...
+				req := &client.Packet{
```
cheftako, line 343: No point in trying to send a close request if the ConnectID == 0.

pkg/agent/client.go
```diff
@@ -501,6 +501,18 @@ func (a *Client) Serve() {
 			ctx, ok := a.connManager.Get(data.ConnectID)
 			if ok {
 				ctx.send(data.Data)
+			} else {
+				klog.V(2).InfoS("received DATA for unrecognized connection", "connectionID", data.ConnectID)
+				a.Send(&client.Packet{
```
cheftako, line 506: Again no point is sending a close on connectID == 0.

cheftako:
I really like the drains, good catch. There are a couple of minor issues. Happy to lgtm/approve as soon as they are fixed.

pkg/server/server.go
jkh52: > I think its fine but if we're worried we could adding a draining connection metric. Increment when we enter the defer and decrement when we exit. If this gets stuck calling the recvCh then the metric will stay up.

IMO: keep the logs for now, and soon add a proxy-server metric for proxy connections labeled by state, similar to the client metric.

jkh52:
> I really like the drains, good catch. There are a couple of minor issues. Happy to lgtm/approve as soon as they are fixed.

+1 lots of good improvements here. I'm excited to tag after merging this, and observe agent memory improvement.

konnectivity-client/pkg/client/client.go
tallclair: The connection ID is only removed from `t.conns` when receiving a CLOSE_RSP from the server, not for a client-side close. So the only way this would be hit is if the server sent a DATA packet _after_ a CLOSE_RSP, which should never happen.

Note that I did change the CLOSE_RSP version of this log (line 374) to an info log, as there are cases where multiple CLOSE_RSP packets might be sent.

konnectivity-client/pkg/client/client.go
tallclair: The connection ID is only removed from `t.conns` when receiving a CLOSE_RSP from the server, not for a client-side close. So the only way this would be hit is if the server sent a DATA packet _after_ a CLOSE_RSP, which should never happen.

Note that I did change the CLOSE_RSP version of this log (line 374) to an info log, as there are cases where multiple CLOSE_RSP packets might be sent.

konnectivity-client/pkg/client/client.go
tallclair: I don't think we'd ever send a DATA packet with a `ConnectID == 0`? We could check for it here, but then we should check for it everywhere, but that seems unnecessary.

pkg/agent/client.go
tallclair: Same comment as above.

konnectivity-client/pkg/client/client.go
tallclair: Per discussion, added this as an extra safety measure.

konnectivity-client/pkg/client/client.go
cheftako: I by that argument (for now) but it suggests that we only clean up the connection map if we get a CLOSE_RSP packet. That seems like it could represent a possible leak. If we decide to fix that possible leak then we may need to revisit this log message.

PR #418, opened 2022-11-17
Clean up konnectivity-client go.mod

tallclair:
I noticed the konnectivity-client build is broken due to incorrect module dependencies. This slipped through since the way we run client tests uses the go.mod in the main module. I updated `make test` (which our CI uses) to run the client tests independently.


jkh52:
Thanks!

/lgtm
/approve

PR #422, opened 2022-11-24
Fix segfault on empty packet

ncopa:
Verify if pkt is nil before logging the type.

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -384,13 +384,12 @@ func (a *Client) Serve() {
 			return
 		}
 
-		klog.V(5).InfoS("[tracing] recv packet", "type", pkt.Type)
-
 		if pkt == nil {
 			klog.V(3).InfoS("empty packet received")
 			continue
 		}
 
+		klog.V(5).InfoS("[tracing] recv packet", "type", pkt.Type)
 		metrics.Metrics.ObservePacket(commonmetrics.SegmentToAgent, pkt.Type)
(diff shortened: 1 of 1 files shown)


jkh52:
This looks like a clear improvement, but the if block seems "unreachable" (because in agent.pb.go, if err is nil, packet should not be: https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/proto/agent/agent.pb.go#L107).

/lgtm
/ok-to-test

jkh52:
@ncopa, see previous comment - was this fix based on code inspection, or did you actually observe a segfault?

ncopa:
> @ncopa, see previous comment - was this fix based on code inspection, or did you actually observe a segfault?

Code inspection. Maybe I should change the commit message to be less dramatic? "Fix potential segfault on empty packet"

Another option is to remove the log line since it is unreachable (and thus dead code). In any case, the current order things happen does not make any sense.

cheftako:
> if block seems "unreachable"
If the block is unreachable then we should not need the 'if pkt nil' check. As long as we have it, I think logging pkt.type statement should come after the nil check.

ncopa:
> > if block seems "unreachable"
> 
> If the block is unreachable then we should not need the 'if pkt nil' check. As long as we have it, I think logging pkt.type statement should come after the nil check.

This is exactly the point here.

ncopa:
I re-based it against current master.

PR #423, opened 2022-11-28
More metrics

jkh52:
<WORK IN PROGRESS>

Add several new metrics, including new KAS client package.

Add similar stream metrics across all three binaries (client, server, agent).
* stream_packets_total
* stream_errors_total

For client, also:
* tunnel_connections (leak detection)

For agent, also:
* endpoint_dial_failure_total (help distinguish dial failures from KNP)
* open_server_connections (leak detection)
* open_endpoint_connections (leak detection)


pkg/agent/metrics/metrics.go
```diff
@@ -46,49 +56,158 @@ var (
...
 func newAgentMetrics() *AgentMetrics {
-	latencies := prometheus.NewHistogramVec(
+	dialLatencies := prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
-			Namespace: namespace,
-			Subsystem: subsystem,
-			Name:      "dial_duration_seconds",
+			Namespace: Namespace,
+			Subsystem: Subsystem,
+			Name:      DialLatencyMetric,
```
logicalhan, line 74: I would inline the metric names, it's confusing to look elsewhere to find the metric name.

pkg/agent/metrics/metrics.go
```diff
@@ -20,13 +20,23 @@ import (
...
 type Direction string
 
 const (
-	namespace = "konnectivity_network_proxy"
-	subsystem = "agent"
+	Namespace = "konnectivity_network_proxy"
+	Subsystem = "agent"
+
+	DialLatencyMetric         = "dial_duration_seconds"
+	ServerFailuresMetric      = "server_connection_failure_count"
```
logicalhan, line 34: counters should be suffixed with `_total`.

pkg/agent/metrics/metrics.go
```diff
@@ -20,13 +20,23 @@ import (
...
+	DialFailuresMetric        = "dial_failure_count"
```
logicalhan, line 35: same.

pkg/agent/metrics/metrics.go
jkh52: I want to stay consistent with https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/400, which seemed motivated by unit test reference. @tallclair WDYT

logicalhan: All of those metric names should be inlined imo.

pkg/agent/metrics/metrics.go
logicalhan: Not inlining the metric names makes it difficult to spot conventional naming errors like this one.

pkg/agent/metrics/metrics.go
jkh52: Ok - I inlined them for client and agent. I kept consistent within server/metrics consistent, since this PR doesn't touch it much.

logicalhan: Thanks!

pkg/agent/metrics/metrics.go
```diff
@@ -20,13 +20,15 @@ import (
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"google.golang.org/grpc/status"
+	"sigs.k8s.io/apiserver-network-proxy/konnectivity-client/proto/client"
```
logicalhan, line 24:
```suggestion
	"google.golang.org/grpc/status"

	"sigs.k8s.io/apiserver-network-proxy/konnectivity-client/proto/client"
```

pkg/agent/metrics/metrics.go
```diff
@@ -46,49 +48,158 @@ var (
...
-	return &AgentMetrics{failures: failures, latencies: latencies}
+	dialFailures := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: Namespace,
+			Subsystem: Subsystem,
+			Name:      "endpoint_dial_failure_total",
+			Help:      "Number of failures dialing the remote endpoint, by reason (example: timeout).",
+		},
+		[]string{"reason"},
+	)
```
logicalhan, line 89: Do we also need a counter for the total number of endpoint dials?

pkg/agent/metrics/metrics.go
jkh52: Fixed.

Does k/k have any automatic metric name validations, that should be added to this repo? I half-remember this from  https://github.com/kubernetes/kubernetes/pull/113520

pkg/agent/metrics/metrics.go
```diff
@@ -46,49 +48,158 @@ var (
...
 		[]string{},
 	)
-	failures := prometheus.NewCounterVec(
+	serverFailures := prometheus.NewCounterVec(
 		prometheus.CounterOpts{
-			Namespace: namespace,
-			Subsystem: subsystem,
+			Namespace: Namespace,
+			Subsystem: Subsystem,
 			Name:      "server_connection_failure_count",
```
logicalhan, line 76: This should be `_total`. Is this a new metric?

pkg/agent/metrics/metrics.go
logicalhan: We don't have it yet, but this is a good idea.

pkg/agent/metrics/metrics.go
jkh52: No, server_connection_failure_count is not new.

pkg/agent/metrics/metrics.go
logicalhan: Ah cool, as long as we have a total metric, this lgtm.

konnectivity-client/pkg/client/metrics/metrics.go
```diff
@@ -0,0 +1,114 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
logicalhan, line 2:
```suggestion
Copyright 2022 The Kubernetes Authors.
```

logicalhan (commented):
/lgtm from a metrics perspective.

Thanks for the iterations!

jkh52:
Test flake: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_apiserver-network-proxy/423/pull-apiserver-network-proxy-test/1597700633178148864

```
--- FAIL: TestServeData_HTTP (0.01s)
    client_test.go:134: client.connContext not released
```

konnectivity-client/go.mod
```diff
@@ -5,18 +5,30 @@ go 1.17
 // Prefer to keep requirements compatible with the oldest supported
 // k/k minor version, to prevent client backport issues.
 require (
```
jkh52, line 7: @tallclair PTAL, I'm not confident about this.

(Trying to stay as backward compat with k/k as possible, without the replace I end up getting protobuf 1.5.4)

jkh52: I fixed this in latest snapshot.

jkh52:
PTAL

In the latest snapshot, I switched from trying to extend the "direction" label (only used by agent, only 2 values, ambiguous from other binaries) and instead introduce "segment" label (4 values, meaningful across all 3 binaries, should make joining metrics much easier).

jkh52:
Converting back to Draft, because testing kube-apiserver compiled against this konnectivity-client does not emit the new metrics, I need to identify the gap in metrics registration.

jkh52:
/assign @liggitt 

Can you help point me to the right way of registering these metrics with KAS?

liggitt:
I think @logicalhan could probably direct you better

jkh52:
This PR directly calls prometheus.MustRegister, but the actually-working KAS metrics I see all seem to register with legacyregistry. It seems incorrect for konnectivity-client to do that. Is there a way for this library to self-register its metrics, or should it be done somewhere in KAS?

logicalhan:
> @logicalhan, this PR directly calls prometheus.MustRegister, but the actually-working KAS metrics I see all seem to register with legacyregistry. It seems incorrect for konnectivity-client to do that. Is there a way for this vendored library to self-register its metrics, or should it be done somewhere in KAS?

You can expose a method so that a registry can be passed in for registration. Something like [this](https://github.com/kubernetes/kubernetes/blob/master/pkg/proxy/metrics/metrics.go#L179-L195), but with Registry passed in as a parameter.

jkh52:
> > @logicalhan, this PR directly calls prometheus.MustRegister, but the actually-working KAS metrics I see all seem to register with legacyregistry. It seems incorrect for konnectivity-client to do that. Is there a way for this vendored library to self-register its metrics, or should it be done somewhere in KAS?
> 
> You can expose a method so that a registry can be passed in for registration. Something like [this](https://github.com/kubernetes/kubernetes/blob/master/pkg/proxy/metrics/metrics.go#L179-L195), but with Registry passed in as a parameter.

@logicalhan all the Registerers that I see in k/k seem to take component-base types. Whatever API this client library exports will need to be callable from k/k - will I be denied a PR there that uses legacyregistry.RawMustRegister?

liggitt:
> @logicalhan all the Registerers that I see in k/k seem to take component-base types. Whatever API this client library exports will need to be callable from k/k - will I be denied a PR there that uses legacyregistry.RawMustRegister?

yeah, type sharing is going to be tricky... no vendored dependency is allowed to transitively depend back on a staging repo, or we risk circular dependencies that cause gridlock

tallclair:
If you want to stick to the direction of defining the mitrics in konnectivity-client, the approach I would recommend is to expose a method that returns the metrics, and make sure the metrics returned implement the `componentbase.Registerable` interface: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-base/metrics/registry.go#L121

jkh52:
The latest snapshot is intended to work something like:

(in apiserver/pkg/server/egressselector/egress_selector.go)

func init() {
    client.Metrics.RegisterMetrics(legacyregistry.RawMustRegister)
}

@logicalhan PTAL, I don't see a clean way to do this that avoids using RawMustRegister.

jkh52:
> If you want to stick to the direction of defining the metrics in konnectivity-client, the approach I would recommend is to expose a method that returns the metrics, and make sure the metrics returned implement the `componentbase.Registerable` interface: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-base/metrics/registry.go#L121

That interface has a few methods that the bare prometheus metrics don't implement.

jkh52:
possible flake:

```
--- FAIL: TestBasicProxy_GRPC (0.58s)
    proxy_test.go:855: found unexpected goroutines:
        [Goroutine 1939 in state select, with net/http.(*persistConn).writeLoop on top of the stack:
        goroutine 1939 [select]:
        net/http.(*persistConn).writeLoop(0xc000984120)
        	/usr/local/go/src/net/http/transport.go:2392 +0x1a5
        created by net/http.(*Transport).dialConn
        	/usr/local/go/src/net/http/transport.go:1751 +0x2556
```

/retest

konnectivity-client/pkg/client/metrics/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+	streamEventsErrors := sharedmetrics.MakeStreamEventsErrorsMetric(Namespace, Subsystem)
+	return &ClientMetrics{
+		tConnections:       tConnections,
+		streamEvents:       streamEvents,
+		streamEventsErrors: streamEventsErrors,
+	}
+}
+
+// RegisterMetrics registers all metrics, given a prometheus MustRegister function.
+func (c *ClientMetrics) RegisterMetrics(mustRegisterFn func(...prometheus.Collector)) {
```
logicalhan, line 67: I would pass in a `prometheus.Registerer` instead.

jkh52: I couldn't find anything in legacyregistry that implements the Registerer interface. Maybe I'm missing something?

logicalhan: https://github.com/kubernetes/kubernetes/blob/3e26e104bdf9d0dc3c4046d6350b93557c67f3f4/staging/src/k8s.io/component-base/metrics/wrappers.go#L133-L140

jkh52: @logicalhan Thanks for the quick turnaround!

Just to be cautious of dependencies I expose both old and new way to register in this PR.

liggitt:
> The latest snapshot is intended to work something like:

Before merging here, I'd suggest getting a WIP PR in k/k that pulls in this change green (other than the presubmit check that complains about a non-canonical dependency repo) to demonstrate registering metrics successfully there.

You can pull in this fork in k/k with this:

```
hack/pin-dependency.sh sigs.k8s.io/apiserver-network-proxy=github.com/jkh52/apiserver-network-proxy more-metrics
hack/update-vendor.sh
```

jkh52:
> Before merging here, I'd suggest getting a WIP PR in k/k that pulls in this change green (other than the presubmit check that complains about a non-canonical dependency repo) to demonstrate registering metrics successfully there.

Good idea @liggitt: https://github.com/kubernetes/kubernetes/pull/114299

jkh52:
Latest snapshot parameterizes namespace and subsystem per discussion with @tallclair

pkg/server/server.go
```diff
@@ -396,7 +404,9 @@ func (s *ProxyServer) readFrontendToChannel(stream client.ProxyService_ProxyServ
 			close(stopCh)
 			return
 		}
+		segment := sharedmetrics.SegmentFromFrontend
```
tallclair, line 407: nit: make this const

pkg/server/server.go
```diff
@@ -713,12 +727,15 @@ func (s *ProxyServer) readBackendToChannel(stream agent.AgentService_ConnectServ
 			close(stopCh)
 			return
 		}
+		segment := sharedmetrics.SegmentFromBackend
```
tallclair, line 730: nit: const

pkg/server/server.go
```diff
@@ -880,7 +898,10 @@ func (s *ProxyServer) sendCloseRequest(stream agent.AgentService_ConnectServer,
 			},
 		},
 	}
+	segment := sharedmetrics.SegmentToBackend
```
tallclair, line 901: const

konnectivity-client/pkg/client/client.go
```diff
@@ -201,14 +212,19 @@ func (t *grpcTunnel) serve(tunnelCtx context.Context) {
 
 	for {
 		pkt, err := t.stream.Recv()
-		if err == io.EOF || t.isClosing() {
+		if err == io.EOF {
 			return
 		}
+		segment := sharedmetrics.SegmentToFrontend
 		if err != nil || pkt == nil {
 			klog.ErrorS(err, "stream read failure")
```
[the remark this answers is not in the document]
jkh52: I'd prefer to put the "trim unreachable code" cleanup in separate PRs. Maybe wait for https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/422 merge and then do a more thorough cleanup.

konnectivity-client/go.mod
```diff
@@ -6,13 +6,19 @@ go 1.17
...
 	github.com/golang/protobuf v1.4.3
+	github.com/prometheus/client_golang v1.0.0
+	github.com/prometheus/client_model v0.2.0 // indirect
+	github.com/prometheus/common v0.10.0 // indirect
```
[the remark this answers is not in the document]
jkh52: Yes, this doesn't seem right.

I just rebased to master since it has changes, and refreshed here.

konnectivity-client/go.mod
```diff
@@ -6,13 +6,19 @@ go 1.17
 // k/k minor version, to prevent client backport issues.
 require (
 	github.com/golang/protobuf v1.4.3
+	github.com/prometheus/client_golang v1.0.0
```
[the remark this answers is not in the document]
jkh52: This is konnectivity-client/go.mod, #432 is top level go.mod for server and agent.

Here I would prefer to be even more backward compat, for example release-1.21 uses v1.7.1.

konnectivity-client/pkg/client/client.go
```diff
@@ -29,6 +29,9 @@ import (
 
 	"google.golang.org/grpc"
 	"k8s.io/klog/v2"
+
+	"sigs.k8s.io/apiserver-network-proxy/konnectivity-client/pkg/client/metrics"
+	sharedmetrics "sigs.k8s.io/apiserver-network-proxy/konnectivity-client/pkg/shared/metrics"
```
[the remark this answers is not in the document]
jkh52: Done.

pkg/agent/client.go
```diff
@@ -272,9 +278,12 @@ func (a *Client) Send(pkt *client.Packet) error {
 	a.sendLock.Lock()
 	defer a.sendLock.Unlock()
 
+	segment := sharedmetrics.SegmentFromBackend
+	metrics.Metrics.ObservePacket(segment, pkt.Type)
 	err := a.stream.Send(pkt)
 	if err != nil && err != io.EOF {
-		metrics.Metrics.ObserveFailure(metrics.DirectionToServer)
+		metrics.Metrics.ObserveServerFailure(metrics.DirectionToServer)
```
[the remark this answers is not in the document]
jkh52: Done.

konnectivity-client/pkg/client/client.go
```diff
@@ -201,14 +212,19 @@ func (t *grpcTunnel) serve(tunnelCtx context.Context) {
 
 	for {
 		pkt, err := t.stream.Recv()
-		if err == io.EOF || t.isClosing() {
```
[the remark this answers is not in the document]
jkh52: I missed that - fixed to quiet log errors after isClosing (but keep the metric).

konnectivity-client/pkg/client/metrics/metrics.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+}
+
+func (c *ClientMetrics) ObservePacket(segment sharedmetrics.Segment, packetType client.PacketType) {
+	if c.isRegistered() {
+		sharedmetrics.ObservePacket(c.streamEvents, segment, packetType)
+	}
+}
+
+func (c *ClientMetrics) ObserveStreamErrorNoPacket(segment sharedmetrics.Segment, err error) {
```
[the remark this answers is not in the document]
jkh52: I briefly tried that but it requires passing *packetType and became more awkward.

konnectivity-client/pkg/shared/metrics/metrics.go
```diff
@@ -0,0 +1,63 @@
...
+	)
+}
+
+func MakeStreamEventsErrorsMetric(namespace, subsystem string) *prometheus.CounterVec {
+	return prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "stream_events_error_total",
+			Help:      "Count of stream_errors, by segment, grpc Code, packet type. (example: from_backend, Code.Unavailable, DIAL_RSP)",
```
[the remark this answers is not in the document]
jkh52: Done.

konnectivity-client/pkg/shared/metrics/metrics.go
```diff
@@ -0,0 +1,63 @@
+// Package metrics provides metric definitions and helpers used by
```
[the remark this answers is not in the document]
jkh52: Done.

konnectivity-client/pkg/client/metrics/metrics.go
```diff
@@ -0,0 +1,139 @@
...
+		mustRegisterFn(c.tConnections)
+		mustRegisterFn(c.streamEvents)
+		mustRegisterFn(c.streamEventsErrors)
+		atomic.StoreUint32(&c.registered, 1)
+	})
+}
+
+// Reset resets the metrics.
+func (c *ClientMetrics) Reset() {
+	if c.isRegistered() {
```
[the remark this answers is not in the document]
jkh52: Done.

konnectivity-client/pkg/client/metrics/metrics.go
```diff
@@ -0,0 +1,139 @@
...
+	return atomic.LoadUint32(&t.registered) != 0
+}
+
+func (c *ClientMetrics) populateMetrics(namespace, subsystem string) {
+	c.tConnections = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "tunnel_connections",
+			Help:      "Current number of open tunnel connections.",
```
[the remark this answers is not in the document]
jkh52: Yes, currently. But the intent is to capture count of in-use (konnectivity-client) net.Conn to measure client footprint, debug leaks, and compare against konnectivity-server metrics.

I'm punting this metric to a separate PR to refine. Among other things it should be labeled to indicate pending dial.

konnectivity-client/pkg/client/metrics/metrics.go
```diff
@@ -0,0 +1,139 @@
...
+		},
+		[]string{},
```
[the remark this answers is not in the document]
jkh52: This repo has this throughout; let's punt a cleanup to separate PR (I reverted this metric anyway).

konnectivity-client/pkg/shared/metrics/metrics.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"sigs.k8s.io/apiserver-network-proxy/konnectivity-client/proto/client"
+)
+
+// Segment identifies one of four tunnel segments (e.g. from server to backend).
+type Segment string
+
+const (
+	// SegmentFromFrontend indicates a packet from client to server.
+	SegmentFromFrontend Segment = "from_frontend"
```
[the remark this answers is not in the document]
jkh52: I may have made things worse by introducing "endpoint" where we already have the name "remote". :(

client seems more recent than frontend, and agent seems more recent than backend; so maybe those should be what we converge toward. Updated these labels.

konnectivity-client/pkg/shared/metrics/metrics.go
```diff
@@ -0,0 +1,63 @@
...
+	// SegmentFromFrontend indicates a packet from server to agent.
+	SegmentToBackend Segment = "to_backend"
+)
+
+func MakeStreamEventsTotalMetric(namespace, subsystem string) *prometheus.CounterVec {
+	return prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "stream_events_total",
```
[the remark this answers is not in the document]
jkh52: I agree enough that I made the change.

It may cause some confusion to a new reader, who may not realize our protocol has "Packet" message type.

konnectivity-client/pkg/shared/metrics/metrics.go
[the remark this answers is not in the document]
tallclair: Also, I think agent might use backend to refer to the endpoint... I like being explicit about client/server/agent/endpoint. Maybe we can do a clean-up later (separate PR) to change out frontend & backend everywhere.

konnectivity-client/pkg/client/metrics/metrics.go
```diff
@@ -0,0 +1,139 @@
...
+		streamEvents:       nil,
+		streamEventsErrors: nil,
+	}
+}
+
+func (t *ClientMetrics) isRegistered() bool {
+	return atomic.LoadUint32(&t.registered) != 0
+}
+
+func (c *ClientMetrics) populateMetrics(namespace, subsystem string) {
```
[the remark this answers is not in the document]
logicalhan: We can use whatever string makes sense, but I would not actually recommend parameterizing namespace and subsystem because it breaks our ability to run static analysis.

konnectivity-client/pkg/shared/metrics/metrics.go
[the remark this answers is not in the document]
tallclair: Ah, that makes sense. I think "message" could be less ambiguous, but I think we should stick with "packet" since that's the actual type name.

konnectivity-client/pkg/common/metrics/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+	SegmentFromClient Segment = "from_client"
+	// SegmentToClient indicates a packet from server to client.
+	SegmentToClient Segment = "to_client"
+	// SegmentFromClient indicates a packet from agent to server.
+	SegmentFromAgent Segment = "from_agent"
+	// SegmentFromClient indicates a packet from server to agent.
+	SegmentToAgent Segment = "to_agent"
+)
+
+func MakeStreamEventsTotalMetric(namespace, subsystem string) *prometheus.CounterVec {
```
tallclair, line 42: nit: change these function names to match the metric name change
```suggestion
func MakeStreamPacketsTotalMetric(namespace, subsystem string) *prometheus.CounterVec {
```

konnectivity-client/pkg/common/metrics/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "stream_packets_total",
+			Help:      "Count of packets processed, by segment and packet type (example: from_client, DIAL_REQ)",
+		},
+		[]string{"segment", "packet_type"},
+	)
+}
+
+func MakeStreamEventsErrorsMetric(namespace, subsystem string) *prometheus.CounterVec {
```
tallclair, line 54:
```suggestion
func MakeStreamErrorsMetric(namespace, subsystem string) *prometheus.CounterVec {
```

pkg/agent/metrics/metrics.go
```diff
@@ -46,49 +49,140 @@ var (
...
 // AgentMetrics includes all the metrics of the proxy agent.
 type AgentMetrics struct {
-	latencies *prometheus.HistogramVec
-	failures  *prometheus.CounterVec
+	dialLatencies       *prometheus.HistogramVec
+	serverFailures      *prometheus.CounterVec
+	dialFailures        *prometheus.CounterVec
+	serverConnections   *prometheus.GaugeVec
+	endpointConnections *prometheus.GaugeVec
+	streamEvents        *prometheus.CounterVec
```
tallclair, line 57: s/streamEvents/streamPackets/g
everywhere

pkg/server/metrics/metrics.go
```diff
@@ -142,77 +147,84 @@ func newServerMetrics() *ServerMetrics {
...
-		dialFailures:      dialFailures,
+		endpointLatencies:  endpointLatencies,
+		frontendLatencies:  frontendLatencies,
+		grpcConnections:    grpcConnections,
+		httpConnections:    httpConnections,
+		backend:            backend,
+		pendingDials:       pendingDials,
+		fullRecvChannels:   fullRecvChannels,
+		dialFailures:       dialFailures,
+		streamEvents:       streamEvents,
```
tallclair, line 171: ditto: s/streamEvents/streamPackets/g

tallclair (commented):
Looks good, but I'd like to get an answer to the question of client metric namespace before merging.

konnectivity-client/pkg/client/metrics/metrics.go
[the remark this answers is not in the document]
jkh52: In this case, I'll use the same namespace as server and agent ('konnectivity_network_proxy') and add subsystem 'client'.

konnectivity-client/pkg/common/metrics/metrics.go
jkh52: Done.

konnectivity-client/pkg/common/metrics/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+	// SegmentFromAgent indicates a packet from agent to server.
+	SegmentFromAgent Segment = "from_agent"
+	// SegmentToAgent indicates a packet from server to agent.
+	SegmentToAgent Segment = "to_agent"
+)
+
+func MakeStreamPacketsTotalMetric(namespace, subsystem string) *prometheus.CounterVec {
+	return prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespace,
```
tallclair, line 45: nit: the namespace should be consistent between all 3 components now, so you could just move it to const in this package and remove the namespace paramater. Fine to follow up with this.

tallclair (commented):
/lgtm

Thanks!

PR #424, opened 2022-11-30
Trim unused proto.

jkh52:
Trim unused proto.


tallclair:
/hold
for a later release

/assign @cheftako

cheftako:
You don't seem to have regenerated konnectivity-client/proto/client/client.pb.go. I think you at minimum need to do that and see what the consequences of those changes are.

jkh52:
> You don't seem to have regenerated konnectivity-client/proto/client/client.pb.go.

whoops - fixed thanks.

jkh52:
@cheftako I rebased this PR after #434, but there appears to be fallout:

```
protoc -I . proto/agent/agent.proto --go_out=. --go_opt=paths=source_relative --go-grpc_out=require_unimplemented_servers=false:. --go-grpc_opt=paths=source_relative
make[1]: *** [Makefile:116: proto/agent/agent.pb.go] Error 127
```

Makefile
```diff
@@ -50,6 +50,7 @@ PROXY_SERVER_IP ?= 127.0.0.1
 ## --------------------------------------
 mock_gen:
 	mkdir -p proto/agent/mocks
+	rm -f proto/agent/mocks/agent_mock.licensed.go proto/agent/agent_grpc.licensed.go
```
cheftako, line 53: I don't think we should need this line as a successful make should always move the file, hence cleaning it up. So I happy to have this in clean but I don't think it should be part of the normal make. The fact that we don't have client/proto/client/client_grpc.licensed.go here supports that.

If we do decide to keep it I would rather have this in the phony top level target can we move the deletes to the proper file targets? (Eg. `proto/agent/agent_grpc.pb.go proto/agent/agent.pb.go:`)

jkh52: You're right that it seems to not belong here.

I removed it and re-tested `make gen` a few times, at first it began mysteriously succeeding (where it failed before, with "unexpected EOF"). Then on the most recent try, it seemed to succeed but the resulting proto/agent/agent_grpc.pb.go was missing the license header. So, I think as-is there is some non-determinism.

PR #426, opened 2022-12-01
Fix test flake.

jkh52:
--- FAIL: TestServeData_HTTP (0.01s)
    client_test.go:134: client.connContext not released

Also make test sleeps a bit more lean.

### DIFF
--- a/pkg/agent/client_test.go
+++ b/pkg/agent/client_test.go
@@ -129,6 +129,9 @@ func TestServeData_HTTP(t *testing.T) {
 		t.Errorf("expect nil closeErr; got %v", closeErr)
 	}
 
+	// give agent time to clean up.
+	time.Sleep(250 * time.Millisecond)
+
 	// Verify internal state is consistent
 	if _, ok := testClient.connManager.Get(connID); ok {
 		t.Error("client.connContext not released")
--- a/pkg/server/server_test.go
+++ b/pkg/server/server_test.go
@@ -265,8 +265,8 @@ func baseServerProxyTestWithoutBackend(t *testing.T, validate func(*agentmock.Mo
 
 	proxyServer.Proxy(frontendConn)
 
-	// add a sleep to make sure `serveRecvFrontend` ends after `Proxy` finished.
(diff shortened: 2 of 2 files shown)


pkg/agent/client_test.go
```diff
@@ -129,6 +129,9 @@ func TestServeData_HTTP(t *testing.T) {
 		t.Errorf("expect nil closeErr; got %v", closeErr)
 	}
 
+	// give agent time to clean up.
+	time.Sleep(250 * time.Millisecond)
```
tallclair, line 133: I strongly dislike sleeps in tests. There is always a tradeoff between test performance & flakiness. In a heavily contended CI environment, it's possible that a process could be paused for multiple seconds, which basically ignores the sleep (since it's wall time).

When it's not possible to get a definitive "done" signal, it's better to poll for the desired state. Fortunately we already have a helper to do just that: `waitForConnectionDeletion(t, testClient, connID)`

pkg/server/server_test.go
```diff
@@ -265,8 +265,8 @@ func baseServerProxyTestWithoutBackend(t *testing.T, validate func(*agentmock.Mo
 
 	proxyServer.Proxy(frontendConn)
 
-	// add a sleep to make sure `serveRecvFrontend` ends after `Proxy` finished.
-	time.Sleep(1 * time.Second)
+	// make sure `serveRecvFrontend` ends after `Proxy` finished.
```
tallclair, line 268: My suggestion here got too long, so I decided it was easier to just send a PR. What do you think of this approach? https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/427

jkh52: Yes that approach is definitely much better.

/hold

jkh52:
Very similar flake: https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/447#issuecomment-1366226385

jkh52:
Obsoleted by https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/427

PR #427, opened 2022-12-02
[server] Block Proxy from returning on both go-routines exiting

tallclair:
This PR prevents the server's `Proxy` method from returning until both `readFrontendToChannel` and `serveRecvFrontend` have completed. This lets us remove the `time.Sleep` from `server_test.go` and de-flake the test.

This also has the added benefit of reducing the number of go-routines started by Proxy by 1.

/assign @jkh52

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -399,7 +399,7 @@ func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
 	klog.V(5).InfoS("Proxy request from client", "userAgent", userAgent, "serverID", s.serverID)
 
 	recvCh := make(chan *client.Packet, xfrChannelSize)
-	stopCh := make(chan error)
+	stopCh := make(chan error, 1)
 
 	frontend := GrpcFrontend{
 		stream: stream,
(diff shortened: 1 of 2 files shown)


jkh52:
LGTM but I there is a test failure or flake.

/hold
Let's include in the next release, unless we become very convinced this has no unintended side effects.

tallclair:
Whoops. Glad we have tests to catch that stop channel deadlock... I'm surprised none of the server unit tests caught it though

pkg/server/server.go
```diff
@@ -369,25 +369,23 @@ func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
...
-	go runpprof.Do(context.Background(), labels, func(context.Context) { s.serveRecvFrontend(stream, recvCh) })
-
-	defer func() {
-		close(recvCh)
-	}()
-
 	// Start goroutine to receive packets from frontend and push to recvCh
 	go runpprof.Do(context.Background(), labels, func(context.Context) { s.readFrontendToChannel(stream, userAgent, recvCh, stopCh) })
 
+	s.serveRecvFrontend(stream, recvCh)
```
cheftako, line 381: serveRecvFrontend is a blocking function. Before we would return from Proxy() immediately if readFrontendToChannel stream was closed or hit an error. Now we need the full receive channel to be emptied, the channel closed and then the clean up be completed. I'm not saying this is wrong but it is a significant change. Would like to be sure this is actually a safe change.

jkh52: I agree it's significant, but think we should move forward, at least in master branch. In a WIP change, I'm doing something similar anyway (block until receive channel is done processing, then clean any lingering connections).

The main risk is `serveRecvFrontend` failing to return; if that happens we will have good visibility via the metric that will never decrement (`metrics.Metrics.ConnectionDec`). I have not seen this metric have any leak behavior, for recent release tags, so if that changes we will know this likely caused it.

jkh52:
@tallclair please rebase

tallclair:
Done. I also moved the closing of `stopCh` to a defer, to guarantee it's always closed.

jkh52:
> Done. I also moved the closing of `stopCh` to a defer, to guarantee it's always closed.

+1 - I almost did that in a different PR.

PR #428, opened 2022-12-06
update dependencies to patch CVEs

ipochi:
Current master has following CVE's, this PR updates the dependencies to mitigate all of them.

CVE-2022-27664
CVE-2022-32149
GHSA-69ch-w2m2-3vjp

Signed-off-by: Imran Pochi <imranpochi@microsoft.com>


cheftako:
Do we also want to update https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/konnectivity-client/go.mod for the client library?

ipochi:
@cheftako CI is green 🟢 😄

konnectivity-client/go.mod
```diff
@@ -5,18 +5,19 @@ go 1.17
 // Prefer to keep requirements compatible with the oldest supported
 // k/k minor version, to prevent client backport issues.
 require (
-	github.com/golang/protobuf v1.4.3
+	github.com/golang/protobuf v1.5.2
```
jkh52, line 8: This would likely prevent backporting konnectivity-client to earlier k/k release branch.

Is there any CVE impact if we simply leave konnectivity-client as is?

jkh52:
Do the Docker parent images need an update? Similar to https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/371

jkh52:
> @jkh52
> 
> Shall I close this in light of #432 ?

#432 seems to fix the x/text/language vulnerabilities, but we still have go 1.18.5 so the first vulnerability seems to need fixing, you could reduce this PR to that. (See my comment - similar to #371)

ipochi:
@jkh52 

I'm out on vacation for the next couple of weeks. I won't be able to update the PR.

Feel free to update or close this.

PR #435, opened 2022-12-09
Add more pprof functions to server and agent’s admin server

zqzten:
This PR added some missing pprof functions to the server and agent’s admin server.
It also updated .gitignore to include some common paraphernalia.

### DIFF
--- a/cmd/agent/app/server.go
+++ b/cmd/agent/app/server.go
@@ -152,6 +152,9 @@ func (a *Agent) runAdminServer(o *options.GrpcProxyAgentOptions) error {
 	if o.EnableProfiling {
 		muxHandler.HandleFunc("/debug/pprof", util.RedirectTo("/debug/pprof/"))
 		muxHandler.HandleFunc("/debug/pprof/", pprof.Index)
+		muxHandler.HandleFunc("/debug/pprof/profile", pprof.Profile)
+		muxHandler.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+		muxHandler.HandleFunc("/debug/pprof/trace", pprof.Trace)
 		if o.EnableContentionProfiling {
 			runtime.SetBlockProfileRate(1)
 		}
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,10 @@
 /easy-rsa-master/
 /easy-rsa.tar.gz
 /easy-rsa
+
+# editor and IDE paraphernalia
+.idea
+.vscode
+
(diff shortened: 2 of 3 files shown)


cmd/agent/app/server.go
```diff
@@ -152,6 +152,10 @@ func (a *Agent) runAdminServer(o *options.GrpcProxyAgentOptions) error {
 	if o.EnableProfiling {
 		muxHandler.HandleFunc("/debug/pprof", util.RedirectTo("/debug/pprof/"))
 		muxHandler.HandleFunc("/debug/pprof/", pprof.Index)
+		muxHandler.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+		muxHandler.HandleFunc("/debug/pprof/profile", pprof.Profile)
```
jkh52, line 156: Is this alone meaningful? (Given that the agent does not integrate with [cpu profiling](https://pkg.go.dev/runtime/pprof#StartCPUProfile))

zqzten: Yeah, `pprof.Profile` does proactive cpu profiling so we don't have to explicitly integrate that elsewhere in the code.

cmd/agent/app/server.go
```diff
@@ -152,6 +152,10 @@ func (a *Agent) runAdminServer(o *options.GrpcProxyAgentOptions) error {
 	if o.EnableProfiling {
 		muxHandler.HandleFunc("/debug/pprof", util.RedirectTo("/debug/pprof/"))
 		muxHandler.HandleFunc("/debug/pprof/", pprof.Index)
+		muxHandler.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
```
cheftako, line 155: You can make an argument that all the pprof methods can expose sensitive information and that is why we require it to be explicitly enabled. That having been said cmdline both seems to potentially expose things the most (eg. AuthenticationAudience and anything we unwittingly add later) but also the least useful given that the cmdline flags are already being logged at startup. Additionally its much easier to do things like mask fields we consider to be too sensitive in the logs. Much harder to do that on pprof/cmdline. I'm ok with the rest but I would prefer we not expose cmdline.

zqzten: That sounds reasonable, removed.

cmd/agent/app/server.go
jkh52: ack, thanks. I'll defer to @cheftako to approve.

PR #441, opened 2022-12-20
konnectivity-client: add a client connections gauge metric.

jkh52:
This follows https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/423 and https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/438

The motivation is to make it easier to monitor runaway resource consumption, and narrow down root cause.

### DIFF
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -134,6 +134,9 @@ type grpcTunnel struct {
 	// closing should only be accessed through atomic methods.
 	// TODO: switch this to an atomic.Bool once the client is exclusively buit with go1.19+
 	closing uint32
+
+	// Stores the current metrics.ClientConnectionStatus
+	prevStatus atomic.Value
 }
 
 type clientConn interface {
@@ -185,7 +188,7 @@ func CreateSingleUseGrpcTunnelWithContext(createCtx, tunnelCtx context.Context, 
 }
 
 func newUnstartedTunnel(stream client.ProxyService_ProxyClient, c clientConn) *grpcTunnel {
-	return &grpcTunnel{
+	t := grpcTunnel{
 		stream:             stream,
 		clientConn:         c,
 		pendingDial:        pendingDialManager{pendingDials: make(map[int64]pendingDial)},
@@ -193,6 +196,36 @@ func newUnstartedTunnel(stream client.ProxyService_ProxyClient, c clientConn) *g
 		readTimeoutSeconds: 10,
 		done:               make(chan struct{}),
 	}
+	s := metrics.ClientConnectionStatusCreated
+	t.prevStatus.Store(s)
+	metrics.Metrics.GetClientConnectionsMetric().WithLabelValues(string(s)).Inc()
+	return &t
+}
+
+func (t *grpcTunnel) updateMetric(status metrics.ClientConnectionStatus) {
+	select {
+	case <-t.Done():
+		return
+	default:
+	}
+
+	prevStatus := t.prevStatus.Swap(status).(metrics.ClientConnectionStatus)
+
+	m := metrics.Metrics.GetClientConnectionsMetric()
+	m.WithLabelValues(string(prevStatus)).Dec()
+	m.WithLabelValues(string(status)).Inc()
+}
+
+// closeMetric should be called exactly once to finalize client_connections metric.
+func (t *grpcTunnel) closeMetric() {
+	select {
+	case <-t.Done():
+		return
+	default:
+	}
+	prevStatus := t.prevStatus.Load().(metrics.ClientConnectionStatus)
+
+	metrics.Metrics.GetClientConnectionsMetric().WithLabelValues(string(prevStatus)).Dec()
 }
 
 func (t *grpcTunnel) serve(tunnelCtx context.Context) {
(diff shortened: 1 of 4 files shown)


jkh52:
FYI `make lint` appears to be broken at HEAD.

jkh52:
/retest

(To see if it's a flake or not.) Lint refresh at #442

jkh52:
/retest

Now that https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/442 is merged

konnectivity-client/pkg/client/client.go
```diff
@@ -185,14 +189,41 @@ func CreateSingleUseGrpcTunnelWithContext(createCtx, tunnelCtx context.Context,
...
+	defer t.statusLock.Unlock()
+	metrics.Metrics.GetTunnelsMetric().WithLabelValues(string(status)).Inc()
+	t.prevStatus = status
+}
+
+func (t *grpcTunnel) updateMetric(status metrics.TunnelStatus) {
+	t.statusLock.Lock()
+	defer t.statusLock.Unlock()
+	m := metrics.Metrics.GetTunnelsMetric()
+	m.WithLabelValues(string(t.prevStatus)).Dec()
```
cheftako, line 217: Dislike having this much under lock, especially when we are acquiring yet more locks under the statusLock. 

Maybe something more like 
func (t *grpcTunnel) setMetric(status metrics.TunnelStatus) metrics.TunnelStatus {
	t.statusLock.Lock()
	defer t.statusLock.Unlock()
	prev := t.prevStatus
	t.prevStatus = status
	return prev
}

func (t *grpcTunnel) updateMetric(status metrics.TunnelStatus) {
	prev := t.setMetric(status)
	m := metrics.Metrics.GetTunnelsMetric()
	m.WithLabelValues(string(prev)).Dec()
	m.WithLabelValues(string(status)).Inc()
}

konnectivity-client/pkg/client/client.go
```diff
@@ -134,6 +134,10 @@ type grpcTunnel struct {
 	// closing should only be accessed through atomic methods.
 	// TODO: switch this to an atomic.Bool once the client is exclusively buit with go1.19+
 	closing uint32
+
+	// Support a gauge tunnels metric by status. Access must be synchronized.
```
cheftako, line 138: Would we be better of replacing these two with https://pkg.go.dev/sync/atomic#Value ?
I think this obviates the need for writing a setMetric helper.

konnectivity-client/pkg/client/client.go
```diff
@@ -185,14 +189,41 @@ func CreateSingleUseGrpcTunnelWithContext(createCtx, tunnelCtx context.Context,
...
+func (t *grpcTunnel) updateMetric(status metrics.TunnelStatus) {
+	t.statusLock.Lock()
+	defer t.statusLock.Unlock()
+	m := metrics.Metrics.GetTunnelsMetric()
+	m.WithLabelValues(string(t.prevStatus)).Dec()
+	m.WithLabelValues(string(status)).Inc()
+	t.prevStatus = status
+}
+
+// closeMetric must be called exactly once to finalize tunnels metric.
```
cheftako, line 222: Yousay it should only be called once but we do nothing to ensure this is true. It also seems like the more important thing is that update should not be called after this has been called. However we do nothing to ensure that is true?

konnectivity-client/pkg/common/metrics/testing/metrics.go
```diff
@@ -30,6 +30,11 @@ const (
 # HELP konnectivity_network_proxy_client_dial_failure_total Number of dial failures observed, by reason (example: remote endpoint error)
 # TYPE konnectivity_network_proxy_client_dial_failure_total counter`
 	clientDialFailureSample = `konnectivity_network_proxy_client_dial_failure_total{reason="%s"} %d`
+
+	tunnelsHeader = `
+# HELP konnectivity_network_proxy_client_tunnels Number of in-use tunnels, by status (Example: dialing)
```
cheftako, line 35: I don't believe tunnels is a well defined term. To me the tunnel is the OS visible network connection over which individual connections are transported (tunneled). However it seems other folks have other ideas. I think it would be clearer to phrase things in terms on GRPC/HTTP-Connect connections and client/request connections.

cheftako: I believe this will become a lot more important if we write the code to allow the konnectivity client to mutex the client connections over a single GRPC connection, the way the server and agent currently do.

cheftako: client_connections seems good for the request level connections.

Still need something for the higher level connection.

konnectivity-client/pkg/client/client.go
jkh52: Thanks, that's perfect here

konnectivity-client/pkg/client/client.go
jkh52: Added some guards using the done channel.

konnectivity-client/pkg/client/client.go
jkh52: Used atomic.Value.

jkh52:
> client_connections seems good for the request level connections.
> 
> Still need something for the higher level connection.

For gRPC level, I would use a separate metric, likely one that matches `konnectivity-server`. References:
* [server](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/server/metrics/metrics.go#L91-L101)
* [agent](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/agent/metrics/metrics.go#L91-L99)

I will put that in a separate PR.

PR #448, opened 2022-12-28
Agent: increase log quality.

jkh52:
Decrease an important error level log to info.

Add more structured log info.

Remove some duplicate log errors in clientset.go

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -235,7 +235,7 @@ func (a *Client) Connect() (int, error) {
 		if ctx, err = a.initializeAuthContext(ctx); err != nil {
 			err := conn.Close()
 			if err != nil {
-				klog.ErrorS(err, "failed to close connection")
+				klog.ErrorS(err, "failed to close gRPC connection", "agentID", a.agentID)
 			}
 			return 0, err
 		}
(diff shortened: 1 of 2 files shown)


pkg/agent/client.go
```diff
@@ -458,10 +459,12 @@ func (a *Client) Serve() {
 						reason = metrics.DialFailureTimeout
 					}
 					metrics.Metrics.ObserveDialFailure(reason)
-					klog.ErrorS(err, "error dialing backend", "dialID", dialReq.Random, "connectionID", connID, "dialAddress", dialReq.Address)
+					// Do not log agent errors for remote unavailable.
+					kvs := []interface{}{"error", err, "dialID", dialReq.Random, "connectionID", connID, "dialAddress", dialReq.Address}
```
dberkov, line 463: out of curiosity,  why defining "kvs" variable is better than passing params to InfoS as is?

jkh52: You're right, simplified.

(I copied the pattern [here](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/konnectivity-client/pkg/client/client.go#L280) but in this case there is no branching decision.)

PR #450, opened 2023-01-11
proxy-server: add a dial failure metric reason for "no agent".

jkh52:
Previous to this change, `proxy-server` would not emit any dial failure metric in this case, and the [client metric](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/9ab611d9c0fdac24c838ed1903e41738b3594675/konnectivity-client/pkg/client/metrics/metrics.go#L50-L65) would also not cleanly identify it.

### DIFF
--- a/pkg/server/metrics/metrics.go
+++ b/pkg/server/metrics/metrics.go
@@ -230,6 +230,7 @@ func (s *ServerMetrics) FullRecvChannel(serviceMethod string) prometheus.Gauge {
 type DialFailureReason string
 
 const (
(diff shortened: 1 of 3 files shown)


jkh52:
Note that agent naming is preferred over backend, the user visible error is "No agent available": https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/9ab611d9c0fdac24c838ed1903e41738b3594675/pkg/server/backend_manager.go#L283

tallclair (commented):
/lgtm

PR #451, opened 2023-01-17
konnectivity-client: add Send and Recv wrappers to simplify metrics.

jkh52:
konnectivity-client: add Send and Recv wrappers to simplify metrics.

Also add the same locking pattern that konnectivity-agent does. Note that we do have separate goroutines (across client.go and conn.go) accessing the stream.

### DIFF
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -118,6 +118,8 @@ func (cm *connectionManager) closeAll() {
 // grpcTunnel implements Tunnel
 type grpcTunnel struct {
 	stream      client.ProxyService_ProxyClient
+	sendLock    sync.Mutex
+	recvLock    sync.Mutex
 	clientConn  clientConn
 	pendingDial pendingDialManager
 	conns       connectionManager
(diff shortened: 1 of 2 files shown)


konnectivity-client/pkg/client/client.go
```diff
@@ -243,20 +245,17 @@ func (t *grpcTunnel) serve(tunnelCtx context.Context) {
...
-		const segment = commonmetrics.SegmentToClient
 		isClosing := t.isClosing()
 		if err != nil || pkt == nil {
 			if !isClosing {
 				klog.ErrorS(err, "stream read failure")
 			}
-			metrics.Metrics.ObserveStreamErrorNoPacket(segment, err)
 			return
 		}
-		metrics.Metrics.ObservePacket(segment, pkt.Type)
```
tallclair, line 259: We lost this metric in this PR.

jkh52: Good catch, sent https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/453

(It became missing because I copied from the agent wrapper, where it was also missing...)

PR #453, opened 2023-01-18
Metrics bugfix: restore ObservePacket.

jkh52:
This was lost in the client at https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/451, but was missing in the agent wrapper.

### DIFF
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -501,11 +501,16 @@ func (t *grpcTunnel) Recv() (*client.Packet, error) {
 	t.recvLock.Lock()
 	defer t.recvLock.Unlock()
 
+	const segment = commonmetrics.SegmentToClient
 	pkt, err := t.stream.Recv()
 	if err != nil && err != io.EOF {
-		metrics.Metrics.ObserveStreamErrorNoPacket(commonmetrics.SegmentToClient, err)
+		metrics.Metrics.ObserveStreamErrorNoPacket(segment, err)
 	}
-	return pkt, err
+	if err != nil {
+		return pkt, err
+	}
+	metrics.Metrics.ObservePacket(segment, pkt.Type)
+	return pkt, nil
 }
 
 func GetDialFailureReason(err error) (isDialFailure bool, reason metrics.DialFailureReason) {
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -293,12 +293,17 @@ func (a *Client) Recv() (*client.Packet, error) {
 	a.recvLock.Lock()
 	defer a.recvLock.Unlock()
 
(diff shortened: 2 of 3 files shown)


jkh52:
Regarding the `stream_errors_total` metric and handling of stream io.EOF:
* RecvMsg: Both ServerStream and ClientStream use it to indicate successful stream completion
* Only ClientStream SendMsg mentions io.EOF (and when it happens, RecvMsg should give stream status)

https://pkg.go.dev/google.golang.org/grpc#ServerStream
https://pkg.go.dev/google.golang.org/grpc#ClientStream

So I'm tempted to take another pass and make Send treatment of io.EOF more consistent; I think we should skip the errors metric on io.EOF for both directions. (But the most important thing here is restoring the lost ObservePacket).

/hold

konnectivity-client/pkg/client/client.go
```diff
@@ -501,11 +501,16 @@ func (t *grpcTunnel) Recv() (*client.Packet, error) {
...
 	defer t.recvLock.Unlock()
 
+	const segment = commonmetrics.SegmentToClient
 	pkt, err := t.stream.Recv()
 	if err != nil && err != io.EOF {
-		metrics.Metrics.ObserveStreamErrorNoPacket(commonmetrics.SegmentToClient, err)
+		metrics.Metrics.ObserveStreamErrorNoPacket(segment, err)
 	}
-	return pkt, err
+	if err != nil {
```
tallclair, line 509: This makes me wonder whether we should record packets as sent if there's a stream error...

pkg/agent/client.go
```diff
@@ -293,12 +293,17 @@ func (a *Client) Recv() (*client.Packet, error) {
...
 	if err != nil && err != io.EOF {
 		metrics.Metrics.ObserveServerFailureDeprecated(metrics.DirectionFromServer)
-		metrics.Metrics.ObserveStreamErrorNoPacket(commonmetrics.SegmentToAgent, err)
+		metrics.Metrics.ObserveStreamErrorNoPacket(segment, err)
 	}
-	return pkt, err
+	if err != nil {
+		return nil, err
+	}
+	metrics.Metrics.ObservePacket(segment, pkt.Type)
```
tallclair, line 305: I think this is double counting with the observe on line 398?

konnectivity-client/pkg/client/client.go
jkh52: I understand your point, but to make comparisons (sent vs received) I prefer the meaning "desired" over "actual", and doing so would lose the former.

pkg/agent/client.go
jkh52: Thanks, fixed.

tallclair (commented):
/lgtm

PR #456, opened 2023-01-19
konnectivity-client: code cleanup.

jkh52:
Add a guard for duplicate Close().
Clarify gRPC connection cleanup.

### DIFF
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -120,7 +120,7 @@ type grpcTunnel struct {
 	stream      client.ProxyService_ProxyClient
 	sendLock    sync.Mutex
 	recvLock    sync.Mutex
-	clientConn  clientConn
+	grpcConn    clientConn
 	pendingDial pendingDialManager
 	conns       connectionManager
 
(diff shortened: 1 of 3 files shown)


konnectivity-client/pkg/client/client.go
```diff
@@ -453,11 +445,17 @@ func (t *grpcTunnel) dialContext(requestCtx context.Context, protocol, address s
...
+			t.closeTunnel()
+		}()
 		return nil, &dialFailure{"dial timeout, backstop", metrics.DialFailureTimeout}
 	case <-requestCtx.Done():
 		klog.V(5).InfoS("Context canceled waiting for DialResp", "ctxErr", requestCtx.Err(), "dialID", random)
-		go t.closeDial(random)
+		go func() {
+			t.sendDialClose(random)
+			t.closeTunnel()
+		}()
```
tallclair, line 458: @cheftako had a concern about what happens if send panics. Options:
1. Add a recover to `t.Send` to prevent panics from bubbling up
2. Call `t.closeTunnel()` from a defer, i.e.

```suggestion
		go func() {
			defer t.closeTunnel()
			t.sendDialClose(random)
		}()
```

jkh52: Done

PR #460, opened 2023-02-11
proxy-server: synchronization improvements

jkh52:
Synchronization improvements:
- replace looping over removeFrontend with a single removeFrontendsForBackendConn.
- fix PendingDial race.


jkh52:
This does not fix the failing test (as I originally thought) at https://github.com/tallclair/apiserver-network-proxy/tree/leak-repro, invoked as `go test tests/proxy_test.go -run TestProxyDial_RequestCancelled_Concurrent_GRPC -count=<large N>`.

The remaining reproducible memory leak seems to be straightforward:

* client thinks dial is pending and sends DIAL_CLS to cancel
* proxy-server has finished dial, PendingDial no longer contains the dialID and DIAL_CLS has no effect.

PR #463, opened 2023-02-15
konnectivity-server: synchronization cleanup.

jkh52:
Guard frontend gRPC stream send with mutex, since both serveRecvFrontend and serveRecvBackend use it. Guard receive just for consistency.

Always use Backend interface to send to and receive from agent. This allows some metrics duplication cleanup.

### DIFF
--- a/pkg/server/backend_manager.go
+++ b/pkg/server/backend_manager.go
@@ -72,6 +72,7 @@ func GenProxyStrategiesFromStr(proxyStrategies string) ([]ProxyStrategy, error) 
 
 type Backend interface {
 	Send(p *client.Packet) error
+	Recv() (*client.Packet, error)
 	Context() context.Context
 }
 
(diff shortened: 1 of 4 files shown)


pkg/server/backend_manager.go
```diff
@@ -81,13 +82,15 @@ var _ Backend = agent.AgentService_ConnectServer(nil)
...
+	readLock sync.Mutex
+	recvLock sync.Mutex
+	conn     agent.AgentService_ConnectServer
 }
 
 func (b *backend) Send(p *client.Packet) error {
-	b.mu.Lock()
-	defer b.mu.Unlock()
+	b.readLock.Lock()
+	defer b.readLock.Unlock()
```
tallclair, line 92: why is this called `readLock`? I'd expect `sendLock`. It seems weird to guard `Send` with a read lock...

pkg/server/backend_manager.go
```diff
@@ -97,6 +100,22 @@ func (b *backend) Send(p *client.Packet) error {
...
+	defer b.recvLock.Unlock()
+
+	const segment = commonmetrics.SegmentFromAgent
+	pkt, err := b.conn.Recv()
+	if err != nil && err != io.EOF {
+		metrics.Metrics.ObserveStreamErrorNoPacket(segment, err)
+	}
+	if err != nil {
+		return nil, err
+	}
```
tallclair, line 114: nit: this would be a bit easier to follow up it were restructured like this:
```suggestion
	if err != nil {
		if err != io.EOF {
			metrics.Metrics.ObserveStreamErrorNoPacket(segment, err)
		}
		return nil, err
	}
```

pkg/server/server.go
```diff
@@ -48,10 +48,45 @@ const xfrChannelSize = 10
...
+	defer g.recvLock.Unlock()
+
+	const segment = commonmetrics.SegmentFromClient
+	pkt, err := g.stream.Recv()
+	if err != nil && err != io.EOF {
+		metrics.Metrics.ObserveStreamErrorNoPacket(segment, err)
+	}
+	if err != nil {
+		return nil, err
+	}
```
tallclair, line 81: Same here, easier to follow with nested if-blocks

pkg/server/backend_manager.go
jkh52: I think my brain meant send but fingers typed read. Fixed.

pkg/server/backend_manager.go
jkh52: Done, including elsewhere.

PR #464, opened 2023-02-17
Separate HTTP Connect serving to a separate package

tallclair:
First 3 commits:

1. Encapsulate different handling of grpc & http frontends behind a `Frontend` interface so that `ProxyClientConnections` no longer needs to be aware of the serving mode
2. Extract common (between grpc & http-connect) dial logic to `NewPendingDial` - this deduplicates some code, and means we don't need to expose as much of the `Server` internals
3. Actual refactor - Required exposing a few additional `ProxyClientConnection` fields. Also fixed some race conditions when HTTP connection is prematurely closed.

This PR ended up a lot larger than I intended...


tallclair:
/hold
For further discussion.

Reflecting on this more, I'm not sure this is worth the level of changes it ended up being. I think the first commit is a good cleanup & worth taking, but I'm not sure about the latter 2.

PR #452, opened 2023-01-17
konnectivity-client: guard single use tunnels from duplicate dial.

jkh52:
konnectivity-client: guard single use tunnels from duplicate dial.

The only client is egress_selector.go, and the only expected fallout from this guard is the concurrent_test.go change (locally).

PR #467, opened 2023-02-17
proxy-server: improve connection cleanup on client cancel.

jkh52:
Before this change, cleanup happened at the end of `serveRecvFrontend` and depended on first witnessing a DATA packet and remembering `ConnectID`. After this change, `proxy-server` will send a best effort cleanup signal even if no DATA packet has been sent (as well as supporting both pending and connected dial states).

This approach eliminates one use of the `Backend, firstConnID, err` at the top of `serveRecvFrontend`, and in the long term future we should strive to reduce that to zero, which would make Proxy RPC capable of supporting multiple tunnel TCP connections.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -27,6 +27,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/google/uuid"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
@@ -49,9 +50,10 @@ const xfrChannelSize = 10
 type key int
 
 type GrpcFrontend struct {
-	stream   client.ProxyService_ProxyServer
-	sendLock sync.Mutex
-	recvLock sync.Mutex
+	stream    client.ProxyService_ProxyServer
+	streamUID string
+	sendLock  sync.Mutex
+	recvLock  sync.Mutex
 }
 
 func (g *GrpcFrontend) Send(pkt *client.Packet) error {
@@ -89,6 +91,7 @@ type ProxyClientConnection struct {
 	frontend    *GrpcFrontend
 	CloseHTTP   func() error
 	connected   chan struct{}
+	dialID      int64
 	connectID   int64
 	agentID     string
 	start       time.Time
@@ -151,6 +154,28 @@ func (pm *PendingDialManager) Remove(random int64) *ProxyClientConnection {
 	delete(pm.pendingDial, random)
 	metrics.Metrics.SetPendingDialCount(len(pm.pendingDial))
 	return pd
+}
+
+// removeForStream removes and returns all pending ProxyClientConnection associated with a
+// given Proxy gRPC connection (expected to be at most 1 while konnectivity-client API gives
+// single-use tunnels).
+func (pm *PendingDialManager) removeForStream(streamUID string) []*ProxyClientConnection {
+	var ret []*ProxyClientConnection
+	if streamUID == "" {
+		return ret
+	}
+	pm.mu.Lock()
+	defer pm.mu.Unlock()
+	for dialID, frontend := range pm.pendingDial {
+		if frontend.frontend == nil {
+			continue
+		}
+		if frontend.frontend.streamUID == streamUID {
+			delete(pm.pendingDial, dialID)
+			ret = append(ret, frontend)
+		}
+	}
+	return ret
 }
 
 // ProxyServer
@@ -345,6 +370,9 @@ func (s *ProxyServer) getFrontend(agentID string, connID int64) (*ProxyClientCon
 
 func (s *ProxyServer) removeFrontendsForBackendConn(agentID string, backend Backend) ([]*ProxyClientConnection, error) {
 	var ret []*ProxyClientConnection
+	if backend == nil {
+		return ret, nil
+	}
 	s.fmu.Lock()
 	defer s.fmu.Unlock()
 	frontends, ok := s.frontends[agentID]
@@ -368,6 +396,33 @@ func (s *ProxyServer) getCount(frontends map[string](map[int64]*ProxyClientConne
 		count = count + len(frontend)
 	}
 	return count
+}
+
+// removeForStream removes and returns all established ProxyClientConnection associated with a given
+// Proxy gRPC connection (expected to be at most 1 while konnectivity-client API gives single-use
+// tunnels).
+func (s *ProxyServer) removeFrontendsForStream(streamUID string) []*ProxyClientConnection {
+	var ret []*ProxyClientConnection
+	if streamUID == "" {
+		return ret
+	}
+	s.fmu.Lock()
+	defer s.fmu.Unlock()
+	for agentID, frontends := range s.frontends {
+		for connID, frontend := range frontends {
+			if frontend.frontend == nil {
+				continue
+			}
+			if frontend.frontend.streamUID == streamUID {
+				delete(frontends, connID)
+				ret = append(ret, frontend)
+			}
+		}
+		if len(frontends) == 0 {
+			delete(s.frontends, agentID)
+		}
+	}
+	return ret
 }
 
 // NewProxyServer creates a new ProxyServer instance
@@ -409,14 +464,32 @@ func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
 		return fmt.Errorf("failed to get context")
 	}
 	userAgent := md.Get(header.UserAgent)
-	klog.V(5).InfoS("Proxy request from client", "userAgent", userAgent, "serverID", s.serverID)
+	streamUID := uuid.New().String()
+	klog.V(5).InfoS("Proxy request from client", "userAgent", userAgent, "serverID", s.serverID, "streamUID", streamUID)
 
 	recvCh := make(chan *client.Packet, xfrChannelSize)
 	stopCh := make(chan error, 1)
 
 	frontend := GrpcFrontend{
-		stream: stream,
+		stream:    stream,
+		streamUID: streamUID,
 	}
+
+	defer func() {
+		// The frontend stream and goroutines are completely shut down, so we will never get a
+		// subsequent close message from the frontend. This can happen for several reasons, such as
+		// stream canceled. Clean any lingering tunnel connections to avoid resource leaks.
+		for _, p := range s.PendingDial.removeForStream(streamUID) {
+			klog.V(2).InfoS("frontend stream shutdown, cleaning dial", "dialID", p.dialID)
+			// TODO: add agent support to handle this
+			s.sendBackendDialClose(p.backend, p.dialID, "frontend stream shutdown")
+		}
+		for _, f := range s.removeFrontendsForStream(streamUID) {
+			klog.V(2).InfoS("frontend stream shutdown, cleaning frontend", "connectionID", f.connectID, "dialID", f.dialID)
+			s.sendBackendClose(f.backend, f.connectID, f.dialID, "frontend stream shutdown")
+		}
+	}()
+
 	labels := runpprof.Labels(
 		"serverCount", strconv.Itoa(s.serverCount),
 		"userAgent", strings.Join(userAgent, ", "),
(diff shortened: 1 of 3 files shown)


pkg/server/server.go
```diff
@@ -153,6 +156,22 @@ func (pm *PendingDialManager) Remove(random int64) *ProxyClientConnection {
 	return pd
 }
 
+func (pm *PendingDialManager) RemoveForStream(uid string) []*ProxyClientConnection {
```
tallclair, line 159: nit: let's stick with `streamUID`
```suggestion
func (pm *PendingDialManager) RemoveForStream(streamUID string) []*ProxyClientConnection {
```

pkg/server/server.go
```diff
@@ -357,6 +376,24 @@ func (s *ProxyServer) removeFrontendsForBackendConn(agentID string, backend Back
 	return ret, nil
 }
 
+func (s *ProxyServer) removeFrontendsForStream(uuid string) []*ProxyClientConnection {
```
tallclair, line 379: same here: `streamUID`

pkg/server/server.go
```diff
@@ -357,6 +376,24 @@ func (s *ProxyServer) removeFrontendsForBackendConn(agentID string, backend Back
...
+	var ret []*ProxyClientConnection
+	s.fmu.Lock()
+	defer s.fmu.Unlock()
+	for agentID, frontends := range s.frontends {
+		for connID, frontend := range frontends {
```
tallclair, line 384: I think we're a bit inconsistent with what `frontend` refers to in the code. Maybe not here, but at some point we should probably audit the existing usages and come up with less ambiguous terms.

pkg/server/server.go
```diff
@@ -595,25 +651,9 @@ func (s *ProxyServer) serveRecvFrontend(frontend *GrpcFrontend, recvCh <-chan *c
 	}
 
 	// TODO: Log an error if the frontend stream is closed with an established tunnel still open.
```
tallclair, line 653: Is this TODO still relevant?

pkg/server/server.go
```diff
@@ -595,25 +651,9 @@ func (s *ProxyServer) serveRecvFrontend(frontend *GrpcFrontend, recvCh <-chan *c
...
-	klog.V(5).InfoS("Close streaming", "connectionID", firstConnID)
-
+	klog.V(5).InfoS("Close frontend streaming", "connectionID", firstConnID)
```
tallclair, line 654: nit: log the streamUID?

pkg/server/server.go
```diff
@@ -595,25 +651,9 @@ func (s *ProxyServer) serveRecvFrontend(frontend *GrpcFrontend, recvCh <-chan *c
...
-					ConnectID: firstConnID,
-				},
-			},
-		}
-		if err := backend.Send(pkt); err != nil {
-			// TODO: Determine whether this is expected, and log at error level if appropriate.
-			// If that connection shutdown was innitiated by the backend, this is expected.
-			// If initiated by the frontend, it is not.
-			klog.V(2).InfoS("CLOSE_REQ to Backend failed", "error", err)
-		}
```
tallclair, line 616: Why delete this? Is the idea just to reduce the places we close the backend by deferring the close to after the FE stream is closed?

pkg/server/server.go
jkh52: Agreed. When I added `GrpcFrontend` I had in mind future renaming. Maybe:

`ProxyClientConnection` -> `ProxyConnection`
`frontends` -> `proxyConnections`

pkg/server/server.go
jkh52: Avoid duplication (both server code and duplicate close messages).

This handling is moved from the bottom of `serveRecvFrontend` (just before exit) to a `Proxy` defer, which isn't very far away.

pkg/server/server.go
jkh52: Removed here and another.

pkg/server/server.go
jkh52: Done.

jkh52:
Using https://github.com/tallclair/apiserver-network-proxy/tree/leak-repro (with counts 10, 25) this seems effective.

pkg/server/server.go
```diff
@@ -396,14 +445,32 @@ func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
...
+		// subsequent close message from the frontend. This can happen for several reasons, such as
+		// stream canceled. Clean any lingering tunnel connections to avoid resource leaks.
+		for _, p := range s.PendingDial.RemoveForStream(streamUID) {
+			klog.V(2).InfoS("frontend stream shutdown, cleaning dial", "dialID", p.dialID)
+			// TODO: add agent support to handle this
+			s.sendBackendDialClose(p.backend, p.dialID, "frontend stream shutdown")
+		}
+		for _, f := range s.removeFrontendsForStream(streamUID) {
+			klog.V(2).InfoS("frontend stream shutdown, cleaning frontend", "connectionID", f.connectID, "dialID", f.dialID)
+			s.sendBackendClose(f.backend, f.connectID, f.dialID, "frontend stream shutdown")
```
tallclair, line 470: It probably doesn't matter much while there's a single connection, but should this (and `sendBackendDialClose` above) be called in a go routine, so they don't block terminating the frontend connection?

tallclair: I guess that would undo the change that prevents this method from returning until all associated go-routines have terminated...

pkg/server/server.go
```diff
@@ -594,26 +661,9 @@ func (s *ProxyServer) serveRecvFrontend(frontend *GrpcFrontend, recvCh <-chan *c
 		}
 	}
 
-	// TODO: Log an error if the frontend stream is closed with an established tunnel still open.
-	klog.V(5).InfoS("Close streaming", "connectionID", firstConnID)
-
+	klog.V(5).InfoS("Close frontend streaming", "connectionID", firstConnID, "streamUID", frontend.streamUID)
 	if backend == nil {
 		klog.V(2).InfoS("Streaming closed before backend intialized")
```
tallclair, line 666: should this be moved to the defer at the top of this method? There are a bunch of places we exit the loop with a return, which would skip these log lines.

pkg/server/server.go
jkh52: I could imagine continuing to block Proxy returning on both stream handling goroutines shutting down, and "leak" one best effort cleanup goroutine.

But I prefer not, because it makes unit tests a bit more non-deterministic. And there is something nice about "all work is done" on Proxy finish (and monitoring the Proxy gauge metric).

pkg/server/server.go
jkh52: Good idea, done.

pkg/server/server.go
tallclair: The only issue would be if the backend stream was backed up, then send could block, and hold this connection open. I doubt that would turn into a permanent leak though...

tallclair:
/lgtm
/hold

I can't remember of @cheftako was going to look at this. Feel free to remove hold if not.

jkh52:
> I can't remember of @cheftako was going to look at this. Feel free to remove hold if not.

@cheftako I was hoping to get this into a release to test effectiveness.

jkh52:
@tallclair 

After un-skipping the `TestProxyDial_RequestCancelled_Concurrent_GRPC` my overall unit test time ~doubles from about 30s to 1m5s. I lowered `concurrentConns` from ~100 to ~50 to improve.

jkh52:
/lgtm

(adding back LGTM that was lost due to merge conflict resolution and trivially un-skipping the new test).

jkh52:
Resolved server_test.go merge conflicts only; re-adding LGTM.

jkh52:
Test flake on `TestBasicProxy_GRPC`:

```
--- FAIL: TestBasicProxy_GRPC (0.59s)
    proxy_test.go:979: found unexpected goroutines:
        [Goroutine 2072 in state select, with sigs.k8s.io/apiserver-network-proxy/konnectivity-client/pkg/client.(*conn).Close on top of the stack:
        goroutine 2072 [select]:
        sigs.k8s.io/apiserver-network-proxy/konnectivity-client/pkg/client.(*conn).Close(0xc0021807d0)
        	/home/prow/go/src/sigs.k8s.io/apiserver-network-proxy/konnectivity-client/pkg/client/conn.go:155 +0x752
        net/http.(*persistConn).closeLocked(0xc0023ba120, {0x254aea0, 0xc000112c30})
        	/usr/local/go/src/net/http/transport.go:2714 +0x21c
        net/http.(*persistConn).close(0xc0023ba120, {0x254aea0, 0xc000112c30})
        	/usr/local/go/src/net/http/transport.go:2699 +0xa9
        net/http.(*persistConn).readLoop.func1()
        	/usr/local/go/src/net/http/transport.go:2061 +0x50
        net/http.(*persistConn).readLoop(0xc0023ba120)
        	/usr/local/go/src/net/http/transport.go:2[236](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_apiserver-network-proxy/467/pull-apiserver-network-proxy-test/1628921105492217856#1:build-log.txt%3A236) +0x1029
        created by net/http.(*Transport).dialConn
        	/usr/local/go/src/net/http/transport.go:1750 +0x24c5
```

The test fails after 0.59s, but the line number the leaked goroutine is on waiting up to 10s (`CloseTimeout`) to receive close response.

```
    select {
    case errMsg := <-c.closeCh:
        if errMsg != "" {
            return errors.New(errMsg)
        }
        return nil
    case <-time.After(CloseTimeout):
    }
```

I updated this PR to have the test block on closing the client connection, it seems to fix the flake, at least for `go test ./... -run TestBasicProxy_GRPC -count=500` (failed before, passes after).

pkg/server/server.go
```diff
@@ -153,6 +156,25 @@ func (pm *PendingDialManager) Remove(random int64) *ProxyClientConnection {
 	return pd
 }
 
+func (pm *PendingDialManager) RemoveForStream(streamUID string) []*ProxyClientConnection {
```
cheftako, line 159: Would be nice to have this documented. Especially that the result is a slice of the removed streams.

pkg/server/server.go
```diff
@@ -153,6 +156,25 @@ func (pm *PendingDialManager) Remove(random int64) *ProxyClientConnection {
...
+	if streamUID == "" {
+		return ret
+	}
+	pm.mu.Lock()
+	defer pm.mu.Unlock()
+	for dialID, frontend := range pm.pendingDial {
+		if frontend.frontend == nil {
+			continue
+		}
+		if frontend.frontend.streamUID == streamUID {
```
cheftako, line 170: If streamUID is a UID, when would we get multiple matches?

pkg/server/server.go
cheftako: Also why are we making this public?

pkg/server/server.go
cheftako: +1. I know we already discussed this but I think we should document some of the terms and then audit to make sure that our usage matches our definitions.

For me:

- Frontend -- The client (KAS) from the Konn servers perspective
- Backend -- The agent from the Konn servers perspective
- Tunnel -- The "outer" TCP/IP connection. Generally equivalent to the GRPC stream (or http-connect). We generally have 2 tunnel segments. One from client to konn server. Another from konn server to konn agent.
- Connection -- Need a better name. The "inner"  TCP/IP connection. The connectioned request from the client to the destination.

pkg/server/server.go
jkh52: > Would be nice to have this documented.

Done.

> Also why are we making this public?

I did it for consistency but agree it can be kept private; done.

pkg/server/server.go
jkh52: Future proofing for the mux case.

And although we added a client library guard https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/452, it is still worth handling here this since 1. the client may be at an older version and 2. the protocol allows receiving multiple DIAL_REQ (and create multiple dials/connections).

pkg/server/server.go
jkh52: I will take this documentation (and rename) to a followup.

PR #470, opened 2023-02-22
Add server "established_connections" metric.

jkh52:
Comparing "grpc_connections" metric with this can help when debugging resource leaks.

### DIFF
--- a/pkg/server/metrics/metrics.go
+++ b/pkg/server/metrics/metrics.go
@@ -29,15 +29,6 @@ const (
 	Namespace = "konnectivity_network_proxy"
 	Subsystem = "server"
 
-	DialLatencyMetric        = "dial_duration_seconds"
-	FrontendLatencyMetric    = "frontend_write_duration_seconds"
-	GRPCConnectionsMetric    = "grpc_connections"
-	HTTPConnectionsMetric    = "http_connections"
-	BackendConnectionsMetric = "ready_backend_connections"
-	PendingDialsMetric       = "pending_backend_dials"
-	FullRecvChannelMetric    = "full_receive_channels"
-	DialFailuresMetric       = "dial_failure_count"
-
 	// Proxy is the ProxyService method used to handle incoming streams.
 	Proxy = "Proxy"
 	// Connect is the AgentService method used to establish next hop.
@@ -60,6 +51,7 @@ type ServerMetrics struct {
 	httpConnections   prometheus.Gauge
 	backend           *prometheus.GaugeVec
 	pendingDials      *prometheus.GaugeVec
+	establishedConns  *prometheus.GaugeVec
 	fullRecvChannels  *prometheus.GaugeVec
 	dialFailures      *prometheus.CounterVec
 	streamPackets     *prometheus.CounterVec
@@ -72,7 +64,7 @@ func newServerMetrics() *ServerMetrics {
 		prometheus.HistogramOpts{
 			Namespace: Namespace,
 			Subsystem: Subsystem,
-			Name:      DialLatencyMetric,
+			Name:      "dial_duration_seconds",
 			Help:      "Latency of dial to the remote endpoint in seconds",
 			Buckets:   latencyBuckets,
 		},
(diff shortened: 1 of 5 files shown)


pkg/server/server.go
```diff
@@ -354,9 +357,19 @@ func (s *ProxyServer) removeFrontendsForBackendConn(agentID string, backend Back
...
+
+	metrics.Metrics.SetProxyConnCount(s.getCount(s.frontends))
 	return ret, nil
 }
 
+func (s *ProxyServer) getCount(frontends map[string](map[int64]*ProxyClientConnection)) int {
+	count := 0
+	for _, frontend := range frontends {
+		count = count + len(frontend)
+	}
```
tallclair, line 369: This is sort of unfortunate... I guess it's not that bad since the number of agents is relatively small, but would it be better to just increment & decrement the count? I imagine this is being done for safety, but we have a small number of methods adding & removing connections. You could even extract the frontend connections to a separate object (`FrontendConnectionManager`?) if you wanted to make it really explicit.

pkg/server/metrics/metrics.go
```diff
@@ -120,16 +112,25 @@ func newServerMetrics() *ServerMetrics {
...
+			Name:      "pending_backend_dials",
 			Help:      "Current number of pending backend dial requests",
 		},
 		[]string{},
 	)
+	proxyConnections := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: Namespace,
+			Subsystem: Subsystem,
+			Name:      "proxy_connections",
```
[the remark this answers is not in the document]
jkh52: >  I was expecting it to count the number of open calls to the Proxy method

The current protocol allows one Proxy bidirectional stream to open multiple proxy connections /  tunnels (via DIAL_REQ). So to me, the name proxy_connection is accurate for each of these (and the reason for multiple metrics).

I've seen the word "tunnel" overloaded. The [client proto](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/konnectivity-client/proto/client/client.proto) mentions connections (per dial response), does not yet mention tunnels.

Long term (post identifier re-keying) I think both the pending dials metric and this new metric (whatever name we decide) should be deprecated by a single metric that has a state label (dialing, connected, closing, etc).

pkg/server/server.go
jkh52: Both this and https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/467 touch `frontends`, and we want to rename it. If we want to use Inc/Dec, to keep accuracy I would add handling for duplicates. Maybe we can defer a rename + hygiene improvement to immediately after?

jkh52: There is a larger refactor we may want to consider: a single storage / manager in front of both pending dials and active connections, with the same mutex. That would buy some protection (until identifier re-keying) against remaining race condition.

pkg/server/metrics/metrics.go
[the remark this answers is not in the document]
jkh52: Per offline discussion the name "established_connections" seems agreeable, I will rename here.

pkg/server/metrics/metrics.go
```diff
@@ -120,16 +112,25 @@ func newServerMetrics() *ServerMetrics {
...
 			Help:      "Current number of pending backend dial requests",
 		},
 		[]string{},
 	)
+	establishedConns := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: Namespace,
+			Subsystem: Subsystem,
+			Name:      "established_connections",
+			Help:      "Current number of established connections (post-dial).",
```
tallclair, line 125: nit: can we add "end-to-end" in the description?
```suggestion
			Help:      "Current number of established end-to-end connections (post-dial).",
```

tallclair (commented):
/lgtm
/hold

pkg/server/metrics/metrics.go
jkh52: Done.

jkh52:
/remove-hold

(@tallclair  needs fresh LGTM)

PR #471, opened 2023-02-24
proy-server: always emit backends count metric

jkh52:
This reduces friction to detect `proxy-server` that never had any agent successfully connect.

### DIFF
--- a/pkg/server/backend_manager.go
+++ b/pkg/server/backend_manager.go
@@ -194,6 +194,9 @@ func NewDefaultBackendManager() *DefaultBackendManager {
 
 // NewDefaultBackendStorage returns a DefaultBackendStorage
 func NewDefaultBackendStorage(idTypes []pkgagent.IdentifierType) *DefaultBackendStorage {
+	// Set an explicit value, so that the metric is emitted even when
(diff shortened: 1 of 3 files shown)


pkg/server/backend_manager.go
```diff
@@ -194,6 +194,7 @@ func NewDefaultBackendManager() *DefaultBackendManager {
 
 // NewDefaultBackendStorage returns a DefaultBackendStorage
 func NewDefaultBackendStorage(idTypes []pkgagent.IdentifierType) *DefaultBackendStorage {
+	metrics.Metrics.SetBackendCount(0)
```
tallclair, line 197: nit: add a comment for why this is being done.

tallclair (commented):
/lgtm
/hold

pkg/server/backend_manager.go
jkh52: Done.

/lgtm

jkh52:
I found https://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics which implies this should not be necessary. But I tried the latest version 1.14.0 of github.com/prometheus/client_golang and makes no difference.

Issue #443, opened 2022-12-21
Konnectivity server fails to restart when UDS socket already exists after having been killed forcefully


Nirusu:
Using Konnectivity almost 1:1 as described here in the examples here with UDS: https://kubernetes.io/docs/tasks/extend-kubernetes/setup-konnectivity/

When the UDS `konnectivity-server.socket` already exists and the Konnectivity server is forcefully killed on the host (e.g. killing all Konnectivity process PIDs with `kill -9`), the Konnectivity server will fail to restart since it's trying to re-create the UDS socket, but fails due to it already existing and not having been cleaned up correctly:

```
E1219 09:16:02.451736       1 main.go:49] error: failed to run the frontend server: failed to get uds listener: failed to listen(unix) name /etc/kubernetes/konnectivity-server/konnectivity-server.socket: listen unix /etc/kubernetes/konnectivity-server/konnectivity-server.socket: bind: address already in use
```

Manually removing `konnectivity-server.socket` seems to fix this issue.

Full logs:
```
I1219 09:16:02.447539       1 options.go:124] ServerCert set to "".
I1219 09:16:02.447627       1 options.go:125] ServerKey set to "".
I1219 09:16:02.447633       1 options.go:126] ServerCACert set to "".
I1219 09:16:02.447639       1 options.go:127] ClusterCert set to "/etc/kubernetes/pki/apiserver.crt".
I1219 09:16:02.447644       1 options.go:128] ClusterKey set to "/etc/kubernetes/pki/apiserver.key".
I1219 09:16:02.447648       1 options.go:129] ClusterCACert set to "".
I1219 09:16:02.447650       1 options.go:130] Mode set to "grpc".
I1219 09:16:02.447653       1 options.go:131] UDSName set to "/etc/kubernetes/konnectivity-server/konnectivity-server.socket".
I1219 09:16:02.447661       1 options.go:132] DeleteUDSFile set to false.
I1219 09:16:02.447664       1 options.go:133] Server port set to 0.
I1219 09:16:02.447669       1 options.go:134] Agent port set to 8132.
I1219 09:16:02.447672       1 options.go:135] Admin port set to 8133.
I1219 09:16:02.447674       1 options.go:136] Health port set to 8134.
I1219 09:16:02.447677       1 options.go:137] Keepalive time set to 1h0m0s.
I1219 09:16:02.447696       1 options.go:138] Frontend keepalive time set to 1h0m0s.
[...]

PR #476, opened 2023-03-09
delete stale socket file, if it exists

guettli:
we had issues were the server could not start because an old socket file was still around:

 [main.go:49] error: failed to run the frontend server: failed to get uds listener: failed to listen(unix) name /etc/kubernetes/konnectivity-server/konnectivity-server.socket: listen unix /etc/kubernetes/konnectivity-server/konnectivity-server.socket: bind: address already in use

### DIFF
--- a/cmd/server/app/server.go
+++ b/cmd/server/app/server.go
@@ -188,6 +188,7 @@ func getUDSListener(ctx context.Context, udsName string) (net.Listener, error) {
 	defer udsListenerLock.Unlock()
 	oldUmask := syscall.Umask(0007)
 	defer syscall.Umask(oldUmask)
+	os.Remove(udsName) // delete stale socket file, if it exists
 	var lc net.ListenConfig
 	lis, err := lc.Listen(ctx, "unix", udsName)
 	if err != nil {


cmd/server/app/server.go
```diff
@@ -188,6 +188,7 @@ func getUDSListener(ctx context.Context, udsName string) (net.Listener, error) {
 	defer udsListenerLock.Unlock()
 	oldUmask := syscall.Umask(0007)
 	defer syscall.Umask(oldUmask)
+	os.Remove(udsName) // delete stale socket file, if it exists
```
jkh52, line 191: This should already be supported by passing the flag `--delete-existing-uds-file=true` ([code reference](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/e162f71742aa5594f0aa12743ed9d7d8fb7c6bf7/cmd/server/app/server.go#L208))

jkh52: /hold

tallclair: Is there a reason we don't enable that flag by default?

jkh52: I suspect it is to avoid technically a backward incompatible change, but this has come up before (https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/443) and I would not be opposed since we're pre-1.0.

@cheftako WDYT?

guettli: @jkh52 @tallclair  I plan create a new PR which removes the option and always deletes the stale socket file, if it exists. I see no good reason to have an option to keep this file. Is this ok for you?

jkh52: > new PR which removes the option and always delete ...

That would have the down side of removing a flag, which is itself a breaking change (for operators upgrading patch versions).

I propose we:

1. update the Kubernetes docs as @batistein suggests (valuable in any case for existing image versions). I can take this task.

2. Separately we can change the flag default. Would you adapt this PR to do that? It looks like @tallclair and I would both agree; @cheftako do you have any concern with it? (Justification: we are pre-1.0, and can't think of a reasonable use case for the false value behavior).

jkh52: Sent https://github.com/kubernetes/website/pull/39977

jkh52:
I sent https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/498 to follow up on my last suggestion.

Once merged I would consider this resolved. Please remove `stale` label if you disagree.

cheftako:
Much prefer changing the default to blanket changing the behavior. Its not technically backward compatible, you can at least change the flags to restore the old behavior if needed.

jkh52:
Thank you @guettli. I will close as https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/498 is merged, and I believe there is not remaining work.

PR #477, opened 2023-03-10
Encapsulate grpc/http connection logic behind Frontend interface

tallclair:
Encapsulate different handling of grpc & http frontends behind a Frontend interface so that ProxyClientConnections no longer needs to be aware of the serving mode.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -17,12 +17,10 @@ limitations under the License.
 package server
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"io"
-	"net/http"
 	runpprof "runtime/pprof"
 	"strconv"
 	"strings"
@@ -51,14 +49,19 @@ const xfrChannelSize = 10
 
 type key int
 
-type GrpcFrontend struct {
+type Frontend interface {
+	Send(*client.Packet) error
+	Recv() (*client.Packet, error)
+}
+
+type grpcFrontend struct {
 	stream    client.ProxyService_ProxyServer
 	streamUID string
 	sendLock  sync.Mutex
 	recvLock  sync.Mutex
 }
 
-func (g *GrpcFrontend) Send(pkt *client.Packet) error {
+func (g *grpcFrontend) Send(pkt *client.Packet) error {
 	g.sendLock.Lock()
 	defer g.sendLock.Unlock()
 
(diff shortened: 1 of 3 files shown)


pkg/server/server.go
```diff
@@ -413,7 +400,12 @@ func (s *ProxyServer) removeFrontendsForStream(streamUID string) []*ProxyClientC
 			if frontend.frontend == nil {
 				continue
 			}
-			if frontend.frontend.streamUID == streamUID {
+			grpcFe, ok := frontend.frontend.(*grpcFrontend)
+			if !ok {
+				klog.V(5).ErrorS(nil, "Frontend is not a grpc Frontend", "connectID", frontend.connectID)
```
cheftako, line 405: This seems like noise. Prior to this change we used the continue in the previous if statement for http-connect. We still need to continue for http-connect but now http-connect is a frontend. So it needs to continue here but I don't think we need the klog.

pkg/server/tunnel.go
```diff
@@ -20,15 +20,55 @@ import (
...
+	closeOnce sync.Once
+	closed    chan struct{}
+}
+
+func (h *httpFrontend) Send(pkt *client.Packet) error {
+	const segment = commonmetrics.SegmentToClient
+	metrics.Metrics.ObservePacket(segment, pkt.Type)
+
+	if pkt.Type == client.PacketType_CLOSE_RSP {
+		h.close()
```
cheftako, line 45: Is there a reason we are dropping this potential error? I'm fine if we only return an error on the initial close and subsequent calls return null. (Eg https://go.dev/play/p/6C1_0QlYDRL). Ditto later places we are no longer returning errors.

cheftako:
I generally like the change but a little concerned about places we no longer return errors.

pkg/server/tunnel.go
tallclair: We never returned an error on close. The old `CloseHTTP` function just unconditionally returned `nil` (deleted code on old lines 89-93):

https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/60b5f0e02800910cbb363668e285ec0e78d7cae1/pkg/server/tunnel.go#L89-L93

tallclair:
This is ready for another look. See my response about inline response about omitting the errors.

PR #480, opened 2023-04-03
[WIP - Don't review yet] represent connection state in pod condition

ipochi:
Reviving an old PR #198.

This is quite helpful in terms of signal to users when there is network problem and the agent is unable to connect to the konnectivity server.

### DIFF
--- a/pkg/agent/clientset.go
+++ b/pkg/agent/clientset.go
@@ -18,6 +18,7 @@ package agent
 
 import (
 	"context"
+	"fmt"
 	"math"
 	runpprof "runtime/pprof"
 	"sync"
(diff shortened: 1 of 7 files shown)


pkg/agent/clientset.go
```diff
@@ -246,3 +273,72 @@ func (cs *ClientSet) shutdown() {
...
+	if current == 0 || desired == 0 {
+		condition.Status = corev1.ConditionFalse
+		condition.Reason = disconnectedPodConditionReason
+		return condition
+	}
+
+	if current < desired {
+		condition.Status = corev1.ConditionFalse
+		condition.Reason = partiallyConnectedPodConditionReason
+		condition.Message = fmt.Sprintf("connected to %d of %d servers", current, desired)
```
jveski, line 337: should we find a way to represent specific client errors? It would be useful to have at least a high-level indication of the client error e.g. connection timeout, dns can't resolve, etc. Although this may be difficult to implement.

jkh52:
My understanding is that https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/485 replaces this PR.

PR #485, opened 2023-04-12
agent: smarter readiness check

ipochi:
This commit introduces a smarter way for the readiness check.

Currently as soon as the agent is up, it signals itself as ready regardless of whether its actually connected to the server or not.

This commit introduces a callback function and a connected field in ClientSet struct to update the status of the agent -> proxy-server connection at regular intervals.

Concern: This makes the assumption that the client is connected to at least one proxy server and therefore the status is ready.

I've thought about updating the readiness of the agent only when all proxy-servers are connected but wasn't sure about the implications from the kubernetes side, ~~since this being a readiness check , kubernetes might treat the pod as not ready and not send any traffic.~~

EDIT: Since the agent pods are not behind a Kubernetes service, the readiness of the pod doesn't really matter. 

Please share your reviews,concerns to the approach I've taken.

/cc: @jveski @cheftako @jkh52 @tallclair

### DIFF
--- a/cmd/agent/app/server.go
+++ b/cmd/agent/app/server.go
@@ -17,6 +17,7 @@ limitations under the License.
 package app
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
 	"fmt"
@@ -26,6 +27,7 @@ import (
 	"runtime"
 	runpprof "runtime/pprof"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -107,14 +109,38 @@ func (a *Agent) runHealthServer(o *options.GrpcProxyAgentOptions, cs agent.Readi
 	livenessHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		fmt.Fprintf(w, "ok")
 	})
+
+	checks := []agent.HealthChecker{agent.Ping, agent.NewServerConnected(cs)}
 	readinessHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if cs.Ready() {
-			w.WriteHeader(http.StatusOK)
-			fmt.Fprintf(w, "ok")
-		} else {
-			w.WriteHeader(http.StatusServiceUnavailable)
-			fmt.Fprintf(w, "not ready")
+		var failedChecks []string
+		var individualCheckOutput bytes.Buffer
+		for _, check := range checks {
+			if err := check.Check(r); err != nil {
+				fmt.Fprintf(&individualCheckOutput, "[-]%s failed: %v\n", check.Name(), err)
+				failedChecks = append(failedChecks, check.Name())
+			} else {
+				fmt.Fprintf(&individualCheckOutput, "[+]%s ok\n", check.Name())
+			}
 		}
+
+		// Always be verbose if the check has failed
+		if len(failedChecks) > 0 {
+			klog.V(0).Infoln("%s check failed: \n%v", strings.Join(failedChecks, ","), individualCheckOutput.String())
+			w.WriteHeader(http.StatusServiceUnavailable)
+			fmt.Fprintf(w, individualCheckOutput.String())
+			return
+		}
+
+		if _, found := r.URL.Query()["verbose"]; !found {
+			w.WriteHeader(http.StatusOK)
+			fmt.Fprint(w, "ok")
+			return
+		}
+
+		fmt.Fprintf(&individualCheckOutput, "check passed\n")
+
+		w.WriteHeader(http.StatusOK)
+		fmt.Fprint(w, individualCheckOutput.String())
 	})
 
 	muxHandler := http.NewServeMux()
--- a/cmd/agent/app/server.go
+++ b/cmd/agent/app/server.go
@@ -36,6 +36,7 @@ import (
 	"k8s.io/klog/v2"
 
 	"sigs.k8s.io/apiserver-network-proxy/cmd/agent/app/options"
+	"sigs.k8s.io/apiserver-network-proxy/pkg/agent"
 	"sigs.k8s.io/apiserver-network-proxy/pkg/util"
 )
 
@@ -63,11 +64,13 @@ func (a *Agent) run(o *options.GrpcProxyAgentOptions) error {
 	}
 
 	stopCh := make(chan struct{})
-	if err := a.runProxyConnection(o, stopCh); err != nil {
+
+	cs, err := a.runProxyConnection(o, stopCh)
+	if err != nil {
 		return fmt.Errorf("failed to run proxy connection with %v", err)
 	}
 
-	if err := a.runHealthServer(o); err != nil {
+	if err := a.runHealthServer(o, cs); err != nil {
 		return fmt.Errorf("failed to run health server with %v", err)
 	}
 
@@ -80,11 +83,11 @@ func (a *Agent) run(o *options.GrpcProxyAgentOptions) error {
 	return nil
 }
 
-func (a *Agent) runProxyConnection(o *options.GrpcProxyAgentOptions, stopCh <-chan struct{}) error {
+func (a *Agent) runProxyConnection(o *options.GrpcProxyAgentOptions, stopCh <-chan struct{}) (agent.ReadinessManager, error) {
 	var tlsConfig *tls.Config
 	var err error
 	if tlsConfig, err = util.GetClientTLSConfig(o.CaCert, o.AgentCert, o.AgentKey, o.ProxyServerHost, o.AlpnProtos); err != nil {
-		return err
+		return nil, err
 	}
 	dialOptions := []grpc.DialOption{
 		grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig)),
@@ -97,15 +100,21 @@ func (a *Agent) runProxyConnection(o *options.GrpcProxyAgentOptions, stopCh <-ch
 	cs := cc.NewAgentClientSet(stopCh)
 	cs.Serve()
 
-	return nil
+	return cs, nil
 }
 
-func (a *Agent) runHealthServer(o *options.GrpcProxyAgentOptions) error {
+func (a *Agent) runHealthServer(o *options.GrpcProxyAgentOptions, cs agent.ReadinessManager) error {
 	livenessHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		fmt.Fprintf(w, "ok")
 	})
 	readinessHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		fmt.Fprintf(w, "ok")
+		if cs.Ready() {
+			w.WriteHeader(http.StatusOK)
+			fmt.Fprintf(w, "ok")
+		} else {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			fmt.Fprintf(w, "not ready")
+		}
 	})
 
 	muxHandler := http.NewServeMux()
--- a/pkg/agent/readiness.go
+++ b/pkg/agent/readiness.go
@@ -16,6 +16,11 @@ limitations under the license.
 
 package agent
 
+import (
+	"fmt"
+	"net/http"
+)
+
 // ReadinessManager supports checking if the agent is ready.
 type ReadinessManager interface {
 	// Ready returns true the proxy server is ready.
@@ -27,4 +32,47 @@ var _ ReadinessManager = &ClientSet{}
 func (cs *ClientSet) Ready() bool {
 	// Returns true if the agent is connected to at least one server.
 	return cs.HealthyClientsCount() > 0
+}
+
+// HealthChecker represents an entity capable of performing health checks.
+type HealthChecker interface {
+	Name() string
+	Check(req *http.Request) error
+}
+
+// ping implements the simplest possible healthz checker.
+type ping struct{}
+
+var Ping HealthChecker = &ping{}
+
+func (p *ping) Name() string {
+	return "ping"
+}
+
+func (p *ping) Check(_ *http.Request) error {
+	return nil
+}
+
+type serverConnected struct {
+	rm ReadinessManager
+}
+
+var ServerConnected HealthChecker = &serverConnected{}
+
+func NewServerConnected(cs ReadinessManager) HealthChecker {
+	return &serverConnected{
+		rm: cs,
+	}
+}
+
+func (s *serverConnected) Name() string {
+	return "server-connected"
+}
+
+func (s *serverConnected) Check(_ *http.Request) error {
+	if s.rm.Ready() {
+		return nil
+	}
+
+	return fmt.Errorf("no servers connected")
 }
--- a/pkg/agent/readiness.go
+++ b/pkg/agent/readiness.go
@@ -1,0 +1,30 @@
+/*
+copyright 2023 the kubernetes authors.
+
+licensed under the apache license, version 2.0 (the "license");
+you may not use this file except in compliance with the license.
+you may obtain a copy of the license at
+
+    http://www.apache.org/licenses/license-2.0
+
+unless required by applicable law or agreed to in writing, software
+distributed under the license is distributed on an "as is" basis,
+without warranties or conditions of any kind, either express or implied.
+see the license for the specific language governing permissions and
+limitations under the license.
+*/
+
+package agent
+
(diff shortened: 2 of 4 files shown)


cmd/agent/app/server.go
```diff
@@ -105,7 +110,13 @@ func (a *Agent) runHealthServer(o *options.GrpcProxyAgentOptions) error {
 		fmt.Fprintf(w, "ok")
 	})
 	readinessHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		fmt.Fprintf(w, "ok")
+		if a.connected {
+			w.WriteHeader(http.StatusOK)
+			_, _ = w.Write([]byte("OK"))
```
jkh52, line 115: I would keep consistent with like 110 `fmt.Fprintf(w, "ok")`

cmd/agent/app/server.go
```diff
@@ -54,6 +54,11 @@ func NewAgentCommand(a *Agent, o *options.GrpcProxyAgentOptions) *cobra.Command
 }
 
 type Agent struct {
+	connected bool
+}
+
+func (a *Agent) onConnect(status bool) {
```
jkh52, line 60: I like the interface that proxy-server defines in `pkg/server/readiness_manager.go`, it seems repeatable for proxy-agent. What do you think?

In the agent case, the ClientSet could expose `cs.Readiness.Ready()` instead of this callback.

tests/readiness_test.go
```diff
@@ -38,8 +38,42 @@ func TestReadiness(t *testing.T) {
 	clientset := runAgent(proxy.agent, stopCh)
 	waitForConnectedServerCount(t, 1, clientset)
 
+	// check the agent connected status
```
jkh52, line 41: I would rename this test `TestServerReadiness` and keep agent readiness tests separate.

jkh52: Or, maybe we should move this file to `tests/server_readiness.go` and introduce `tests/agent_readiness.go`

pkg/agent/clientset.go
```diff
@@ -187,10 +205,16 @@ func (cs *ClientSet) sync() {
 			} else {
 				klog.ErrorS(err, "cannot connect once")
 				duration = backoff.Step()
+				if cs.onConnect != nil {
+					cs.SetConnectionStatus(false)
+				}
```
jkh52, line 210: In the multiple-control plane nodes case, won't this set to False incorrectly in some cases?

What about: setting status (somewhere in the for loop, unconditionally), based on `cs.ClientsCount() > 0`?

tests/readiness_test.go
ipochi: I've renamed the tests to depict what they mean. Not sure if have two different files is a good idea.

for example: the tests would be test grpc server and agent readiness, test http connect server and agent readiness and test agent readiness without the server.

WDYT ?

jkh52: @ipochi did you upload your latest changes? I don't yet see the corresponding changes.

ipochi: @jkh52 Sorry about that. Pushed the changes.

I would like to keep the logic "Agent reports Ready, when it is connected to atleast one konnectivity-server".

If we have the logic of "Ready only when all servers are connected" then lets say one of the konnectivity-server is in a crashloop state, then even though tunnel is working fine in a degraded state, kubernetes marks the agent-pods as not ready.

Which may not be just the visual look as there might be other sub-systems that might kick in based on the non-readiness.

ipochi: @jkh52 Sorry about that. Pushed the changes.

I would like to keep the logic "Agent reports Ready, when it is connected to atleast one konnectivity-server".

If we have the logic of "Ready only when all servers are connected" then lets say one of the konnectivity-server is in a crashloop state, then even though tunnel is working fine in a degraded state, kubernetes marks the agent-pods as not ready.

Which may not be just the visual look as there might be other sub-systems that might kick in based on the non-readiness.

pkg/agent/clientset.go
ipochi: https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/485#discussion_r1213143798

cmd/agent/app/server.go
```diff
@@ -97,15 +99,21 @@ func (a *Agent) runProxyConnection(o *options.GrpcProxyAgentOptions, stopCh <-ch
 	cs := cc.NewAgentClientSet(stopCh)
 	cs.Serve()
 
-	return nil
+	return cs, nil
 }
 
-func (a *Agent) runHealthServer(o *options.GrpcProxyAgentOptions) error {
+func (a *Agent) runHealthServer(o *options.GrpcProxyAgentOptions, cs *agent.ClientSet) error {
```
jkh52, line 105: I would use the ReadinessManager interface here.

pkg/agent/readiness.go
```diff
@@ -0,0 +1,33 @@
+/*
+copyright 2020 the kubernetes authors.
```
jkh52, line 2: 2023

pkg/agent/readiness.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+var _ ReadinessManager = &ClientSet{}
+
+func (cs *ClientSet) Ready() bool {
+	if cs.HealthyClientsCount() > 0 {
+		return true
+	}
+
+	return false
```
jkh52, line 32: could simplify: "return cs.HealthyClientsCount() > 0"

tests/readiness_test.go
```diff
@@ -38,8 +36,42 @@ func TestReadiness(t *testing.T) {
...
+	if !clientset.Ready() {
+		t.Fatalf("expected connection status 'true', got: :%t", clientset.Ready())
```
jkh52, line 41: nit: extraneous colon? same elsewhere

tests/readiness_test.go
jkh52: +1 to simply changing the default meaning of readiness to "at least 1".

To repeat what @cheftako mentioned in this week's biweekly meeting: a user who may want the more complex definition "connected == ServerCount" can ask for it using a URL parameter (to be implemented later if desired). There is precedence elsewhere (kube-apiserver).

tests/readiness_test.go
jkh52, line 41: Nit:

I would simplify verifidation to be similar to the server check above `t.Fatalf("expected not ready")`

The only possible failure value is "false", and the second call to clientsets.Ready() actually performs work (takes a lock, is subject to non-determinism bugs, etc). So if you prefer to keep it, please capture the value instead of calling twice.

tests/readiness_test.go
```diff
@@ -38,8 +36,43 @@ func TestReadiness(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cleanup()
+
+	clientset := runAgent(proxy.agent, stopCh)
+	waitForConnectedServerCount(t, 1, clientset)
+
+	// check the agent connected status
+	if !clientset.Ready() {
```
jkh52, line 64: please make this consistent with the pattern above.

tests/readiness_test.go
```diff
@@ -38,8 +36,43 @@ func TestReadiness(t *testing.T) {
...
+	}
+}
+
+func TestAgentReadinessWithoutServer(t *testing.T) {
+	stopCh := make(chan struct{})
+	defer close(stopCh)
+
+	clientset := runAgent("localhost:8080", stopCh)
+	// check the agent connected status
+	if clientset.Ready() {
```
jkh52, line 75: same

jkh52:
I see that @mihivagyok has verified this - awesome!

/lgtm
/approve

/hold
to give a chance to fix the remaining nits

ipochi:
@jkh52  All comments resolved. Review commit squashed as well.  Once the tests are green we can merge this.

jkh52:
/approve

Thanks for this improvement @ipochi!

I was discussing this PR with @cheftako and he wanted to have an input, so I will leave the LGTM to him.

/assign @cheftako

cmd/agent/app/server.go
```diff
@@ -97,15 +100,21 @@ func (a *Agent) runProxyConnection(o *options.GrpcProxyAgentOptions, stopCh <-ch
...
-func (a *Agent) runHealthServer(o *options.GrpcProxyAgentOptions) error {
+func (a *Agent) runHealthServer(o *options.GrpcProxyAgentOptions, cs agent.ReadinessManager) error {
 	livenessHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		fmt.Fprintf(w, "ok")
 	})
 	readinessHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		fmt.Fprintf(w, "ok")
+		if cs.Ready() {
+			w.WriteHeader(http.StatusOK)
+			fmt.Fprintf(w, "ok")
```
cheftako, line 113: I generally like the change. However I'm a little conerned about this as its the actual external facing API surface. I would suggest we break this down as an overall ready/not ready but also we have named checks as per [https://github.com/kubernetes/kubernetes/blob/589d6f388680b29f7f7961bda4f4262733b54210/test/e2e/apimachinery/health_handlers.go#L76](https://www.google.com/url?q=https://github.com/kubernetes/kubernetes/blob/589d6f388680b29f7f7961bda4f4262733b54210/test/e2e/apimachinery/health_handlers.go%23L76&sa=D&source=docs&ust=1686640049965118&usg=AOvVaw1biHPmKn7GLxXClPAsjSlj). Ready here is really checking whether there is an attached KonnServer. That should e a named check. I think for you purposes that is what you want to look at. While it would feed into the overall readiness, I think readiness should include more that that,.

jkh52:
With the extra complexity added in latest snapshot, I would prefer to have some unit test coverage. (Cross reference: https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/497)

Is there any reason we need both readiness.go and healthz.go? They both seem specific to readiness.

ipochi:
> With the extra complexity added in latest snapshot, I would prefer to have some unit test coverage. (Cross reference: #497)
> 
> Is there any reason we need both readiness.go and healthz.go? They both seem specific to readiness.

Thanks. Haven't been able to get to this but will address your feedback next week.

pkg/agent/healthz.go
```diff
@@ -0,0 +1,65 @@
...
+func (s *serverConnected) Name() string {
+	return "server-connected"
+}
+
+func (s *serverConnected) Check(_ *http.Request) error {
+	if s.rm.Ready() {
+		return nil
+	}
+
+	return fmt.Errorf("no servers connected")
```
cheftako, line 64: Healthz/liveness are used to determine if the kubelet is going to restart the pod. I don;t think we want to defaulting the agent to being restarted by kubelet because it hasn't been able to connect to konnectivity server. If we want to add that as a configurable behavior (turned on by flag) I could see the value but I don't think it should be the default behavior.

jkh52:
FYI I'll be on vacation and minimally online for the first half of July. But I'd really like to see this feature included in the next 0.1.x release tag, hopefully both "soon".

I see a specific question about whether `ping` trivial check is intended to be added to agent liveness. But there is an overall question of whether (and if so when) we should similarly support sub-health for agent liveness (or konnectivity-server). Even if we agree, I suggest keeping this PR scoped only to agent readiness.

jkh52:
Friendly ping.

pkg/agent/healthz.go
ipochi: We are not changing the behavior of the healthz/liveness probe. This only changes the behavior of readiness probe.

This too is not behind a flag, so if the flag is enabled, we have two readiness checks (ping, atleast one server connected) else only a ping check (which is quite similar to what we had before)

/cc @jkh52

jkh52: > We are not changing the behavior of the healthz/liveness probe. This only changes the behavior of readiness probe.

Looking at the change to server.go I agree. But can you make small adjustments to make it more obvious to future readers? For example, only introduce readiness.go (not healthz.go), and avoid code naming like "HealthZ" (which some readers may think implies liveness health).

cmd/agent/app/options/options.go
```diff
@@ -117,6 +119,7 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.AgentIdentifiers, "agent-identifiers", o.AgentIdentifiers, "Identifiers of the agent that will be used by the server when choosing agent. N.B. the list of identifiers must be in URL encoded format. e.g.,host=localhost&host=node1.mydomain.com&cidr=127.0.0.1/16&ipv4=1.2.3.4&ipv4=5.6.7.8&ipv6=:::::&default-route=true")
 	flags.BoolVar(&o.WarnOnChannelLimit, "warn-on-channel-limit", o.WarnOnChannelLimit, "Turns on a warning if the system is going to push to a full channel. The check involves an unsafe read.")
 	flags.BoolVar(&o.SyncForever, "sync-forever", o.SyncForever, "If true, the agent continues syncing, in order to support server count changes.")
+	flags.BoolVar(&o.EnableAgentServerReadiness, "agent-server-readiness", o.EnableAgentServerReadiness, "If true, the agent readiness probe checks if its connected to atleast one server.")
```
jkh52, line 122: > flags.BoolVar(&o.EnableAgentServerReadiness, "agent-server-readiness",

Did you add this flag based on a misunderstanding with @cheftako comments (about changing liveness behavior)? I don't remember saying that we need a flag to enable the new readiness behavior (I am ok with not flag guarding the readiness change).

If we keep the flag: since this is the agent binary, "agent" in name seems somewhat redundant. Maybe "--enable-connected-readiness" or similar?

cmd/agent/app/server.go
```diff
@@ -97,15 +102,50 @@ func (a *Agent) runProxyConnection(o *options.GrpcProxyAgentOptions, stopCh <-ch
...
+
+	checks := []agent.HealthChecker{agent.PingHealthz}
+
+	if o.EnableAgentServerReadiness {
+		checks = append(checks, agent.NewServerConnectedHealthz(cs))
+	}
+
 	readinessHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		fmt.Fprintf(w, "ok")
+		var failedChecks []string
```
jkh52, line 120: This section is very close to fully generic. It would be nice to organize it somewhere different than cmd/agent to keep it minimal. It would also be good to add some unit test coverage (for example: exercise the verbose codepath).

I'm ok with deferring these suggestions, to unblock making the feature available.

ipochi: IMO, I'd like to get this merged as is.

 The addition of unit tests, while useful would require more thoughts and time and the result would be a very trivial check and not worth the effort and time.

cmd/agent/app/options/options.go
ipochi: Yes, I think there was a confusion on that. I too don't think a flag is necessary.
I'm happy to remove the commit related to addition of the flag.

ipochi: removed the commit.

ipochi:
@jkh52 @cheftako 

friendly ping. Shall we merge this ?

jkh52:
> @jkh52 @cheftako
> 
> friendly ping. Shall we merge this ?

Did you see the feedback: *But can you make small adjustments to make it more obvious to future readers? For example, only introduce readiness.go (not healthz.go), and avoid code naming like "HealthZ" (which some readers may think implies liveness health).*

I'm actually OK with merging as-is, and we can always address the above separately. Adding a hold just so that @cheftako can also look.

/hold
/lgtm
/approve

pkg/agent/healthz.go
ipochi: @jkh52 Sorry I missed this.

So two points if I understand correctly:
1. keep one file `readiness.go` , merge `healthz.go` into readiness.go
2. PingHealthz -> Ping / PingCheck/ PingHealth ? and similar places replacing Healtz.

I'm okay with just  Ping / ServerConnected or PingCheck  / ServerConnectedCheck for both the readiness checks.

jkh52: Sounds reasonable.

ipochi:
> /lgtm
> 
> Feel free to remove the hold in a day or two, if no other reviewers have feedback.

How does one remove hold ?

jkh52:
> > /lgtm
> > Feel free to remove the hold in a day or two, if no other reviewers have feedback.
> 
> @jkh52 How does one remove hold ?

I believe it is "/remove-hold"

PR #489, opened 2023-05-05
Fix admin bind address description

timoreimann:
The default address is localhost ([source](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/f6480fca6c74f62d0cbf3bf5c34657dc7f0c2afb/cmd/server/app/options/options.go#L341)).


jkh52:
Good catch - thank you.

/lgtm
/approve

PR #493, opened 2023-05-31
[WIP] Made connection management extensible.

cheftako:
TODO


cmd/agent/app/options/options.go
```diff
@@ -166,8 +202,23 @@ func (o *GrpcProxyAgentOptions) Validate() error {
...
+			return fmt.Errorf("proxy server list cannot be defined with the HA connection manager")
+		}
+	case agent.IPListConnectionManager:
+		if err := validateAddressList(o.ProxyServerList); err != nil {
+			return fmt.Errorf("proxy server address list is invalid: #{err}")
+		}
+		if len(o.ProxyServerHost) > 0 || o.ProxyServerPort != 0 {
+			return fmt.Errorf("cannot define proxyServerList with either proxyServerHost or proxyServerPort")
+		}
+	default:
```
jkh52, line 220: wouldn't defaulting to HA be better (backward compatible)?

jkh52:
overall direction LGTM.

Comments:
- We should track integration test coverage of the new configuration
- In the new configuration, agents stop relying on server count but proxy-server will continue to produce it. Long term, should there be any accompanying proxy-server or agent+server protocol change (for example: an agent can advertise that it doesn't need it)?
  * My own opinion: make no change, other than maybe a warning log if agent sees mis-matched values.

jkh52:
@cheftako do you still plan to do this, and do you have a reference for what user / company wanted this feature? It seems relevant to the design under discussion at https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/635

Issue #358, opened 2022-05-31
Improve dynamic server count detect logic of agent


zqzten:
Currently if we want the agent to dynamic detect the server count change, we have to enable the `syncForever` mode, however, with this mode on, agent just call the `Connect` method of server constantly, and for most cases it just grap the sever count and close the connection (since the server count is not chaging that frequently). This cause a lot of garbage error log (specifically `server.go:767] "stream read failure" err="rpc error: code = Canceled desc = context canceled"` on server side) and a waste of networking and computing resource.

So would it be applicable that we introduce a lightweight method `ServerCount` to server and let agent call that method to sync server count instead of the heavier `Connect`?

ipochi:
@cheftako @tallclair 

I'd be interested to know more about what others in the community feel regarding this and is there a scope/priority for a better solution ?

I'd like to collaborate with relevant folks and formulate an approach.

tallclair:
I agree that the current approach is problematic. Introducing a new lightweight method makes sense to me, but I'd like someone with more experience on this project to weigh in.

zqzten:
> One thing we haven't address yet is how the server gets that value.

Good point. I've noticed that there was another rotting issue #273 tracking on this. We may consider that together with this issue.

jkh52:
> Good point. I've noticed that there was another rotting issue https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/273 tracking on this. We may consider that together with this issue.

We could keep this issue open, to improve "how the agents learn server count". There are some ideas here that weren't considered there. https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/273 can track the remaining server half of the feature.

carreter:
FYI, if an agent receives an increased server count from a server it is already connected to, it doesn't reset the backoff and just uses the last set duration. I'm guessing that this isn't the desired behavior and that the agent should immediately reset the backoff and shift into "fast sync" mode if it is supposed to be connected to more servers than it currently is.

This will only be a problem once #273 is implemented as there is currently no way to update the server count on the server side.

carreter:
Fixed by #643 !

PR #285, opened 2021-11-05
proxy-agent: support new SyncForever mode.

jkh52:
This is a minimal proxy-agent half of
https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/273

It is already useful, in the scenario where proxy-server are restarted.

PR #494, opened 2023-05-31
Quiet more error logs on stream canceled.

jkh52:
This mitigates log error spam noted at https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/358 (when agent uses --sync-forever, it frequently triggers stream canceled in server codepath).

In general, we already log at info level for stream EOF, and stream canceled seems comparable.

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -30,7 +30,9 @@ import (
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/connectivity"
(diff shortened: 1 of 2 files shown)


jkh52:
When I worked on https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/285, I remember prioritizing agent logs having no ERROR level in the "steady state". But I didn't pay the same level of attention to server logs.

zqzten (approved):
/lgtm

PR #497, opened 2023-06-08
Enable coverage measurement

mihivagyok:
- also generate html report

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ mock_gen:
 
 .PHONY: test
 test:
-	go test -race -covermode=atomic -coverprofile=server.out ./... && go tool cover -html=server.out -o=server.html
+	go test -race -covermode=atomic -coverprofile=konnectivity.out ./... && go tool cover -html=konnectivity.out -o=konnectivity.html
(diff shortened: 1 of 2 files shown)


jkh52:
@mihivagyok - I see the cncf-cla label was removed, please restore it so this can be merged.

mihivagyok:
@jkh52 Thank you! It is restored!

Makefile
```diff
@@ -57,8 +57,8 @@ mock_gen:
 
 .PHONY: test
 test:
-	go test -race ./...
-	cd konnectivity-client && go test -race ./...
+	go test -race -covermode=atomic -coverprofile=server.out ./... && go tool cover -html=server.out -o=server.html
```
jkh52, line 60: the name `server` is misleading because we use the top level module with both server and agent. Maybe a name like `test-coverage`, or `konnectivity`?

jkh52:
It would be nice for `make clean` to remove the generated files.

Is the typical pattern to check these in to source? (I'm unfamiliar here.) If not, then I would add them to .gitignore.

mihivagyok:
Thank you! I will submit the changes very soon!

PR #498, opened 2023-06-12
proxy-server: change --delete-existing-uds-file default to true.

jkh52:
This is followup from discusssion at https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/476

### DIFF
--- a/cmd/server/app/options/options.go
+++ b/cmd/server/app/options/options.go
@@ -114,7 +114,7 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.ClusterCaCert, "cluster-ca-cert", o.ClusterCaCert, "If non-empty the CA we use to validate Agent clients.")
 	flags.StringVar(&o.Mode, "mode", o.Mode, "mode can be either 'grpc' or 'http-connect'.")
 	flags.StringVar(&o.UdsName, "uds-name", o.UdsName, "uds-name should be empty for TCP traffic. For UDS set to its name.")
-	flags.BoolVar(&o.DeleteUDSFile, "delete-existing-uds-file", o.DeleteUDSFile, "If true and if file UdsName already exists, delete the file before listen on that UDS file")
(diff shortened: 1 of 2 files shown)


ipochi:
@jkh52 @cheftako 

Does it not go against the default value of bool in golang and also in general ? 

I mean if you want to set a bool flag, you set it explicitly such as (--delete-exisiting-uds-file), if it is not set then the default should be assumed to be false.

Wouldn't it make setting the flag `--delete-existing-uds-file` redundant if the default is true.

guettli:
@ipochi 

> Wouldn't it make setting the flag --delete-existing-uds-file redundant if the default is true.

I had the same idea on my mind. I think there is no reason to not delete the file. An alternative would be to make all the things which systemd does to ensure servers don't get accidentaly created twice. But this makes no sense, since this script gets run by systemd (at least in the places I am aware of).

jkh52:
> @ipochi
> 
> > Wouldn't it make setting the flag --delete-existing-uds-file redundant if the default is true.
> 
> I had the same idea on my mind. I think there is no reason to not delete the file. An alternative would be to make all the things which systemd does to ensure servers don't get accidentaly created twice. But this makes no sense, since this script gets run by systemd (at least in the places I am aware of).

 If we go this route (let's say option B), we should mark the flag deprecated for some time: [example](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/cmd/server/app/options/options.go#L142).

I only slightly prefer option A (this PR currently), and if I'm in the minority I'll switch to B. @cheftako do you have an opinion?

It does seem true that `getUdsListener` cannot possibly succeed if `udsName` already exists.

jkh52:
This was discussed in the OSS biweekly this morning: [Meeting Notes](https://docs.google.com/document/d/1aPgGRl4WewM3txrCYvkepsxLUvGdMG1EzlVfCNeV74M). From discussion: an operator running multiple proxy servers on a single host could possibly want to manage UDS file cleanup in some careful or particular way. There was unanimous consensus to go forward with keeping the flag rather than hardcoding the cleanup.

PR #500, opened 2023-06-21
chore: pkg imported more than once

testwill:


jkh52:
Thanks for the cleanup!

We can probably add a relevant linter, found: https://golangci-lint.run/usage/linters/#importas

/lgtm
/approve

Issue #350, opened 2022-04-11
Need a way to seamlessly rotate the Proxy Server Certs on the Agent side.


cheftako:
The Proxy Server is authenticated to the agents via its certificate. The KAS (K8s API Server) handles this by dynamically setting the appropriate certificate based on the SNI in the incoming handshake request. It would seem good if were to do the same for the Proxy Server. This would allow the Proxy Server to properly advertise its authentication materials (both the old and new certs) during a cert rotation.

PR #501, opened 2023-06-27
Add file watcher for TLS certificate files

afritzler:
Currently the agent and server are not honouring the change of TLS certificates provided to them. This PR introduces a file watcher for both agent and server to restart the corresponding components in case there as been a change in the TLS certificates.

## Changes

- Add fsnotify dependency to go.mod
- Add file watcher for certificate files in `cmd/agent/app/server.go` and `cmd/server/app/server.go`
- Reload TLS config and credentials on certificate file change in `cmd/agent/app/server.go` and `cmd/server/app/server.go`
- Refactor and clean up file watcher code in `cmd/agent/app/server.go` and `cmd/server/app/server.go`

### DIFF
--- a/cmd/agent/app/server.go
+++ b/cmd/agent/app/server.go
@@ -28,6 +28,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/fsnotify/fsnotify"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/spf13/cobra"
 	"google.golang.org/grpc"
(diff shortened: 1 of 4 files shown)


jkh52:
Initial thoughts:

Some admins may want a security posture that disallows this, it could be better to opt-in to the new feature via flag (or at a minimum support an opt-out).

We need to make sure this doesn't make https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/350 harder to implement (seems like it shouldn't).

It would be nice to see some unit test coverage.

afritzler:
@jkh52 thanks a lot for the feedback. Let me address the point of disabling the rotation + adding some test coverage.

jkh52:
@afritzler - see the effort at https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/577 (it would give you attribution, but 'EasyCLA check' does not pass. Can you fix that part?)

afritzler:
> @afritzler - see the effort at #577 (it would give you attribution, but 'EasyCLA check' does not pass. Can you fix that part?)

Thanks a lot for the update. I fixed the CLA issue. I guess we can close this PR in favor of #577.

PR #505, opened 2023-07-25
bump k8s dependencies to 1.27

mfojtik:
This bumps k8s deps to 1.27.4 to mitigate some security issues with some of the third-party go modules.


mfojtik:
this might require https://github.com/kubernetes/test-infra/pull/30203

jkh52:
It looks like your intention is to bump vendored modules, but the vendored library is [konnectivity-client](https://github.com/kubernetes-sigs/apiserver-network-proxy/tree/master/konnectivity-client). (The top level go.mod only affect the `konnectivitity-server` and `konnectivity-agent` binaries which are not vendored).

mfojtik:
@jkh52 I only intend to bump k8s modules because they use the old version of k8s, which is affected by various CVE's/etc. For example goproxy module was dropped in Kubernetes 1.27 entirely.

jkh52:
/retest

/lgtm

Agreed that we may need that testinfra PR to merge first.

PR #508, opened 2023-08-02
update genproto

liggitt:
This updates the genproto library to a version matching kubernetes 1.28, which carves out the rpc subpackage (see https://github.com/googleapis/go-genproto/issues/1015 for details)


konnectivity-client/go.mod
```diff
@@ -1,28 +1,28 @@
...
 
-go 1.17
+go 1.20
 
 // Prefer to keep requirements compatible with the oldest supported
 // k/k minor version, to prevent client backport issues.
 require (
 	github.com/prometheus/client_golang v1.11.1
 	go.uber.org/goleak v1.2.0
-	golang.org/x/net v0.7.0 // indirect
```
cheftako, line 10: I'm fine with the top level go.mod changes. 
Our [docs](https://github.com/kubernetes/kubernetes/blob/release-1.27/go.mod#L81) currently indicate that the main line is meant for K/K 1.27+. However 1.27 [go.mod](https://github.com/kubernetes/kubernetes/blob/release-1.27/go.mod#L81) show this would force an upgrade of libraries on the KAS. 

Sounds like we should probably create a 0.2 (?0.28) branch to correspond the the 1.28 K/K release.

Might be worth breaking this into two PRs, so we can get the ANP binaries fixed.

jkh52:
https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/510 tracks separate branching proposal.

konnectivity-client/go.mod
jkh52: https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/510 is resolved, and per last biweekly OSS meeting we can proceed to create a "new scheme" branch release-0.28 and cut a 0.28.x tag (in this repo).

But, k/k at release-1.28 still has older versions than what I see here, is it expected?

https://github.com/kubernetes/kubernetes/blob/release-1.28/go.mod#L91-L92

liggitt: https://github.com/kubernetes/kubernetes/pull/118269/files has the minimum genproto versions we'd want to update to here... I can pick those instead of latest (which is what pulled in the newer grpc)

liggitt: pushed an update to match versions used in k8s 1.28

does the resolution of https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/510 mean we would:
1. create release-0.27 from master before this PR (to give a target for updates destined for k8s 1.27)
2. create release-0.28 from master before this PR (to give a target for updates destined for k8s 1.28)
3. merge this PR master and cherry-pick it to release-0.28

jkh52: Essentially number 1, except we'll start the new mapping scheme in 1.28.

I sent https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/517 to propose new branch names in the short term.

jkh52:
/lgtm
/approve

I created release-0.1 branch at HEAD (for 1.27 and previous), so we can begin using master branch for 1.28.

Feel free to unhold when you're ready.

/hold

PR #511, opened 2023-08-10
fix: metric frontend_write_duration_seconds has wrong value

Abirdcfly:
And maybe this issue is helpful: golang/go#60048


jkh52:
Thank you!

/lgtm
/approve

Issue #513, opened 2023-08-14
proxy-server panic when setting `--proxy-strategies=destHost`.


xuezhaojun:
The panic error msg:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x15e4a2d]
goroutine 56 [running]:
sigs.k8s.io/apiserver-network-proxy/pkg/server.(*ProxyServer).readBackendToChannel(0xc000510f20?, {0x0, 0x0}, 0x17b5a80?, 0x290f070?)
/go/src/github.com/stolostron/cluster-proxy-addon/_output/apiserver-network-proxy/pkg/server/server.go:853 +0x6d
sigs.k8s.io/apiserver-network-proxy/pkg/server.(*ProxyServer).Connect.func3({0x1c72db8, 0xc00068f2c0})
/go/src/github.com/stolostron/cluster-proxy-addon/_output/apiserver-network-proxy/pkg/server/server.go:846 +0x2d
runtime/pprof.Do({0x1c72d48?, 0xc000046050?}, {{0xc0006201c0?, 0x0?, 0x0?}}, 0xc00068f260)
/usr/local/go/src/runtime/pprof/runtime.go:40 +0xa3
created by sigs.k8s.io/apiserver-network-proxy/pkg/server.(*ProxyServer).Connect
/go/src/github.com/stolostron/cluster-proxy-addon/_output/apiserver-network-proxy/pkg/server/server.go:846 +0xc85
```

The root cause is this function, when `--proxy-strategies=destHost` the return value `backend` is nil:  https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/9c1359e68b932458a07464bc607f61a06e879ff3/pkg/server/server.go#L258

```go
func (s *ProxyServer) addBackend(agentID string, conn agent.AgentService_ConnectServer) (backend Backend) {
	for i := 0; i < len(s.BackendManagers); i++ {
		switch s.BackendManagers[i].(type) {
		case *DestHostBackendManager:
			agentIdentifiers, err := getAgentIdentifiers(conn)
			if err != nil {
				klog.ErrorS(err, "fail to get the agent identifiers", "agentID", agentID)
				break
			}
			for _, ipv4 := range agentIdentifiers.IPv4 {
				klog.V(5).InfoS("Add the agent to DestHostBackendManager", "agent address", ipv4)
[...]

PR #514, opened 2023-08-21
proxy-server: Wrap Backend more completely.

jkh52:
Fix nil deref at https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/513

We suspect that https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/463 caused this, by adding more callsites to Backend interface, which interacted with an inconsistency in [ProxyServer.addBackend](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/9c1359e68b932458a07464bc607f61a06e879ff3/pkg/server/server.go#L258-L295) (sometimes returning nil).

This change also (by design) removes the interchangeability between agent.AgentService_ConnectServer and Backend, because the backend storages rely on pointer comparison and it would be a mistake to mix them.

### DIFF
--- a/pkg/server/backend_manager.go
+++ b/pkg/server/backend_manager.go
@@ -25,6 +25,7 @@ import (
 	"sync"
 	"time"
 
+	"google.golang.org/grpc/metadata"
 	"k8s.io/klog/v2"
 
 	commonmetrics "sigs.k8s.io/apiserver-network-proxy/konnectivity-client/pkg/common/metrics"
@@ -70,18 +71,21 @@ func GenProxyStrategiesFromStr(proxyStrategies string) ([]ProxyStrategy, error) 
 	return ps, nil
 }
 
+// Backend abstracts a connected Konnectivity agent.
+//
+// In the only currently supported case (gRPC), it wraps an
+// agent.AgentService_ConnectServer, provides synchronization and
+// emits common stream metrics.
 type Backend interface {
 	Send(p *client.Packet) error
 	Recv() (*client.Packet, error)
 	Context() context.Context
+	GetAgentIdentifiers() (header.Identifiers, error)
 }
 
 var _ Backend = &backend{}
-var _ Backend = agent.AgentService_ConnectServer(nil)
 
 type backend struct {
-	// TODO: this is a multi-writer single-reader pattern, it's tricky to
-	// write it using channel. Let's worry about performance later.
 	sendLock sync.Mutex
 	recvLock sync.Mutex
 	conn     agent.AgentService_ConnectServer
@@ -121,7 +125,24 @@ func (b *backend) Context() context.Context {
 	return b.conn.Context()
 }
 
-func newBackend(conn agent.AgentService_ConnectServer) *backend {
+func (b *backend) GetAgentIdentifiers() (header.Identifiers, error) {
+	var agentIdentifiers header.Identifiers
+	md, ok := metadata.FromIncomingContext(b.Context())
+	if !ok {
+		return agentIdentifiers, fmt.Errorf("failed to get metadata from context")
+	}
+	agentIDs := md.Get(header.AgentIdentifiers)
+	if len(agentIDs) > 1 {
+		return agentIdentifiers, fmt.Errorf("expected at most one set of agent IDs in the context, got %v", agentIDs)
+	}
+	if len(agentIDs) == 0 {
+		return agentIdentifiers, nil
+	}
+
+	return header.GenAgentIdentifiers(agentIDs[0])
+}
+
+func NewBackend(conn agent.AgentService_ConnectServer) Backend {
 	return &backend{conn: conn}
 }
 
@@ -129,9 +150,9 @@ func newBackend(conn agent.AgentService_ConnectServer) *backend {
 // connections, i.e., get, add and remove
 type BackendStorage interface {
 	// AddBackend adds a backend.
-	AddBackend(identifier string, idType header.IdentifierType, conn agent.AgentService_ConnectServer) Backend
+	AddBackend(identifier string, idType header.IdentifierType, backend Backend)
 	// RemoveBackend removes a backend.
-	RemoveBackend(identifier string, idType header.IdentifierType, conn agent.AgentService_ConnectServer)
+	RemoveBackend(identifier string, idType header.IdentifierType, backend Backend)
 	// NumBackends returns the number of backends.
 	NumBackends() int
 }
@@ -168,7 +189,7 @@ type DefaultBackendStorage struct {
 	// For a given agent, ProxyServer prefers backends[agentID][0] to send
 	// traffic, because backends[agentID][1:] are more likely to be closed
 	// by the agent to deduplicate connections to the same server.
-	backends map[string][]*backend
+	backends map[string][]Backend
 	// agentID is tracked in this slice to enable randomly picking an
 	// agentID in the Backend() method. There is no reliable way to
 	// randomly pick a key from a map (in this case, the backends) in
@@ -198,7 +219,7 @@ func NewDefaultBackendStorage(idTypes []header.IdentifierType) *DefaultBackendSt
 	// no agent ever successfully connects.
 	metrics.Metrics.SetBackendCount(0)
 	return &DefaultBackendStorage{
-		backends: make(map[string][]*backend),
+		backends: make(map[string][]Backend),
 		random:   rand.New(rand.NewSource(time.Now().UnixNano())),
 		idTypes:  idTypes,
 	} /* #nosec G404 */
@@ -214,42 +235,40 @@ func containIDType(idTypes []header.IdentifierType, idType header.IdentifierType
 }
 
 // AddBackend adds a backend.
-func (s *DefaultBackendStorage) AddBackend(identifier string, idType header.IdentifierType, conn agent.AgentService_ConnectServer) Backend {
+func (s *DefaultBackendStorage) AddBackend(identifier string, idType header.IdentifierType, backend Backend) {
 	if !containIDType(s.idTypes, idType) {
 		klog.V(4).InfoS("fail to add backend", "backend", identifier, "error", &ErrWrongIDType{idType, s.idTypes})
-		return nil
+		return
 	}
-	klog.V(5).InfoS("Register backend for agent", "connection", conn, "agentID", identifier)
+	klog.V(5).InfoS("Register backend for agent", "agentID", identifier)
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	_, ok := s.backends[identifier]
-	addedBackend := newBackend(conn)
 	if ok {
-		for _, v := range s.backends[identifier] {
-			if v.conn == conn {
-				klog.V(1).InfoS("This should not happen. Adding existing backend for agent", "connection", conn, "agentID", identifier)
-				return v
+		for _, b := range s.backends[identifier] {
+			if b == backend {
+				klog.V(1).InfoS("This should not happen. Adding existing backend for agent", "agentID", identifier)
+				return
 			}
 		}
-		s.backends[identifier] = append(s.backends[identifier], addedBackend)
-		return addedBackend
+		s.backends[identifier] = append(s.backends[identifier], backend)
+		return
 	}
-	s.backends[identifier] = []*backend{addedBackend}
+	s.backends[identifier] = []Backend{backend}
 	metrics.Metrics.SetBackendCount(len(s.backends))
 	s.agentIDs = append(s.agentIDs, identifier)
 	if idType == header.DefaultRoute {
 		s.defaultRouteAgentIDs = append(s.defaultRouteAgentIDs, identifier)
 	}
-	return addedBackend
 }
 
 // RemoveBackend removes a backend.
-func (s *DefaultBackendStorage) RemoveBackend(identifier string, idType header.IdentifierType, conn agent.AgentService_ConnectServer) {
+func (s *DefaultBackendStorage) RemoveBackend(identifier string, idType header.IdentifierType, backend Backend) {
 	if !containIDType(s.idTypes, idType) {
 		klog.ErrorS(&ErrWrongIDType{idType, s.idTypes}, "fail to remove backend")
 		return
 	}
-	klog.V(5).InfoS("Remove connection for agent", "connection", conn, "identifier", identifier)
+	klog.V(5).InfoS("Remove connection for agent", "agentID", identifier)
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	backends, ok := s.backends[identifier]
@@ -258,11 +277,11 @@ func (s *DefaultBackendStorage) RemoveBackend(identifier string, idType header.I
 		return
 	}
 	var found bool
-	for i, c := range backends {
-		if c.conn == conn {
+	for i, b := range backends {
+		if b == backend {
 			s.backends[identifier] = append(s.backends[identifier][:i], s.backends[identifier][i+1:]...)
 			if i == 0 && len(s.backends[identifier]) != 0 {
-				klog.V(1).InfoS("This should not happen. Removed connection that is not the first connection", "connection", conn, "remainingConnections", s.backends[identifier])
+				klog.V(1).InfoS("This should not happen. Removed connection that is not the first connection", "agentID", identifier)
 			}
 			found = true
 		}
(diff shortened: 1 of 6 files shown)


jkh52:
/hold

While chasing down some uncovered unit test race conditions.

jkh52:
/assign @cheftako 

PTAL

pkg/server/backend_manager.go
```diff
@@ -121,17 +120,39 @@ func (b *backend) Context() context.Context {
 	return b.conn.Context()
 }
 
-func newBackend(conn agent.AgentService_ConnectServer) *backend {
+func (b *backend) GetAgentIdentifiers() (header.Identifiers, error) {
+	var agentIdentifiers header.Identifiers
+	md, ok := metadata.FromIncomingContext(b.Context())
+	if !ok {
+		return agentIdentifiers, fmt.Errorf("failed to get context")
```
tallclair, line 127: nit: "failed to get metadata" might be more accurate?
```suggestion
		return agentIdentifiers, fmt.Errorf("failed to get metadata from context")
```

pkg/server/backend_manager.go
```diff
@@ -121,17 +120,39 @@ func (b *backend) Context() context.Context {
...
+	}
+	agentIDs := md.Get(header.AgentIdentifiers)
+	if len(agentIDs) > 1 {
+		return agentIdentifiers, fmt.Errorf("expected at most one agent IP in the context, got %v", agentIDs)
```
tallclair, line 131:
```suggestion
		return agentIdentifiers, fmt.Errorf("expected at most one agent ID in the context, got %v", agentIDs)
```

Actually... this is expecting 1 list of agent IDs, so maybe this is more accurate:

```suggestion
		return agentIdentifiers, fmt.Errorf("expected at most one set of agent IDs in the context, got %v", agentIDs)
```

Alternatively, we could just allow this:

```suggestion
		agentIDs = []string{strings.Join(agentIDs, ",")}
```

pkg/server/backend_manager.go
```diff
@@ -121,17 +120,39 @@ func (b *backend) Context() context.Context {
...
+	}
+	if len(agentIDs) == 0 {
```
tallclair, line 133: nit
```suggestion
	else if len(agentIDs) == 0 {
```

pkg/server/backend_manager.go
```diff
@@ -121,17 +120,39 @@ func (b *backend) Context() context.Context {
...
+		return agentIdentifiers, nil
+	}
+
+	agentIdentifiers, err := header.GenAgentIdentifiers(agentIDs[0])
+	if err != nil {
+		return agentIdentifiers, err
+	}
+	return agentIdentifiers, nil
+
```
tallclair, line 142: nit: simplify this to
```suggestion
	return header.GenAgentIdentifiers(agentIDs[0])
```

pkg/server/backend_manager.go
```diff
@@ -121,17 +120,39 @@ func (b *backend) Context() context.Context {
 	return b.conn.Context()
 }
 
-func newBackend(conn agent.AgentService_ConnectServer) *backend {
+func (b *backend) GetAgentIdentifiers() (header.Identifiers, error) {
```
tallclair, line 123: optional: consider doing this work when `NewBackend(...)` is called, and caching the agent identifiers in the `backend` struct instead.

pkg/server/backend_manager.go
```diff
@@ -214,42 +235,40 @@ func containIDType(idTypes []header.IdentifierType, idType header.IdentifierType
...
 	defer s.mu.Unlock()
 	_, ok := s.backends[identifier]
-	addedBackend := newBackend(conn)
 	if ok {
 		for _, v := range s.backends[identifier] {
-			if v.conn == conn {
-				klog.V(1).InfoS("This should not happen. Adding existing backend for agent", "connection", conn, "agentID", identifier)
-				return v
+			if v == backend {
+				klog.V(1).InfoS("This should not happen. Adding existing backend for agent", "connection", backend, "agentID", identifier)
```
tallclair: I doubt `backend` will serialize well, maybe just drop it from this log? Also consider adding ID type.

Also, if this shouldn't happen, should it be error level?

```suggestion
				klog.V(1).ErrorS(nil, "Adding duplicate backend for agent. This should not happen", "agentID", identifier, "idType", idType)
```

pkg/server/backend_manager.go
```diff
@@ -259,10 +278,10 @@ func (s *DefaultBackendStorage) RemoveBackend(identifier string, idType header.I
 	}
 	var found bool
 	for i, c := range backends {
```
tallclair, line 280: nit (since backends are no longer just connection objects)
```suggestion
	for i, b := range backends {
```

pkg/server/backend_manager.go
```diff
@@ -259,10 +278,10 @@ func (s *DefaultBackendStorage) RemoveBackend(identifier string, idType header.I
...
-		if c.conn == conn {
+		if c == backend {
 			s.backends[identifier] = append(s.backends[identifier][:i], s.backends[identifier][i+1:]...)
 			if i == 0 && len(s.backends[identifier]) != 0 {
-				klog.V(1).InfoS("This should not happen. Removed connection that is not the first connection", "connection", conn, "remainingConnections", s.backends[identifier])
+				klog.V(1).InfoS("This should not happen. Removed connection that is not the first connection", "connection", backend, "remainingConnections", s.backends[identifier])
```
tallclair:
```suggestion
				klog.V(1).ErrorS(nil, "This should not happen. Removed connection that is not the first connection", "agentID", identifier, "idType", idType)
```

pkg/server/backend_manager.go
```diff
@@ -286,7 +305,7 @@ func (s *DefaultBackendStorage) RemoveBackend(identifier string, idType header.I
 		}
 	}
 	if !found {
-		klog.V(1).InfoS("Could not find connection matching identifier to remove", "connection", conn, "identifier", identifier)
+		klog.V(1).InfoS("Could not find connection matching identifier to remove", "connection", backend, "identifier", identifier)
```
tallclair:
```suggestion
		klog.V(1).InfoS("Could not find connection matching identifier to remove", "identifier", identifier, "idType", idType)
```

pkg/server/backend_manager.go
```diff
@@ -259,10 +278,10 @@ func (s *DefaultBackendStorage) RemoveBackend(identifier string, idType header.I
 	}
 	var found bool
 	for i, c := range backends {
-		if c.conn == conn {
+		if c == backend {
```
tallclair, line 281: Any risk that multiple `backend` structs will be constructed for a single connection? If so, it might be safer to check `c.conn == backend.conn`

pkg/server/server.go
```diff
@@ -255,79 +255,79 @@ func (s *ProxyServer) getBackend(reqHost string) (Backend, error) {
 	return nil, &ErrNotFound{}
 }
 
-func (s *ProxyServer) addBackend(agentID string, conn agent.AgentService_ConnectServer) (backend Backend) {
+func (s *ProxyServer) addBackend(agentID string, backend Backend) {
 	for i := 0; i < len(s.BackendManagers); i++ {
 		switch s.BackendManagers[i].(type) {
 		case *DestHostBackendManager:
```
tallclair, line 261: bleh... this feels like an abstraction failure. I'd prefer to encapsulate this logic in the `BackendManager`, rather than switching on the type here.

The way I would do it is:
1. make the current `BackendManager.AddBackend(...)` method private
2. make a new `BackendManager.AddBackend(backend)` method that encapsulates the logic here: call the private version for all the relevant IDs. If the backend doesn't have IDs that are relevant for that BackendManager, then it's a no-op


Then, this method `proxyServer.addBackend` just becomes a simple loop through the backend managers

You don't need to address this in this PR if you don't want.

jkh52:
/remove hold

I fixed the test race.

pkg/server/server.go
```diff
@@ -255,79 +255,79 @@ func (s *ProxyServer) getBackend(reqHost string) (Backend, error) {
...
 			klog.V(5).InfoS("Add the agent to DefaultBackendManager", "agentID", agentID)
-			backend = s.BackendManagers[i].AddBackend(agentID, header.UID, conn)
+			s.BackendManagers[i].AddBackend(agentID, header.UID, backend)
 		}
 	}
 	return
 }
 
-func (s *ProxyServer) removeBackend(agentID string, conn agent.AgentService_ConnectServer) {
+func (s *ProxyServer) removeBackend(agentID string, backend Backend) {
```
tallclair, line 297: same feedback as `addBackend` above applies here

pkg/server/backend_manager.go
jkh52: Done.

pkg/server/backend_manager.go
jkh52: Done (except the last suggestion - just to reduce risk).

pkg/server/backend_manager.go
jkh52: Note this was only a code move, let's not change the logic in the same PR as this bugfix.

pkg/server/backend_manager.go
jkh52: I had just removed it (in order to fix the test race) as you sent this feedback. :)

I don't understand the "This should not happen" codepaths in depth. I noticed some unit tests emit them; let's change log to ERROR separately after better understanding it.

pkg/server/backend_manager.go
jkh52: Done

pkg/server/backend_manager.go
jkh52: Done.

Currently that risk does not exist. But (after changing Backend from interface to type in latest snapshot) I converted the comparison also to be consistent with before.

pkg/server/backend_manager.go
jkh52: Done

jkh52:
The reason for switching Backend from interface to type is http://go/gotip/episodes/49, and conforming to the 'make-lint' check.

pkg/server/backend_manager.go
```diff
@@ -70,24 +71,17 @@ func GenProxyStrategiesFromStr(proxyStrategies string) ([]ProxyStrategy, error)
...
 
-type Backend interface {
-	Send(p *client.Packet) error
-	Recv() (*client.Packet, error)
-	Context() context.Context
-}
-
-var _ Backend = &backend{}
-var _ Backend = agent.AgentService_ConnectServer(nil)
+var _ *Backend = &Backend{}
```
elmiko, line 74: i'm curious, what is the point of having this instance of the variable?

pkg/server/backend_manager.go
```diff
@@ -214,42 +225,40 @@ func containIDType(idTypes []header.IdentifierType, idType header.IdentifierType
 }
 
 // AddBackend adds a backend.
-func (s *DefaultBackendStorage) AddBackend(identifier string, idType header.IdentifierType, conn agent.AgentService_ConnectServer) Backend {
+func (s *DefaultBackendStorage) AddBackend(identifier string, idType header.IdentifierType, backend *Backend) {
```
elmiko, line 228: since we are removing the return value and it seems like there is at least one error condition, should we be returning some sort of error from this function?

pkg/server/backend_manager.go
```diff
@@ -214,42 +225,40 @@ func containIDType(idTypes []header.IdentifierType, idType header.IdentifierType
...
 	s.agentIDs = append(s.agentIDs, identifier)
 	if idType == header.DefaultRoute {
 		s.defaultRouteAgentIDs = append(s.defaultRouteAgentIDs, identifier)
 	}
-	return addedBackend
 }
 
 // RemoveBackend removes a backend.
-func (s *DefaultBackendStorage) RemoveBackend(identifier string, idType header.IdentifierType, conn agent.AgentService_ConnectServer) {
+func (s *DefaultBackendStorage) RemoveBackend(identifier string, idType header.IdentifierType, backend *Backend) {
```
elmiko, line 256: similar question as for `AddBackend` about the error return.

elmiko (commented):
i think this patch makes sense to me, but i have a few questions.

pkg/server/backend_manager.go
jkh52: Similar to @tallclair 's comment about ProxyServer and BackendManager separation, I don't disagree but prefer to keep out of this PR scope.

pkg/server/backend_manager.go
jkh52: It survives from https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/78

@cheftako do you have an answer?

pkg/server/backend_manager.go
elmiko: makes sense to me. happy to see the change in a followup PR.

pkg/server/backend_manager.go
elmiko: i'm guessing we don't need this, but i'd be happy to see it cleaned up in another PR  to minimize the scope on this PR.

jkh52:
> @jkh52 this is looking good to me, but i'd love to get another review from someone more familiar with the code.

I'll look for one of @cheftako or @tallclair to LGTM

pkg/server/server.go
jkh52: opened https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/516

pkg/server/backend_manager.go
jkh52: @cheftako explained that previously (when Backend was an interface) these deleted lines provided a compile time check that the given types implement that interface. So given this change, it is appropriate to remove.

elmiko: thanks for the extra context @jkh52 !

pkg/server/backend_manager.go
```diff
@@ -70,24 +71,15 @@ func GenProxyStrategiesFromStr(proxyStrategies string) ([]ProxyStrategy, error)
...
-
-var _ Backend = &backend{}
-var _ Backend = agent.AgentService_ConnectServer(nil)
-
-type backend struct {
-	// TODO: this is a multi-writer single-reader pattern, it's tricky to
-	// write it using channel. Let's worry about performance later.
+// Backend wraps an agent.AgentService_ConnectServer, provides
+// synchronization and emits common stream metrics.
+type Backend struct {
```
cheftako, line 76: Mixed opinions about removing the Backend abstraction. With the interface we were in a place to be able to swap out the communication system between Agent and Server. With this change we are dictating it must be our GRPC tunnel. The original idea that if a provider wanted to swap it out for something like a VPN tunnel it would be relatively easy to do so. Granted that use case has never some up but wondering if we really want to lose that.

pkg/server/backend_manager_test.go
```diff
@@ -29,17 +29,17 @@ type fakeAgentServiceConnectServer struct {
 }
 
 func TestAddRemoveBackends(t *testing.T) {
-	conn1 := new(fakeAgentServiceConnectServer)
-	conn12 := new(fakeAgentServiceConnectServer)
-	conn2 := new(fakeAgentServiceConnectServer)
-	conn22 := new(fakeAgentServiceConnectServer)
-	conn3 := new(fakeAgentServiceConnectServer)
+	conn1 := NewBackend(new(fakeAgentServiceConnectServer))
```
cheftako, line 32: Would be better to rename these to backend1 etc given they are no longer conn objects.

pkg/server/server.go
```diff
@@ -255,79 +255,79 @@ func (s *ProxyServer) getBackend(reqHost string) (Backend, error) {
 	return nil, &ErrNotFound{}
 }
 
-func (s *ProxyServer) addBackend(agentID string, conn agent.AgentService_ConnectServer) (backend Backend) {
+func (s *ProxyServer) addBackend(agentID string, backend *Backend) {
```
cheftako, line 258: Having switched these from an interface being passed to a (pointer to)struct, are we sure we got all the pointer to parts? Otherwise we could have a nasty and subtle bug.

cheftako:
Not convinced we are gaining anything by removing the Backend interface. Otherwise looks good.

pkg/server/backend_manager_test.go
jkh52: Done.

pkg/server/server.go
jkh52: Moot point now, but:

One of the nasty and subtle bugs I chased down was that a unit test was still passing agent.AgentService_ConnectServer when it should pass a Backend.

I intentionally added a new method to Backend interface in order to break the interchangeability between the gRPC stream and the abstraction.

jkh52:
> Not convinced we are gaining anything by removing the Backend interface. Otherwise looks good.

See an earlier comment, I removed it based on http://go/gotip/episodes/49 and the fact that there is only a single concrete type left that implements Backend. I don't think the current Backend interface (which mostly copies gRPC stream API) necessarily lends itself to another implementation.

But, no harm in keeping this PR smaller until we have more consensus.

PR #525, opened 2023-09-29
proxy-server: strengthen the Backend interface.

jkh52:
As first proposed during review of https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/514, cache parsed agent identifiers. This moves error handling to happen sooner and helps clean up some APIs.

Notice the "agentID" naming collision fix in server.go (we were giving klog a func instead of a string...)

UPDATES:

Updated konnectivity-server to relax agent identifier parsing strictness. (Since it now always parses, instead of conditionally.)

Add relevant header_test.go.

### DIFF
--- a/pkg/server/backend_manager.go
+++ b/pkg/server/backend_manager.go
@@ -80,7 +80,8 @@ type Backend interface {
 	Send(p *client.Packet) error
 	Recv() (*client.Packet, error)
 	Context() context.Context
-	GetAgentIdentifiers() (header.Identifiers, error)
+	GetAgentID() string
+	GetAgentIdentifiers() header.Identifiers
 }
 
 var _ Backend = &backend{}
@@ -89,6 +90,10 @@ type backend struct {
 	sendLock sync.Mutex
 	recvLock sync.Mutex
 	conn     agent.AgentService_ConnectServer
+
+	// cached from conn.Context()
+	id     string
+	idents header.Identifiers
 }
 
 func (b *backend) Send(p *client.Packet) error {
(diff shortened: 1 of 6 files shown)


pkg/server/backend_manager.go
```diff
@@ -125,25 +130,53 @@ func (b *backend) Context() context.Context {
...
-func (b *backend) GetAgentIdentifiers() (header.Identifiers, error) {
+func (b *backend) GetAgentID() string {
+	return b.id
+}
+
+func (b *backend) GetAgentIdentifiers() header.Identifiers {
+	return b.idents
+}
+
+func getAgentID(stream agent.AgentService_ConnectServer) (string, error) {
```
jkh52, line 141: note to reviewer: this is moved from server.go

pkg/server/backend_manager.go
```diff
@@ -125,25 +130,53 @@ func (b *backend) Context() context.Context {
...
+	}
+	return agentIDs[0], nil
+}
+
+func getAgentIdentifiers(conn agent.AgentService_ConnectServer) (header.Identifiers, error) {
 	var agentIdentifiers header.Identifiers
-	md, ok := metadata.FromIncomingContext(b.Context())
+	md, ok := metadata.FromIncomingContext(conn.Context())
 	if !ok {
 		return agentIdentifiers, fmt.Errorf("failed to get metadata from context")
```
cheftako, line 157: One subtle difference. In the old code, for the DefaultBackendManager/default strategy case, we would not call GetAgentIdentifiers. Now we call if for all strategies. So previously if there would have been an error getting agent identifiers while only using the default strategy, I believe it would have "worked" but now it won't. I think thats probably ok but its worth paying attention to.

pkg/server/server_test.go
```diff
@@ -362,18 +362,17 @@ func prepareFrontendConn(ctrl *gomock.Controller) *agentmock.MockAgentService_Co
...
+		"agentid":          []string{uuid.New().String()},
 		"agentidentifiers": []string{},
 		"content-type":     []string{"application/grpc"},
 		"user-agent":       []string{"grpc-go/1.42.0"},
 	}
 	agentConnCtx := metadata.NewIncomingContext(context.Background(), agentConnMD)
 	agentConn.EXPECT().Context().Return(agentConnCtx).AnyTimes()
-	backend := NewBackend(agentConn)
-	proxyServer.addBackend(agentID, backend)
+	backend, _ := NewBackend(agentConn)
```
cheftako, line 374: Even in a test I don't think its a good idea to ignore this error. I grant we may only log it but if we hit this error in a test run I do think we will want it to have been logged.

pkg/server/server.go
```diff
@@ -829,6 +804,7 @@ func (s *ProxyServer) Connect(stream agent.AgentService_ConnectServer) error {
 }
 
 func (s *ProxyServer) readBackendToChannel(backend Backend, recvCh chan *client.Packet, stopCh chan error) {
+	agentID := backend.GetAgentID()
```
cheftako, line 807: Nice catch :smile:

cheftako:
Please fix the logging of the error in the test and then lgtm.

pkg/server/backend_manager.go
jkh52: I missed that, good point.

Maybe there is a case for `konnectivity-server` gracefully handling bad identifiers: if the agent binary ever introduces a new identifier type, then with binary skew (new agent, old server) the server could have problems. 

If there is agreement I can update this PR.

jkh52:
/hold

Open question: fail-open agent identifers parsing?

elmiko:
discussed at [today's community meeting](https://docs.google.com/document/d/1aPgGRl4WewM3txrCYvkepsxLUvGdMG1EzlVfCNeV74M/edit#bookmark=id.quzsc3mpig5g)

jkh52:
/hold cancel

In the latest snapshot, konnectivity-server will fail open on trivial agent identifier parsing. We can add the more strict validation to konnectivity-agent in a separate PR.

PR #528, opened 2023-10-09
Allow admin port to be disabled

tallclair:
Allow the agent & server "admin ports" to be disabled by setting them to 0. This is backwards compatible since the options validation previously forbid a value of 0.

My motivation is to have fewer ports to manage in integration tests, but this could also be useful for security reasons, especially since it's served over HTTP (no TLS)


tallclair:
/hold

I realized that although the agent serves metrics on the health port, the proxy server still uses the admin server for metrics. Needs further discussion.

PR #530, opened 2023-10-17
Upgrade grpc/http to fix CVE-2023-44487

muraee:
mitigates [CVE-2023-44487](https://github.com/advisories/GHSA-qppj-fm5r-hxr3)

see grpc pkg release notes https://github.com/grpc/grpc-go/releases/tag/v1.56.3


muraee:
cc @anfernee @jkh52

jkh52:
LGTM, but let's block this PR on https://github.com/kubernetes/kubernetes/pull/121273

(because we don't want konnectivity-client to get ahead of k/k for a given branch / minor version.)

jkh52:
This is unblocked by https://github.com/kubernetes/kubernetes/pull/121364

Side note - we could have not blocked on the above, and instead omit the konnectivity-client changes. Since that is only a client library, and no binaries are built from it.

In general, we may prefer to keep top level go.mod and konnectivity-client/go.mod in separate PRs (or at least separate commits). Thanks @cheftako for pointing it out offline.

PR #533, opened 2023-10-18
Update golang.org/x/net to v0.17.0

Argh4k:
Necessary for: https://security-tracker.debian.org/tracker/CVE-2023-3978 and https://security-tracker.debian.org/tracker/CVE-2023-44487


konnectivity-client/go.mod
jkh52, line 1: This is compatible with https://github.com/kubernetes/kubernetes/blob/release-1.28/go.mod

PR #541, opened 2023-11-28
Script the tag (release) process.

jkh52:
Script the current release process.

(Draft only - maybe there's a better approach altogether)


jparrill:
Looking good :), maybe we can go further and use GoReleaser also. WDYT?

cheftako:
/lifecycle/new

Issue #519, opened 2023-08-30
Version skew testing


tallclair:
As we make changes to Konnectivity, it is important that the newer components remain compatible with older components so that the system continues to function during upgrades and other version skew scenarios. However, we currently have no visibility into this. We should add tests that run against skewed versions.

Specifically, we need to test version skew across client/server, and agent/server. I propose we update the tests in the [tests](https://github.com/kubernetes-sigs/apiserver-network-proxy/tree/master/tests) directory to run against skewed versions.

/assign

jkh52:
https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/553 (github actions using `kind`) could be an approach here for how to specify the versions under test. (Similar to that, but use real container image registry + version tags).

tallclair:
The foundation is in place, but we still need to wire it up. Version skew testing is still needed, and will make it much safer to make protocol changes.

/triage accepted

PR #552, opened 2024-01-05
Run integration tests against an external agent process

tallclair:
This PR enables the agent executed by integration tests to be launched as a separate process from a prebuilt binary, chosen by the `--agent-path` flag.

Aside from the new agent runner, most of the changes in this PR are around the way metrics are tested, since they need to be scraped from the external process when used.

I will follow up with the external ProxyServer runner as a separate PR.

For https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/519

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,20 @@
+run:
+  timeout: 3m
+linters:
+  disable-all: true
+  enable:
+  - gofmt
+  - revive
+  - gosec
+  - govet
+  - unused
+issues:
+  fix: true
+  exclude-rules:
+    # Don't run security checks on test files
+    - path: _test\.go
+      linters:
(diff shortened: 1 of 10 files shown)


jkh52:
Please fix the lint error:

```
pkg/testing/metrics/metrics.go:136:9: indent-error-flow: if block ends with a return statement, so drop this else and outdent its block (revive)
	} else {
		return promtest.ScrapeAndCompare(t.Endpoint, strings.NewReader(expected), fqName)
	}
```

/approve

tallclair:
I fixed the security issues by excluding tests from the security linter. I think it's more likely to just create noise on test files.

jkh52:
arm64 build timeout

/retest

PR #553, opened 2024-01-10
kind example

aojea:
Add an example on how to run konnectivity proxy in a kind cluster
Add CI jobs to validate that konnectivity works with ipv4, ipv6 and dual stack

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -1,0 +1,220 @@
+name: e2e
+
+on:
+  push:
+    branches:
+      - 'master'
+    tags:
+      - 'v*'
+  pull_request:
+    branches: [ master ]
+  workflow_dispatch:
+
+env:
+  GO_VERSION: "1.21.4"
+  K8S_VERSION: "v1.27.3"
+  KIND_CLUSTER_NAME: "kind"
+
+jobs:
+  build:
+    name: build
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: ${{ env.GO_VERSION }}
+      id: go
+
+    - name: Check out code
+      uses: actions/checkout@v2
+
+    - name: Build
+      run: |
+        mkdir _output
+
+        docker build -t gcr.io/k8s-staging-kas-network-proxy/proxy-agent:master -f artifacts/images/agent-build.Dockerfile .
+        docker save gcr.io/k8s-staging-kas-network-proxy/proxy-agent:master > _output/konnectivity-agent.tar
+        docker build -t gcr.io/k8s-staging-kas-network-proxy/proxy-server:master -f artifacts/images/server-build.Dockerfile .
+        docker save gcr.io/k8s-staging-kas-network-proxy/proxy-server:master >  > _output/konnectivity-server.tar
+
+    - uses: actions/upload-artifact@v2
+      with:
+        name: konnectivity-server
+        path: _output/konnectivity-server.tar
+
+    - uses: actions/upload-artifact@v2
+      with:
+        name: konnectivity-agent
+        path: _output/konnectivity-agent.tar
+  e2e:
+    name: e2e
+    runs-on: ubuntu-20.04
+    timeout-minutes: 100
+    needs:
+      - build
+    strategy:
+      fail-fast: false
+      matrix:
+        ipFamily: ["ipv4", "ipv6", "dual"]
+    env:
+      JOB_NAME: "kindnetd-e2e-${{ matrix.ipFamily }}"
+      IP_FAMILY: ${{ matrix.ipFamily }}
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v2
+
+    - name: Enable ipv4 and ipv6 forwarding
+      run: |
+        sudo sysctl -w net.ipv6.conf.all.forwarding=1
+        sudo sysctl -w net.ipv4.ip_forward=1
+
+    - name: Set up environment (download dependencies)
+      run: |
+        TMP_DIR=$(mktemp -d)
+        # Test binaries
+        curl -L https://dl.k8s.io/${{ env.K8S_VERSION }}/kubernetes-test-linux-amd64.tar.gz -o ${TMP_DIR}/kubernetes-test-linux-amd64.tar.gz
+        tar xvzf ${TMP_DIR}/kubernetes-test-linux-amd64.tar.gz \
+          --directory ${TMP_DIR} \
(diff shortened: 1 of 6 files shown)


examples/kind/konnectivity-agent-ds.yaml
```diff
@@ -0,0 +1,70 @@
...
+          - mountPath: /var/run/secrets/tokens
+            name: konnectivity-agent-token
+      serviceAccountName: konnectivity-agent
+      volumes:
+      - name: konnectivity-agent-token
+        projected:
+          sources:
+          - serviceAccountToken:
+              path: konnectivity-agent-token
+              audience: system:konnectivity-server
```
aojea, line 70: @jkh52 @liggitt what is the easiest way to set up is for testing despite how insecure it is?

jkh52: The examples in this directory are relatively stale, I would start from the manifests in k/k which are kept up to date and have active prow test coverage (see https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/manifests/konnectivity-server.yaml and https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/addons/konnectivity-agent/konnectivity-agent-ds.yaml).

liggitt: what is the token used for?

jkh52: It looks like the token used for agent authentication.

But note that konnectivity-server can be configured to [skip agent authentication](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/cmd/server/app/server.go#L124)

aojea: ok, I managed to get it working, I'm adding CI in github too, I want to validate that it workswith ipv6 only

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,220 @@
+name: e2e
```
aojea, line 1: @BenTheElder @jkh52 do you know why the repo does not run the actions? do we have to enable them somewhere?

aojea: @jkh52 you have admin privileges in the repo https://github.com/kubernetes/org/blob/08dbac9ce837641b11cda031977364d94aa0258f/config/kubernetes-sigs/sig-cloud-provider/teams.yaml#L8

Can you go to the Settings field and enable github actions https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#managing-github-actions-permissions-for-your-repository

or do you want me to get admin permissions and enable them?

aojea: I want to have CI to validate on presubmit ipv4, ipv6 and dual stack, since we don't have it anywhere

aojea: ok, let me change the path

examples/kind/README.md
```diff
@@ -0,0 +1,51 @@
...
+```sh
+$ kubectl run test --image httpd:2
+pod/test created
+$ kubectl get pods
+NAME   READY   STATUS              RESTARTS   AGE
+test   0/1     ContainerCreating   0          4s
+$ kubectl get pods
+NAME   READY   STATUS    RESTARTS   AGE
+test   1/1     Running   0          6s
+$ kubectl exec -it test bash
```
jkh52, line 48: nit: the comment above (and the e2e.yaml) both say `kubectl logs`, want to keep exec or make it consistent?

.github/workflows/e2e.yaml
jkh52: Agreed, actions seem enabled (but I'm not experienced with them). We can look together if it is not working.

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,220 @@
...
+  e2e:
+    name: e2e
+    runs-on: ubuntu-20.04
+    timeout-minutes: 100
+    needs:
+      - build
+    strategy:
+      fail-fast: false
+      matrix:
+        ipFamily: ["ipv4", "ipv6", "dual"]
```
jkh52, line 59: in the ipv6 case, is the `httpd` pod below only given an IPv6 IP address? It might be nice to print it out for inspection.

aojea: Yes, that is guaranteed with kind ci, we can give that for granted

examples/kind/README.md
aojea: I think we can merge and iterate , I just needed to proof that ipv6 works, I will circle back to this but can't say when, in the meantime this adds a net value

aojea: oh, damn, I see now what you mean, let me change it

.github/workflows/e2e.yaml
aojea: we can check later why it does not work, but it does not seem a problem to merge it by now

examples/kind/README.md
aojea: fixed

jkh52:
flake?

```
--- FAIL: TestServerProxyNoBackend (0.00s)
    server.go:67: Unexpected call to *mock_agent.MockAgentService_ConnectServer.Send([type:DIAL_RSP dialResponse:{error:"No agent available" random:111}]) at /home/prow/go/src/sigs.k8s.io/apiserver-network-proxy/pkg/server/server.go:67 because: 
        expected call at /home/prow/go/src/sigs.k8s.io/apiserver-network-proxy/pkg/server/server_test.go:439 doesn't have a prerequisite call satisfied:
        *mock_agent.MockAgentService_ConnectServer.Recv() /home/prow/go/src/sigs.k8s.io/apiserver-network-proxy/pkg/server/server_test.go:436
        should be called before:
        *mock_agent.MockAgentService_ConnectServer.Send(is equal to type:DIAL_RSP dialResponse:{error:"No agent available" random:111} (*client.Packet)) /home/prow/go/src/sigs.k8s.io/apiserver-network-proxy/pkg/server/server_test.go:439
    controller.go:269: missing call(s) to *mock_agent.MockAgentService_ConnectServer.Send(is equal to type:DIAL_RSP dialResponse:{error:"No agent available" random:111} (*client.Packet)) /home/prow/go/src/sigs.k8s.io/apiserver-network-proxy/pkg/server/server_test.go:439
    controller.go:269: aborting test due to missing call(s)
```

/retest

jkh52:
/lgtm
/approve

Agreed this is a net positive, even if we don't see the github action run yet. Feel free to un-hold when ready to merge.

/hold

aojea:
/hold cancel

Thanks

PR #555, opened 2024-01-17
Fix github action e2e test.

liangyuanpeng:
This PR fixes the problem of e2e github action.

The main changes are as follows:

~~- Fix wrong shell command~~
~~- The kind config used in github action e2e uses the same file as exmaples/kind/kind.config~~
- Update kind version to v0.20.0
- Fix wrong config of kind.
- Temporary solution to flaky issue

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -80,7 +80,7 @@ jobs:
         # kubectl
         curl -L https://dl.k8s.io/${{ env.K8S_VERSION }}/bin/linux/amd64/kubectl -o ${TMP_DIR}/kubectl
         # kind
-        curl -Lo ${TMP_DIR}/kind https://kind.sigs.k8s.io/dl/v0.17.0/kind-linux-amd64
+        curl -Lo ${TMP_DIR}/kind https://kind.sigs.k8s.io/dl/v0.20.0/kind-linux-amd64
         # Install
         sudo cp ${TMP_DIR}/ginkgo /usr/local/bin/ginkgo
         sudo cp ${TMP_DIR}/e2e.test /usr/local/bin/e2e.test
@@ -103,6 +103,12 @@ jobs:
           ipFamily: ${IP_FAMILY}
         nodes:
           - role: control-plane
+            kubeadmConfigPatchesJSON6902:
+            - kind: ClusterConfiguration
+              patch: |
+                - op: add
+                  path: /apiServer/certSANs/-
+                  value: konnectivity-server.kube-system.svc.cluster.local
             kubeadmConfigPatches:
             - |
               kind: ClusterConfiguration
@@ -121,7 +127,10 @@ jobs:
             extraMounts:
             - hostPath: ./examples/kind/egress_selector_configuration.yaml
               containerPath: /etc/kubernetes/konnectivity-server-config/egress_selector_configuration.yaml
+          - role: worker
+          - role: worker
         EOF
+
         # dump the kubeconfig for later
         /usr/local/bin/kind get kubeconfig --name ${{ env.KIND_CLUSTER_NAME}} > _artifacts/kubeconfig.conf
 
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e.yaml
```diff
@@ -80,7 +80,7 @@ jobs:
         # kubectl
         curl -L https://dl.k8s.io/${{ env.K8S_VERSION }}/bin/linux/amd64/kubectl -o ${TMP_DIR}/kubectl
         # kind
-        curl -Lo ${TMP_DIR}/kind https://kind.sigs.k8s.io/dl/v0.17.0/kind-linux-amd64
+        curl -Lo ${TMP_DIR}/kind https://kind.sigs.k8s.io/dl/v0.20.0/kind-linux-amd64
```
liangyuanpeng, line 83: we will get the error if run with kind 0.17.0,  seems like it is https://github.com/kubernetes-sigs/kind/issues/3223

```
  STEP: Destroying namespace "statefulset-7177" for this suite. @ 01/17/24 08:46:32.114
  << Timeline

  [FAILED] RunHostCmd still failed after 5m3.82853998s: error running /usr/local/bin/kubectl --server=https://[::1]:34185 --kubeconfig=/home/runner/work/apiserver-network-proxy/apiserver-network-proxy/_artifacts/kubeconfig.conf --namespace=statefulset-7177 exec ss2-1 -- /bin/sh -x -c mv -v /usr/local/apache2/htdocs/index.html /tmp/ || true:
  Command stdout:

  stderr:
  error: Internal error occurred: error executing command in container: failed to exec in container: failed to start exec "f944756d1ae78400183b73440818e2b483fc255e64aa329e446e1aec455f5be9": OCI runtime exec failed: exec failed: unable to start container process: error adding pid 39487 to cgroups: failed to write 39487: openat2 /sys/fs/cgroup/unified/kubelet.slice/kubelet-kubepods.slice/kubelet-kubepods-besteffort.slice/kubelet-kubepods-besteffort-podb3dfd384_b40c_4151_aa21_aa6d091d86e4.slice/cri-containerd-e85ef49011f794f5fc1131d5e782fc37bb503a37e925a6fc93556bda5aecebbd.scope/cgroup.procs: no such file or directory: unknown
```

Do we have to use kind 0.17.0?  @aojea

.github/workflows/e2e.yaml
```diff
@@ -93,35 +93,12 @@ jobs:
...
+        /usr/local/bin/kind create cluster \
           --name ${{ env.KIND_CLUSTER_NAME}}           \
           --image kindest/node:${{ env.K8S_VERSION }}  \
-          -v7 --wait 1m --retain --config=-
-        kind: Cluster
-        apiVersion: kind.x-k8s.io/v1alpha4
-        networking:
-          ipFamily: ${IP_FAMILY}
-        nodes:
-          - role: control-plane
```
liangyuanpeng, line 105: Loss the config for multi node cluster, it's the reason i change it to use config of `examples/kind/kind.config`,avoid missing something.

.github/workflows/e2e.yaml
```diff
@@ -93,35 +93,12 @@ jobs:
...
-            kubeadmConfigPatches:
```
liangyuanpeng, line 106: Loss the config for ANP cert, it's the reason i change it to use config of `examples/kind/kind.config`,avoid missing something.

liangyuanpeng (commented):
Currently, the e2e test only runs on one k8s version. I also want e2e to run on multiple k8s versions at the same time, such as v1.27 v1.28 v1.29. 

Maybe it can be completed in the next PR.

.github/workflows/e2e.yaml
```diff
@@ -93,35 +93,12 @@ jobs:
         # output_dir
         mkdir -p _artifacts
         # create cluster
-        cat <<EOF | /usr/local/bin/kind create cluster \
+        sed -i  's/IP_FAMILY/'${IP_FAMILY}'/g' examples/kind/kind.config
+        sed -i  's#EGREE_SELECTOR_CONFIGURATION_PATH#./examples/kind/egress_selector_configuration.yaml#g' examples/kind/kind.config
```
aojea, line 97: does this work? kind.config does not have those values, does it?

liangyuanpeng: Thanks,Apparently I forgot to submit this file.

.github/workflows/e2e.yaml
```diff
@@ -93,35 +93,12 @@ jobs:
         # output_dir
         mkdir -p _artifacts
         # create cluster
-        cat <<EOF | /usr/local/bin/kind create cluster \
+        sed -i  's/IP_FAMILY/'${IP_FAMILY}'/g' examples/kind/kind.config
```
tallclair, line 96: nit: consider using shell variables directly in the config, and the `envsubst` utility to fill them.

.github/workflows/e2e.yaml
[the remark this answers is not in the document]
liangyuanpeng: I can see what you mean, but this use case is in CI and unlikely to pick it up  by git. Can you explain a little more? Thanks. :)

I can revert this part If you think it would be more appropriate to use EOF to complete the processing of the configuration files in CI .

jkh52:
I think EGREE_SELECTOR_CONFIGURATION_PATH should be named EGRESS_SELECTOR_CONFIGURATION_PATH

.github/workflows/e2e.yaml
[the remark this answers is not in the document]
liangyuanpeng: Let me revert it to eliminate noise caused by environment variables.

liangyuanpeng:
> I think EGREE_SELECTOR_CONFIGURATION_PATH should be named EGRESS_SELECTOR_CONFIGURATION_PATH

I have Revert it.

> Is that near ready to merge?

```shell
...
pod/coredns-5d78c9869d-k9c2h condition met
pod/coredns-5d78c9869d-xbwc7 condition met
pod/test created
pod/test condition met
Error from server: Get "https://172.18.0.2:10250/containerLogs/default/test/test": No agent available
Error: Process completed with exit code 1.
```

There is a problem waiting to be solved. I will come back to deal with it tomorrow, so it's not prepare to merge it yet.

@jkh52

jkh52:
@aojea and @liangyuanpeng, please note we temporarily disabled the action tests at https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/559. Hopefully this PR is able to make progress and then restore it. Let me know if you get stuck and want a second set of eyes on the "no agent available" problem.

liangyuanpeng:
> "no agent available"

This is a flaky problem, I am checking.

.github/workflows/e2e.yaml
```diff
@@ -143,79 +152,83 @@ jobs:
...
+        /usr/local/bin/kubectl get nodes -o wide
+        /usr/local/bin/kubectl get pods -A
+        /usr/local/bin/kubectl wait --timeout=1m --for=condition=ready pods --namespace=kube-system -l k8s-app=kube-dns
+        # smoke test
+        /usr/local/bin/kubectl run test --image httpd:2
+        /usr/local/bin/kubectl wait --timeout=1m --for=condition=ready pods test
+        /usr/local/bin/kubectl get pods -A -owide
+        if [ ${{ matrix.ipFamily }} = "ipv4" ];then 
+          sleep 50s
+        fi
```
liangyuanpeng, line 168:
```shell
...
2024-01-25T03:45:44.534421586Z stderr F I0125 03:45:44.534167       1 options.go:146] SyncForever set to false.
2024-01-25T03:45:44.552060196Z stderr F E0125 03:45:44.551978       1 clientset.go:188] "cannot connect once" err="rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp: lookup konnectivity-server.kube-system.svc.cluster.local on 10.96.0.10:53: no such host\""
2024-01-25T03:45:49.612075002Z stderr F E0125 03:45:49.611973       1 clientset.go:188] "cannot connect once" err="rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp: lookup konnectivity-server.kube-system.svc.cluster.local on 10.96.0.10:53: no such host\""
2024-01-25T03:45:57.263598167Z stderr F E0125 03:45:57.263510       1 clientset.go:188] "cannot connect once" err="rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp: lookup konnectivity-server.kube-system.svc.cluster.local on 10.96.0.10:53: no such host\""
2024-01-25T03:46:09.62534014Z stderr F E0125 03:46:09.625241       1 clientset.go:188] "cannot connect once" err="rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp: lookup konnectivity-server.kube-system.svc.cluster.local on 10.96.0.10:53: no such host\""
2024-01-25T03:46:27.877018898Z stderr F I0125 03:46:27.876930       1 client.go:210] "Connect to server" serverID="d8477c2e-b773-4c1e-be78-8cf8caa66ce2"
2024-01-25T03:46:27.877040759Z stderr F I0125 03:46:27.876951       1 clientset.go:222] "sync added client connecting to proxy server" serverID="d8477c2e-b773-4c1e-be78-8cf8caa66ce2"
2024-01-25T03:46:27.877080614Z stderr F I0125 03:46:27.876986       1 client.go:321] "Start serving" serverID="d8477c2e-b773-4c1e-be78-8cf8caa66ce2" agentID="ba7e8067-df06-45e4-94f4-c75333c559d5"
```

This is a flaky problem. The ANP agent cannot access the service normally within tens of seconds. see the log of ANP agent ^^

I didn't find the root cause, so I simply dealt with it using a temporary method.

I've tested 30s and it didn't work, i tested 15 times with 50s without problems, so it's a safe value.

liangyuanpeng (commented):
Since I will be on vacation next week, I hope to use a temporary method to merge this PR. After half a month of vacation, if the problem still needs to be handle, I can continue to look at it.

@aojea  @jkh52 @tallclair  PTAL,Thanks!

.github/workflows/e2e.yaml
```diff
@@ -143,79 +152,83 @@ jobs:
...
+        sleep 5
+        /usr/local/bin/kubectl get nodes -o wide
+        /usr/local/bin/kubectl get pods -A
+        /usr/local/bin/kubectl wait --timeout=1m --for=condition=ready pods --namespace=kube-system -l k8s-app=kube-dns
+        # smoke test
+        /usr/local/bin/kubectl run test --image httpd:2
+        /usr/local/bin/kubectl wait --timeout=1m --for=condition=ready pods test
+        /usr/local/bin/kubectl get pods -A -owide
+        if [ ${{ matrix.ipFamily }} = "ipv4" ];then 
+          sleep 50s
```
aojea, line 167: no sleeps, use `/usr/local/bin/kubectl wait --timeout=1m --for=condition=ready pods --namespace=kube-system -l ` and pick a label or somethign that selects the agent

liangyuanpeng: Actually, all of pod is running :( and the svc is fine. But the ANP agent complained that svc could not be found.

liangyuanpeng: The second temporary solution I can think of is to check the metrics of the ANP server. If  `connectivity_network_proxy_server_grpc_connections` = the size of ANP agent, then continue.

But as a temporary solution, I don’t think it’s worth investing in.

aojea: have you tried adding a readiness probe  on the daemonset/pods?

liangyuanpeng: Thanks for your remind, make sense for me, We no longer need sleep now.

I didn't try readiness at the beginning because I remembered that the ANP agent had no implementation that could check if the server was connected, until I found it. https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/485

Tested it 16 times and it's working.

.github/workflows/e2e.yaml
liangyuanpeng: Use readiness of ANP agent and not sleep, Thanks @aojea :)

aojea: thanks to you for fixing it 🥇

aojea:
/lgtm

/assign @jkh52 

I still don't see this action running 🤔

liangyuanpeng:
> I still don't see this action running 🤔

> This workflow requires approval from a maintainer. [Learn more about approving workflows.](https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)

Need the admin of this repo to approve this GitHub workflow. maybe @jkh52 

Seems like it's always need the approve, Because this is not my first contribution.

image

.github/workflows/e2e.yaml
```diff
@@ -143,79 +152,81 @@ jobs:
...
+        # wait network is ready
+        sleep 5
+        /usr/local/bin/kubectl get nodes -o wide
+        /usr/local/bin/kubectl get pods -A
+        /usr/local/bin/kubectl wait --timeout=1m --for=condition=ready pods --namespace=kube-system -l k8s-app=kube-dns
+        # smoke test
+        /usr/local/bin/kubectl run test --image httpd:2
+        /usr/local/bin/kubectl wait --timeout=1m --for=condition=ready pods test
+        /usr/local/bin/kubectl get pods -A -owide
+        /usr/local/bin/kubectl wait --timeout=1m --for=condition=ready pods --namespace=kube-system -l k8s-app=konnectivity-agent
```
jkh52, line 166: It may be worth also waiting for `-l k8s-app=konnectivity-server` to be ready.

(In this 1 control-plane node cluster is is nearly the same, but in a multi control-plane cluster it would be needed.)

jkh52:
/lgtm
/approve

Hold to allow Lan time to inspect logs, etc; feel free to un-hold when ready.
/hold

.github/workflows/e2e.yaml
liangyuanpeng: I'm assuming you're talking about waiting the pod of ANP server with 
readiness probe.(now, the pod of ANP server have not readiness probe)

konnectivity-server'readiness  indicates that at least one Konnectivity Agent is connected.

konnectivity-agent'readiness indicates  that the client is connected to at least one proxy server.

Therefore, if readiness is added at the same time, both will enter an infinite loop.

The reason is ANP agent is using kubernetes svc to connect ANP server and ANP server is running with daemonset.

Remind me if i missed something,Thanks.

liangyuanpeng:
I quickly tested the multi control-plane cluster ( the pod of ANP server without readiness probe) ) scenario using the current configuration and concluded that the test would fail.

There are three ANP servers and two ANP agents. When kubectl logs is requested to ANP serverA, serverA does not have any agent to connect to it, so the request fails.

But the test scenario is necessary and I will create an issue for it later.

/hold cnacel

.github/workflows/e2e.yaml
jkh52: > I'm assuming you're talking about waiting the pod of ANP server with readiness probe.(now, the pod of ANP server have not readiness probe)
> 
> konnectivity-server'readiness indicates that at least one Konnectivity Agent is connected.
> 
> konnectivity-agent'readiness indicates that the client is connected to at least one proxy server.

Yes to above.

> Therefore, if readiness is added at the same time, both will enter an infinite loop.

I don't expect this, because waiting on agent (or any) readiness does not depend on control plane egress.

> 
> The reason is ANP agent is using kubernetes svc to connect ANP server and ANP server is running with daemonset.
> 
> Remind me if i missed something,Thanks.

What really matters (for a given proxy request ,like `kubectl logs`) is whether the apiserver that handles the request has a konnectivity-server with at least one useful agent. Since the agent readiness only covers "at least 1", it is insufficient in this case. (There was discussion on that feature to add agent readiness mode "connected to all servers" but it is not implemented.) This would be the gap, fixed by waiting for all konnectivity-server to be ready.

jkh52:
Hooray, nice coverage @liangyuanpeng and @aojea !

aojea:
👏

PR #568, opened 2024-02-21
Fix e2e test for rootless image.

liangyuanpeng:
e2e CI is broken (https://github.com/kubernetes-sigs/apiserver-network-proxy/actions/runs/7935785136) after merge https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/564, the reason is permission denied when reading `/etc/kubernetes` with rootless image. This PR adds an initContainer to change the `/etc/kubernetes` directory permissions to fix this issue.

Another option is run root as user for ANP server pod,But this will not testing rootless container images.

```shell
...
      securityContext:
        runAsUser: 0
...
```

The error logs:
```shell
...
2024-02-16T20:20:47.774302668Z stderr F 
2024-02-16T20:20:47.774304462Z stderr F E0216 20:20:47.774123       1 main.go:48] error: failed to load kubernetes client config: error loading config file "/etc/kubernetes/admin.conf": open /etc/kubernetes/admin.conf: permission denied
```

/cc @jkh52 @cheftako @ipochi

### DIFF
--- a/examples/kind/konnectivity-server.yaml
+++ b/examples/kind/konnectivity-server.yaml
@@ -53,6 +53,9 @@ spec:
       nodeSelector:
         node-role.kubernetes.io/control-plane: ""
       hostNetwork: true
+      securityContext:
+        runAsUser: 1002
+        runAsGroup: 1000
       containers:
       - name: konnectivity-server-container
         image: gcr.io/k8s-staging-kas-network-proxy/proxy-server:master
@@ -64,9 +67,9 @@ spec:
           "--log-file=/var/log/konnectivity-server.log",
           "--logtostderr=true",
           "--log-file-max-size=0",
-          "--uds-name=/etc/kubernetes/konnectivity-server/konnectivity-server.socket",
-          "--cluster-cert=/etc/kubernetes/pki/apiserver.crt",
-          "--cluster-key=/etc/kubernetes/pki/apiserver.key",
+          "--uds-name=/etc/srv/kubernetes/konnectivity-server/konnectivity-server.socket",
+          "--cluster-cert=/etc/srv/kubernetes/pki/apiserver.crt",
+          "--cluster-key=/etc/srv/kubernetes/pki/apiserver.key",
           "--server-port=0",
           "--agent-port=8091",
           "--health-port=8092",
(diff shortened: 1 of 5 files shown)


examples/kind/konnectivity-server.yaml
```diff
@@ -53,6 +53,17 @@ spec:
...
         node-role.kubernetes.io/control-plane: ""
       hostNetwork: true
+      initContainers:
+      - name: busybox
+        image: busybox
+        command:
+        - sh
+        - -c
+        - |
+          chmod -R 777 /etc/kubernetes
```
cheftako, line 63: 774? 744? 444?

examples/kind/konnectivity-server.yaml
```diff
@@ -53,6 +53,17 @@ spec:
       nodeSelector:
         node-role.kubernetes.io/control-plane: ""
       hostNetwork: true
+      initContainers:
+      - name: busybox
```
cheftako, line 57: @tallclair Is busybox still the preferred image?

examples/kind/konnectivity-server.yaml
jkh52, line 63: in the manifest I see `/etc/srv/kubernetes`, should this match?

https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/examples/kubernetes/konnectivity-server.yaml

liangyuanpeng: > should this match?

Make sense for me.

liangyuanpeng: I have tested it with 444 446 and 447, only working with 447, This is a bit counter-intuitive since the ANP server only requires read permissions (except /etc/kubernetes/konnectivity-server ).

How about change this directory to `/etc/srv/kubernetes` ? To keep it consistent with `examples/kubernetes/konnectivity-server.yaml`, but the apiserver key still world-readable on `/etc/srv/kubernetes`  :(

liangyuanpeng: Does manually creating the required files and starting ANP with the 1002 user increase the confidence in merging the PR?

```shell
...
securityContext:
        runAsUser: 1002
        runAsGroup: 1000
...
```

PR #565, opened 2024-02-15
Bump kubernetes deps to v1.29.2.

jkh52:
Catch up master branch to kubernetes v1.29.2. After this we can do similar in release-0.29 branch, for https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/562.

PR #569, opened 2024-02-21
[release-0.29] Bump k8s deps to v0.29.2

### LINKED ISSUE #562: make release-0.29 branch consistent with k/k libs
We should do the same as https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/551, for release-0.29 branch. It currently uses [client-go v0.27.4](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/release-0.29/go.mod#L18C19-L18C26), but should use a v0.29 version.

liangyuanpeng:
- Bump k8s deps to v0.29.2
- Update golang version to 1.21

Fixes #562


jkh52:
Please see the two checks failing.

/lgtm

liangyuanpeng:
> Need to go mod tidy

Have no change after `go mod tidy` and it's working for `make test` on my local, let's test again with prow.

liangyuanpeng:
The reason of `pull-apiserver-network-proxy-test` failed is  prow testing with golang 1.21 and ANP release-0.29 is using 1.20. Kubernetes 1.29 is using 1.21, do we need to update golang version of ANP release-0.29 to 1.21?

@jkh52  If you want release-0.29 to continue to use 1.20, then I will create a PR for test-infra and add a golang1.20 job to work for release-0.29.

jkh52:
https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/565 just merged, let's proceed here.

> The reason of `pull-apiserver-network-proxy-test` failed is prow testing with golang 1.21 and ANP release-0.29 is using 1.20. Kubernetes 1.29 is using 1.21, do we need to update golang version of ANP release-0.29 to 1.21?
> 
> @jkh52 If you prefer release-0.29 to continue to use 1.20, then I will create a PR for test-infra and add a golang1.20 job to work for release-0.29.

It does seem better to use 1.21 in release-0.29, to align with k/k. I think it would be reasonable to do in this PR.

(What about older branches, like release-0.28? It seems unfortunate if those are similarly affected.)

PR #572, opened 2024-02-27
Docker build: parameterize base images.

jkh52:
Docker build: parameterize base images.

PR #579, opened 2024-03-08
Change Dockerfiles (and make targets) to use cross-compile.

jkh52:
Continue supporting ARCH env for backward compat, but introduce {BUILDARCH, TARGETARCH}.

This is a big speedup, compared to qemu. On my 10y old machine, `make docker-build-all` took less than 5m. For comparison, see historical latency (source: https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/274):

> Time consumed for each arch:
> 
> [arm](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_apiserver-network-proxy/274/pull-apiserver-network-proxy-docker-build-all/1460466982116659200): 1h28m59s
> [amd64](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_apiserver-network-proxy/274/pull-apiserver-network-proxy-docker-build-all/1460464877125505024): 2m13s
> [arm64](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_apiserver-network-proxy/274/pull-apiserver-network-proxy-docker-build-all/1460491386754174976):16m2s
> [ppc64le](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_apiserver-network-proxy/274/pull-apiserver-network-proxy-docker-build-all/1460567105475710976): 19m17s
> [s390x](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_apiserver-network-proxy/274/pull-apiserver-network-proxy-docker-build-all/1460573556063604736): 14m14s

At https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/572 we already changed the dockerfiles (support GO_TOOLCHAIN and GO_VERSION). It would be nice to group this additional change together (add to release-0.30).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# ARCH specifies the golang target platform, for docker build
+# targetes. It is supported for backward compatibility, but prefer
+# BUILDARCH and TARGETARCH.
 ARCH ?= amd64
+# BUILDARCH specifies the golang build platform, for docker build targets.
+BUILDARCH ?= amd64
+# TARGETARCH specifies the golang target platform, for docker build targets.
+TARGETARCH ?= $(ARCH)
 ARCH_LIST ?= amd64 arm arm64 ppc64le s390x
 RELEASE_ARCH_LIST = amd64 arm64
 # The output type could either be docker (local), or registry.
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -12,7 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# ARCH is still supported, but prefer BUILDARCH and TARGETARCH.
 ARCH ?= amd64
+BUILDARCH ?= amd64
+TARGETARCH ?= $(ARCH)
```
tallclair, line 18: What's the difference between BUILDARCH and TARGETARCH? I'm guessing BUILDARCH refers to the machine performing the build, and the TARGETARCH is the architecture the binary is built for?

jkh52: Right. Naming slightly influenced by https://docs.docker.com/build/building/multi-platform/#cross-compilation

Makefile
```diff
@@ -12,7 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# ARCH is still supported, but prefer BUILDARCH and TARGETARCH.
 ARCH ?= amd64
```
mainred, line 16: We removed ARCH in docker files and makefile targets, so it's not supported after this change?
It's a good idea to consider keeping ARCH for a few releases considering backward compatibility.

Makefile
mainred: BUILDARCH and TARGETARCH are user-facing, how about we add a comment to tell the difference?

Makefile
jkh52: That is the goal here. See line 18 - we assign ARCH as the default TARGETARCH. This will keep supporting build commands like `make docker-build/proxy-agent ARCH=...`.

Makefile
jkh52: > how about we add a comment to tell the difference?

Done.

jkh52:
Flake:

```
--- FAIL: TestProxyDial_RequestCancelled_Concurrent_GRPC (61.95s)
    proxy_test.go:399: Timed out waiting for tunnel to close
    proxy_test.go:426: Agent connections leaked: 
        
        Diff:
        --- metric output does not match expectation; want
        +++ got:
        @@ -2,3 +2,3 @@
         # TYPE konnectivity_network_proxy_agent_open_endpoint_connections gauge
        -konnectivity_network_proxy_agent_open_endpoint_connections 0
        +konnectivity_network_proxy_agent_open_endpoint_connections 1
```

/retest

PR #582, opened 2024-03-12
Set matrix k8s version for e2e ghaction.

liangyuanpeng:
matrix k8s version with v1.27.11, v1.28.7, v1.29.2.

Maybe releated: https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/519


jkh52:
Very cool! This provides nice skew coverage for the client/server case described in https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/519. It will still be good to target agent/server skew.

/lgtm
/approve

Issue #566, opened 2024-02-16
Ensure webhook availability during Konnectivity Agent rolling update


dippynark:
We are running Gatekeeper as a validating webhook on GKE (although I don't think the webhook implementation or cloud provider matters) and we have a test where we perform a rolling update of Gatekeeper while continuously making Kubernetes API requests that should be rejected to ensure requests/connections are being drained properly.

However, if we also delete Konnectivity Agent Pods while rolling Gatekeeper (gradually, to ensure that the Konnectivity Agent Pods aren't all down at the same time) or perform a rolling update (`kubectl rollout restart deployment -n kube-system konnectivity-agent`) then a few requests are allowed through (the `ValidatingWebhookConfiguration` is configured to fail open).

My question is whether this is an issue or whether Konnectivity Agent is behaving as expected? I guess this is happening because the long-lived HTTP keepalive connections between the Kubernetes API server and Gatekeeper (via Konnectivity Server and Konnectivity Agent) are being broken when Konnectivity Agent terminates and are not being drained properly (because there is no way for Konnectivity Agent to inspect the encrypted requests and disable keepalive before shutting down).

Should the Kubernetes API server be able to detect such TCP disconnects and retry validation after reconnecting?

dippynark:
@jkh52 In general we try to configure all cluster workloads to gracefully handle being evicted and rescheduled to a different node without any user impact, just to make it easy for us to change infrastructure configuration without users noticing. Specifically for Konnectivity Agent, we would like to avoid API server requests being unexpectedly rejected (when using fail closed webhooks) or unexpectedly accepted (when using fail open webhooks).

The following are the most common scenarios that would cause Konnectivity Agent to be evicted/rescheduled (and therefore could cause a potential webhook request failure):

- Cluster scale down (as you mentioned)
- Google reclaiming a compute instance (we are heavy users of spot VMs so this happens regularly)
- Changing the machine type of a node pool
- Upgrading a node pool
- Upgrading Konnectivity Agent

As you said, we can take steps to reduce the chance of Konnectivty Agent being evicted/rescheduled but we should expect failures under normal operation, so I think this issue can be closed unless there is any value in keeping it open

jkh52:
Thanks for providing details. Sounds good, I will close this.

jkh52:
Re-opening to track potential improvements around agent shutdown.

[...]

PR #584, opened 2024-03-12
Agent drain: implement agent side. (Remaining: server)

jkh52:
Agent drain: first half of https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/566

- Agent sends drain signal to Server.
- Server only logs it.
- Test cases.

Note that if a new agent binary sends the DRAIN message to an old server binary, it is harmless and will be ignored: https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/30266153048bd27662cdf878eae36efbb079941e/pkg/server/server.go#L994


cmd/agent/app/server.go
```diff
@@ -92,7 +95,31 @@ func (a *Agent) Run(o *options.GrpcProxyAgentOptions, stopCh <-chan struct{}) er
...
+	go runpprof.Do(context.Background(), labels, func(context.Context) { handleSignals(c, drain, stop) })
+
+	return drain, stop
+}
+
+func handleSignals(signalCh chan os.Signal, drainCh, stopCh chan struct{}) {
+	s := <-signalCh
+	klog.V(2).InfoS("Received first signal", "signal", s)
+	close(drainCh)
+	s = <-signalCh
```
cheftako, line 117: Once we begin draining, do we want to timeout and stop if no second signal is sent?

cheftako:
To be clear this is doing nothing to actually "drain"? Its just a prep step for that later PR?

cmd/agent/app/server.go
jkh52: I think the most cautious (and simple) decision is no, keep an agent in this state fully functioning.

jkh52:
> To be clear this is doing nothing to actually "drain"? Its just a prep step for that later PR?

Right, this PR should be a functional no-op for the agent, except for sending the draining hint to the server (for later PR: agent selection).

cheftako:
/lgtm
Looking forward to the actual drain logic being added. We clearly need to discuss edge cases such as what to do when ALL agents are in the drain state.

PR #599, opened 2024-03-25
Add test coverage around BackendManager and BackendStorage.

jkh52:
Includes a trivial API refactor.

Having this test coverage will give us confidence in a later BackendStorage refactor.


Makefile
```diff
@@ -66,6 +66,11 @@ mock_gen:
 
 .PHONY: test
 test:
+	go test -mod=vendor -race ./...
+	cd konnectivity-client && go test -race ./...
+
+.PHONY: test-coverage
+test-coverage: test
```
cheftako, line 73: A little concerned we have lowered what `make test` covers?

jkh52: per discussion I will keep `test` as-is, and split out a `fast-test`.

PR #602, opened 2024-03-30
chore: update actions upload-artifact & download-artifact to v4.

liangyuanpeng:
Resolve #598 and #597

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -222,7 +222,7 @@ jobs:
       if: always()
       uses: actions/upload-artifact@v4
       with:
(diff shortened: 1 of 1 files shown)


jkh52:
> Resolve #598 and #597

Both this PR and these two from dependabot have e2e failures, I haven't looked into it. @liangyuanpeng

liangyuanpeng:
the first problem is upload-artifact and download-artifact need to update with the same time,  it's resoved by the first commit of this PR.

The second problem is that the name of the log artifact lacks the k8s version, which leads to the problem of repeatedly uploading artifacts.It's resoved by the second commit.

All of CI is green now, tested on my fork repo.

PR #605, opened 2024-04-02
Fix lint: upgrade to first version that supports go1.21

jkh52:
Update lint version to one that supports go1.21: https://github.com/golangci/golangci-lint/releases/tag/v1.54.0 and fix fallout.

Although `make lint` passes on a dev machine, this PR fails lint check. Could the problem be that we need a newer [kubekins image](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/apiserver-network-proxy/apiserver-network-proxy-presubmits-master.yaml#L115)? Or, should we just pass that container an explicit GO_VERSION >= 1.21?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ endif
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
 INSTALL_LOCATION:=$(shell go env GOPATH)/bin
-GOLANGCI_LINT_VERSION ?= 1.51.2
+GOLANGCI_LINT_VERSION ?= 1.54.0
 GOSEC_VERSION ?= 2.13.1
 
 REGISTRY ?= gcr.io/$(shell gcloud config get-value project)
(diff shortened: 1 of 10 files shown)


jkh52:
@liangyuanpeng can you look? I suspect there is a test-infra change needed.

liangyuanpeng:
Yes, the go version is the reason of `make lint` failed, i have tested it and failed with 1.20 and success with 1.21.

And send the PR https://github.com/kubernetes/test-infra/pull/32366 to update the image of kubekins.

cheftako:
> Yes, the go version is the reason of `make lint` failed, i have tested it and failed with 1.20 and success with 1.21.
> 
> And send the PR [kubernetes/test-infra#32366](https://github.com/kubernetes/test-infra/pull/32366) to update the image of kubekins.

Approved the test-infra change. Lets wait for it to land and then retest here.

jkh52:
In latest lint run here, `prowjob.json` shows `gcr.io/k8s-staging-test-infra/kubekins-e2e:v20240329-fc17bdd16b-1.26` as desired.

PR #610, opened 2024-04-15
Refactor BackendManager / BackendStorage.

jkh52:
Refactor BackendManager / BackendStorage.

The primary reason I want this refactor is to make it simpler to finish https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/566. But this already large change has these benefits:

* fixes https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/294
* Moves more `--proxy-strategies` logic from ProxyServer to BackendManager (better separation of concerns)
* Clarify that "backends" map is not always keyed by agentID


pkg/server/backend_manager.go
```diff
@@ -196,176 +192,105 @@ func NewBackend(conn agent.AgentService_ConnectServer) (*Backend, error) {
...
-	// Set an explicit value, so that the metric is emitted even when
-	// no agent ever successfully connects.
-	metrics.Metrics.SetBackendCount(0)
-	return &DefaultBackendStorage{
-		backends: make(map[string][]*Backend),
-		random:   rand.New(rand.NewSource(time.Now().UnixNano())),
-		idTypes:  idTypes,
-	} /* #nosec G404 */
+func (s *DefaultBackendManager) Backend(addr string) (*Backend, error) {
+	for _, strategy := range s.proxyStrategies {
```
cheftako, line 235: Why do we want to consolidate all possible proxy strategies into a single Manager?

jkh52: > Why do we want to consolidate all possible proxy strategies into a single Manager?

At HEAD we have something of a mess, with 1:1 ProxyStrategy to BackendManager / BackendStorage, but logic spread between pkg/server/.*backend_manager.go and pkg/server/server.go. This change improves separation of concerns. For example, the main loop (pkg/server/server.go) does not need to be aware of proxy strategies.

After this consolidation, I can make the (single) BackendManager aware of agent drained state more safely.

Issue #586, opened 2024-03-15
Rationale behind xfrChannelSize set to const value of 10 and 150


ipochi:
I'd like to understand the reasons for choosing `const xfrChannelSize = 150` in [pkg/agent/client.go](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/bd2c976dcfc2dea2fb89a17da314efcb87a0354b/pkg/agent/client.go#L46) and `const xfrChannelSize = 10` in [pkg/server/server.go](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/bd2c976dcfc2dea2fb89a17da314efcb87a0354b/pkg/server/server.go#L50)

I see a lot of logs in the konnectivity-server with 
```
I0315 11:16:53.457102       1 server.go:853] "Receive channel from agent is full" agentID=0x15e8e00
I0315 11:16:53.457420       1 server.go:853] "Receive channel from agent is full" agentID=0x15e8e00
I0315 11:16:53.457882       1 server.go:853] "Receive channel from agent is full" agentID=0x15e8e00
I0315 11:16:53.458218       1 server.go:853] "Receive channel from agent is full" agentID=0x15e8e00
I0315 11:16:53.458260       1 server.go:853] "Receive channel from agent is full" agentID=0x15e8e00
I0315 11:16:53.458283       1 server.go:853] "Receive channel from agent is full" agentID=0x15e8e00
I0315 11:16:53.458292       1 server.go:853] "Receive channel from agent is full" agentID=0x15e8e00
I0315 11:16:53.458301       1 server.go:853] "Receive channel from agent is full" agentID=0x15e8e00
I0315 11:16:53.458317       1 server.go:853] "Receive channel from agent is full" agentID=0x15e8e00
I0315 11:16:53.458328       1 server.go:853] "Receive channel from agent is full" agentID=0x15e8e00
I0315 11:16:53.458340       1 server.go:853] "Receive channel from agent is full" agentID=0x15e8e00
I0315 11:16:53.458361       1 server.go:853] "Receive channel from agent is full" agentID=0x15e8e00
I0315 11:16:53.458382       1 server.go:853] "Receive channel from agent is full" agentID=0x15e8e00
I0315 11:16:53.458402       1 server.go:853] "Receive channel from agent is full" agentID=0x15e8e00
I0315 11:16:53.458419       1 server.go:853] "Receive channel from agent is full" agentID=0x15e8e00
[...]

PR #624, opened 2024-05-06
Allow configuration of transport channel size in agent and server

cnvergence:
See also:
https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/586

### DIFF
--- a/cmd/agent/app/options/options.go
+++ b/cmd/agent/app/options/options.go
@@ -79,7 +79,7 @@ type GrpcProxyAgentOptions struct {
 	WarnOnChannelLimit bool
 
 	SyncForever    bool
-	XfrChannelSize uint
+	XfrChannelSize int
 }
 
 func (o *GrpcProxyAgentOptions) ClientSetConfig(dialOptions ...grpc.DialOption) *agent.ClientSetConfig {
@@ -121,7 +121,7 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.AgentIdentifiers, "agent-identifiers", o.AgentIdentifiers, "Identifiers of the agent that will be used by the server when choosing agent. N.B. the list of identifiers must be in URL encoded format. e.g.,host=localhost&host=node1.mydomain.com&cidr=127.0.0.1/16&ipv4=1.2.3.4&ipv4=5.6.7.8&ipv6=:::::&default-route=true")
 	flags.BoolVar(&o.WarnOnChannelLimit, "warn-on-channel-limit", o.WarnOnChannelLimit, "Turns on a warning if the system is going to push to a full channel. The check involves an unsafe read.")
 	flags.BoolVar(&o.SyncForever, "sync-forever", o.SyncForever, "If true, the agent continues syncing, in order to support server count changes.")
-	flags.UintVar(&o.XfrChannelSize, "xfr-channel-size", 150, "Set the size of the channel")
+	flags.IntVar(&o.XfrChannelSize, "xfr-channel-size", 150, "Set the size of the channel for transferring data between the agent and the proxy server.")
 	return flags
 }
 
--- a/cmd/agent/app/options/options.go
+++ b/cmd/agent/app/options/options.go
@@ -78,7 +78,8 @@ type GrpcProxyAgentOptions struct {
 	// The check is an "unlocked" read but is still use at your own peril.
 	WarnOnChannelLimit bool
 
-	SyncForever bool
+	SyncForever    bool
+	XrfChannelSize int
 }
 
 func (o *GrpcProxyAgentOptions) ClientSetConfig(dialOptions ...grpc.DialOption) *agent.ClientSetConfig {
@@ -93,6 +94,7 @@ func (o *GrpcProxyAgentOptions) ClientSetConfig(dialOptions ...grpc.DialOption) 
 		ServiceAccountTokenPath: o.ServiceAccountTokenPath,
 		WarnOnChannelLimit:      o.WarnOnChannelLimit,
 		SyncForever:             o.SyncForever,
+		XrfChannelSize:          o.XrfChannelSize,
 	}
 }
 
@@ -119,6 +121,7 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.AgentIdentifiers, "agent-identifiers", o.AgentIdentifiers, "Identifiers of the agent that will be used by the server when choosing agent. N.B. the list of identifiers must be in URL encoded format. e.g.,host=localhost&host=node1.mydomain.com&cidr=127.0.0.1/16&ipv4=1.2.3.4&ipv4=5.6.7.8&ipv6=:::::&default-route=true")
 	flags.BoolVar(&o.WarnOnChannelLimit, "warn-on-channel-limit", o.WarnOnChannelLimit, "Turns on a warning if the system is going to push to a full channel. The check involves an unsafe read.")
 	flags.BoolVar(&o.SyncForever, "sync-forever", o.SyncForever, "If true, the agent continues syncing, in order to support server count changes.")
+	flags.IntVar(&o.XrfChannelSize, "channel-size", 150, "Set the size of the channel")
 	return flags
 }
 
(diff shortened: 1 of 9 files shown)


cmd/agent/app/options/options.go
```diff
@@ -119,6 +121,7 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.AgentIdentifiers, "agent-identifiers", o.AgentIdentifiers, "Identifiers of the agent that will be used by the server when choosing agent. N.B. the list of identifiers must be in URL encoded format. e.g.,host=localhost&host=node1.mydomain.com&cidr=127.0.0.1/16&ipv4=1.2.3.4&ipv4=5.6.7.8&ipv6=:::::&default-route=true")
 	flags.BoolVar(&o.WarnOnChannelLimit, "warn-on-channel-limit", o.WarnOnChannelLimit, "Turns on a warning if the system is going to push to a full channel. The check involves an unsafe read.")
 	flags.BoolVar(&o.SyncForever, "sync-forever", o.SyncForever, "If true, the agent continues syncing, in order to support server count changes.")
+	flags.IntVar(&o.XrfChannelSize, "channel-size", 150, "Set the size of the channel")
```
jkh52, line 124: Likely typos:

* XrfChannelSize -> XfrChannelSize
* "channel-size" -> "xfr-channel-size"

both for consistency between agent and server flags.

cmd/server/app/options/options.go
```diff
@@ -136,6 +137,7 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.AuthenticationAudience, "authentication-audience", o.AuthenticationAudience, "Expected agent's token authentication audience (used with agent-namespace, agent-service-account, kubeconfig).")
 	flags.StringVar(&o.ProxyStrategies, "proxy-strategies", o.ProxyStrategies, "The list of proxy strategies used by the server to pick an agent/tunnel, available strategies are: default, destHost, defaultRoute.")
 	flags.StringSliceVar(&o.CipherSuites, "cipher-suites", o.CipherSuites, "The comma separated list of allowed cipher suites. Has no effect on TLS1.3. Empty means allow default list.")
+	flags.IntVar(&o.XrfChannelSize, "xfr-channel-size", o.XrfChannelSize, "The size of the channel for transferring data between the proxy server and the agent.")
```
jkh52, line 140: I would validate (both) to reject zero value. (Buffered channel of size zero would not function.) We may also want to reject too-small of values, but not sure what the bare minimum could be. @cheftako

jkh52, line 140: Prevent negative values with flags.UintVar

cmd/agent/app/options/options.go
cnvergence: Right, thanks @jkh52!

cmd/server/app/options/options.go
cnvergence: I have added validation for the 0 value, but let's agree on the value of minimum.
I will add test cases

cmd/agent/app/options/options.go
```diff
@@ -119,6 +121,7 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.AgentIdentifiers, "agent-identifiers", o.AgentIdentifiers, "Identifiers of the agent that will be used by the server when choosing agent. N.B. the list of identifiers must be in URL encoded format. e.g.,host=localhost&host=node1.mydomain.com&cidr=127.0.0.1/16&ipv4=1.2.3.4&ipv4=5.6.7.8&ipv6=:::::&default-route=true")
 	flags.BoolVar(&o.WarnOnChannelLimit, "warn-on-channel-limit", o.WarnOnChannelLimit, "Turns on a warning if the system is going to push to a full channel. The check involves an unsafe read.")
 	flags.BoolVar(&o.SyncForever, "sync-forever", o.SyncForever, "If true, the agent continues syncing, in order to support server count changes.")
+	flags.UintVar(&o.XfrChannelSize, "xfr-channel-size", 150, "Set the size of the channel")
```
cheftako, line 124: Would be good to indicate the "direction" of the channel transfer (into or out of the cluster)

cheftako: 150 -> o.XfrChannelSize

cmd/agent/app/options/options.go
```diff
@@ -78,7 +78,8 @@ type GrpcProxyAgentOptions struct {
 	// The check is an "unlocked" read but is still use at your own peril.
 	WarnOnChannelLimit bool
 
-	SyncForever bool
+	SyncForever    bool
+	XfrChannelSize uint
```
cheftako, line 82: Would it be easier to make this an int to avoid casting and possible overflow errors?

cmd/server/app/options/options.go
```diff
@@ -136,6 +137,7 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.AuthenticationAudience, "authentication-audience", o.AuthenticationAudience, "Expected agent's token authentication audience (used with agent-namespace, agent-service-account, kubeconfig).")
 	flags.StringVar(&o.ProxyStrategies, "proxy-strategies", o.ProxyStrategies, "The list of proxy strategies used by the server to pick an agent/tunnel, available strategies are: default, destHost, defaultRoute.")
 	flags.StringSliceVar(&o.CipherSuites, "cipher-suites", o.CipherSuites, "The comma separated list of allowed cipher suites. Has no effect on TLS1.3. Empty means allow default list.")
+	flags.UintVar(&o.XfrChannelSize, "xfr-channel-size", o.XfrChannelSize, "The size of the channel for transferring data between the proxy server and the agent.")
```
cheftako, line 140: We have 2 transfer channels in the server. 1) KAS -> KNP Server -> KNP Agent and 2) KNP Agent -> KNP Server -> KAS. This makes both channels be set with the same transfer channel size. Do we want to allow them to be configured to different values?

cheftako:
I like the change it seems valuable. Just a couple of minor questions it would be good to iron out first.

cmd/agent/app/options/options.go
cnvergence: I have changes to uint due to the previous comment, but taking overflow errors into account, I will use int again and validate against negative and zero values.

cmd/server/app/options/options.go
cnvergence: Initially, they were set to the same value, but in this case we could also make it configurable

cnvergence:
PTAL again :)

pkg/server/server.go
```diff
@@ -417,7 +417,7 @@ func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
 	streamUID := uuid.New().String()
 	klog.V(5).InfoS("Proxy request from client", "userAgent", userAgent, "serverID", s.serverID, "streamUID", streamUID)
 
-	recvCh := make(chan *client.Packet, xfrChannelSize)
+	recvCh := make(chan *client.Packet, s.xfrChannelSize)
```
cheftako, line 420: This is the transfer channel for traffic coming from the KAS (frontend). However the options indicate that the size is only for the tunnel segment between the server and agent. We should either fix the flag description to mention both tunnel segments or add a separate configuration for each tunnel segment.

cnvergence: I have decided to fix the flag, as I didn't want to add another configuration flag, to keep it at the same value as it was originally set.

cmd/server/app/options/options.go
```diff
@@ -136,6 +137,7 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.AuthenticationAudience, "authentication-audience", o.AuthenticationAudience, "Expected agent's token authentication audience (used with agent-namespace, agent-service-account, kubeconfig).")
 	flags.StringVar(&o.ProxyStrategies, "proxy-strategies", o.ProxyStrategies, "The list of proxy strategies used by the server to pick an agent/tunnel, available strategies are: default, destHost, defaultRoute.")
 	flags.StringSliceVar(&o.CipherSuites, "cipher-suites", o.CipherSuites, "The comma separated list of allowed cipher suites. Has no effect on TLS1.3. Empty means allow default list.")
+	flags.IntVar(&o.XfrChannelSize, "xfr-channel-size", o.XfrChannelSize, "The size of the two KNP server channels used in server for transferring data. One channel is for data coming from the Kubernetes API Server, and the other one is for data coming from the KNP agent.")
```
cnvergence, line 140: @cheftako just let me know, how it does look for you now

cheftako:
/lgtm
/approve
Thanks for making this change!

PR #628, opened 2024-06-07
Nit: go get -> go install in build instructions

carreter:
`go get` has been deprecated for package installation since [1.16](https://tip.golang.org/doc/go1.16#modules):

> go install, with or without a version suffix (as described above), is now the recommended way to build and install packages in module mode. go get should be used with the -d flag to adjust the current module’s dependencies without building packages, and use of go get to build and install packages is deprecated. In a future release, the -d flag will always be enabled.

Using `go get` instead of `go install` resulted in some strange errors when trying to build. Changing to `go install` fixed the issue.

Also add mention of adding `~/go/bin` to the `PATH`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ For local builds these can be set to anything.
 For image builds these determine the location of your image.
 For GCE the registry should be gcr.io and PROJECT_ID should be the project you
 want to use the images in.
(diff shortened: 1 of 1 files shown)


jkh52:
Thanks for improving docs!

/ok-to-test

Please see the cncf-cla requirement and fix it.

PR #631, opened 2024-06-12
Add script to quickly deploy kind cluster with custom KNP image

carreter:
Thought I'd share a script I wrote while setting up my local dev environment.

`./quickstart-kind.sh` lets you set how many KCP + worker nodes the cluster has, modify which image is pulled for both the proxy server and agent, and optionally sideload the images into the cluster to avoid the need for a publishing the image on a public or local registry.


carreter:
@avrittrohwer for visibility :)

examples/kind-multinode-kcp/egress_selector_configuration.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: apiserver.k8s.io/v1beta1
+kind: EgressSelectorConfiguration
+egressSelections:
+- name: cluster
+  connection:
+    proxyProtocol: GRPC
+    transport:
+      uds:
+        udsName: /etc/kubernetes/konnectivity-server/konnectivity-server.socket
+- name: master
```
jkh52, line 10: this is deprecated in favor of 'controlplane'

examples/kind-multinode-kcp/templates/k8s/konnectivity-agent-ds.yaml
```diff
@@ -0,0 +1,95 @@
...
+        k8s-app: konnectivity-agent
+    spec:
+      priorityClassName: system-cluster-critical
+      tolerations:
+        - key: "CriticalAddonsOnly"
+          operator: "Exists"
+        - operator: "Exists"
+          effect: "NoExecute"
+      nodeSelector:
+        kubernetes.io/os: linux
```
jkh52, line 35: Could add `node-role.kubernetes.io/worker: ''` to prevent agents on control plane nodes?

examples/kind-multinode-kcp/egress_selector_configuration.yaml
cheftako: +1

examples/kind-multinode-kcp/quickstart-kind.sh
```diff
@@ -0,0 +1,116 @@
...
+do
+ cat templates/kind/control-plane.config >> rendered/kind.config
+done
+for i in $(seq 0 "$NUM_WORKER_NODES")
+do
+ cat templates/kind/worker.config >> rendered/kind.config
+done
+
+echo "Setting server image to $SERVER_IMAGE and agent image to $AGENT_IMAGE"
+sed "s/image: .*/image: $(quoteSubst "$AGENT_IMAGE")/" <templates/k8s/konnectivity-agent-ds.yaml >rendered/konnectivity-agent-ds.yaml
```
tallclair, line 89: Rather than escaping the image string, can you just change the sed delimiter?
```suggestion
sed "s|image: .*|image: ${AGENT_IMAGE}|" <templates/k8s/konnectivity-agent-ds.yaml >rendered/konnectivity-agent-ds.yaml
```

examples/kind-multinode-kcp/templates/k8s/konnectivity-agent-ds.yaml
```diff
@@ -0,0 +1,95 @@
...
+          "--logtostderr=true",
+          "--ca-cert=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
+          "--proxy-server-host=konnectivity-server.kube-system.svc.cluster.local",
+          "--proxy-server-port=8091",
+          "--sync-interval=5s",
+          "--sync-interval-cap=30s",
+          "--sync-forever",
+          "--probe-interval=5s",
+          "--service-account-token-path=/var/run/secrets/tokens/konnectivity-agent-token",
+          "--agent-identifiers=ipv4=$(HOST_IP)"
```
tallclair, line 56: This should be a shell variable, not command, right?
```suggestion
          "--agent-identifiers=ipv4=${HOST_IP}"
```

carreter: So *that's* the difference between `${}` and `$()`...

Fixed, thank you!

examples/kind-multinode-kcp/templates/k8s/konnectivity-agent-ds.yaml
carreter: Done!

examples/kind-multinode-kcp/quickstart-kind.sh
carreter: Didn't know you could do that, that's awesome!

examples/kind-multinode-kcp/templates/k8s/konnectivity-agent-ds.yaml
carreter: Actually, seems like setting this label on the node in the `kind.config` breaks the cluster for some reason. `kind` doesn't label the worker nodes as such by default for some reason, and `kubelet` doesn't start if I provide a custom role label. Leaving this out for now.

carreter: Did some more testing and opened an upstream issue: https://github.com/kubernetes-sigs/kind/issues/3657

BenTheElder: kubelet doesn't permit adding arbitrary lables in the k8s.io / kubernetes.io namespace because those are reserved for Kubernetes and need to be approved by the API review team.

You can instead use taints and tolerations to schedule only to worker nodes. 

Kind removes the control plane taint only if there is a single control plane node and no other nodes (so a one node cluster), otherwise the control plane node(s) are tainted, see:
https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm/#control-plane-node-isolation

carreter:
Apologies for force-pushing and messing up the commit chain, I borked my local git repo.

Addressed review comments and made it so the script also sets the appropriate `--server-count` flag on the proxy server based on the number of requested KCP nodes (you can override and set it manually with `--server-count-override <NUM>`).

examples/kind-multinode-kcp/templates/k8s/konnectivity-agent-ds.yaml
carreter: @jkh52 Is it ok if I leave this as-is? I haven't noticed proxy agents getting spun up on the control plane anyway.

jkh52: No objections. Are taints + tolerations the reason that you don't see agents on control plane? Did you try a one-node cluster as well?

BenTheElder: On a one-node cluster, that node is the worker and the control plane host, so scheduling user workloads and control plane workloads would be expected, though I'm not sure it makes sense to even try deploying this application to a one node cluster ..?

On a multi-node cluster there's the well-known kubeadm taint on control plane nodes, unless you tolerate _all_ taints or that [taint](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) specifically your pod won't schedule there. It's pretty typical that a cluster with registered control plane nodes would have a similar taint if not the same taint.

cheftako:
/lgtm
/approve
Thanks for adding these instructions!

PR #632, opened 2024-06-13
Always reset agent connection backoff and enter fast sync when client count < server count

carreter:
Part of #358 (see https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/358#issuecomment-2166813376).

## Current behavior
Currently, if an agent receives an increased server count from a server it is already connected to, it doesn't reset the backoff and just uses the last set duration. This could result in an agent taking a long time to connect to new proxy servers if it receives the updated server count from a server it is already connected to.

This cannot currently happen as there is no way to update the server count on a proxy server without restarting it and dropping the connection, but this will be an issue once #273 is implemented.

## Testing the fix
Using #631, I set up https://github.com/carreter/apiserver-network-proxy/tree/backoff-reset-fix-test-example . Run the following in the repo root to set up a cluster with 1 proxy server on a KCP node and 1 proxy agent on a worker node:

```sh
make docker-build

cd examples/kind-multinode

# These are the default values of the registry, image name, and tag used by the Makefile.
# Edit them if necessary.
REGISTRY=gcr.io/$(gcloud config get-value project)
TAG=$(git rev-parse HEAD)
TARGET_ARCH="amd64"
SERVER_IMAGE="$REGISTRY/proxy-server-$TARGET_ARCH:$TAG"
AGENT_IMAGE="$REGISTRY/proxy-agent-$TARGET_ARCH:$TAG"

# Bring up the cluster!
./quickstart-kind.sh --cluster-name custom-knp-test --server-image "$SERVER_IMAGE" --agent-image "$AGENT_IMAGE" \
  --num-kcp-nodes 1 --num-worker-nodes 1 --sideload-images
```

Then, apply `server-count-file.yaml` with the following to upload a file with an updated server count:
```sh
kubectl apply -f server-count-file.yaml
```

Give it a couple seconds and you should see logs showing that the client is issuing new connection requests every 5 secs (i.e. `--sync-interval`)  instead of every 30 secs (i.e. `--sync-interval-cap`) like it would without this fix!
```
I0613 22:04:19.702443       1 client.go:215] "Connect to server" serverID="50ea06e6-89fa-444a-9f8e-1ee9bf0d4a22"
I0613 22:04:19.702687       1 clientset.go:186] "duplicate server" serverID="50ea06e6-89fa-444a-9f8e-1ee9bf0d4a22" serverCount=1 clientsCount=1
I0613 22:04:19.703466       1 client.go:474] "received DATA" connectionID=1
[...]

Issue #273, opened 2021-10-15
Support dynamic number of proxy-servers


jkh52:
Feature Request: support clusters with dynamic number of proxy-servers.

Example use case: gracefully add a 2nd control plane node to a 1 control plane node cluster.

Current state:
* proxy-server takes static --server-count flag for the life of the process.
* proxy-agent learns the count from a Connect RPC response, and caches the first-seen value for the life of the process.

**Proxy Agent**
We need proxy-agent syncOnce() to stop short circuiting as aggressively.

Option 1: add new server RPC GetServerCount(); call at the top of syncOnce()

This seems logically the cleanest, but a big downside is an additional authentication (compared with Connect).

Option 2: have syncOnce() still try at some lower rate, even when connected to the last-seen server count.

**Proxy Server**

We could keep the --server-count flag but add a mutually exclusive --server-count-file to support a dynamic config value (avoid restarting the process).

Any other suggestions or example patterns for this side?

**Expected Behavior**

Add one master node:
* proxy-agent will notice and establish connections within a predictable / deterministic time.
* tunnels through existing proxy-server(s) should not be interrupted.

Subtract one master node:
* we cannot avoid disrupting some tunnels in this case.
* proxy-server process shutdown is the mechanism that will result in proxy-agent cleaning up old connections (keepalive).
* proxy-agent may briefly see lower server count value than actual observed unique server_ids. This shouldn't cause it to panic or trigger any cleanup.

jtherin:
> but add a mutually exclusive --server-count-file to support a dynamic config value (avoid restarting the process).
Why not a ConfigMap ?

jkh52:
> Why not a ConfigMap ?

Assuming you mean: proxy-server use a ConfigMap (set by a cluster admin or bootstrapping infra) to learn accurate ServerCount. This is unfortunate because then the value is potentially visible to the cluster (even the proxy-agent) when we already have Connect() RPC protocol for the agent to learn server count.

zqzten:
+1 for this feature

For server side, I wonder if we can find a way for servers to communicate with each other to dynamically get the replica count rather than depending on explicit external config.

janiskemper:
+1 - would be very useful for clusters managed by Cluster API

jkh52:
The current status:

**Proxy Agent**

[...]

PR #634, opened 2024-06-18
New proxy server counting logic for agent and server

carreter:
An initial pass at the groundwork for #358 and #273 (i.e. allowing a dynamic proxy server count on both the server and the agent in a backwards-compatible manner). Currently a draft, depends on #632. Design doc [here](https://docs.google.com/document/d/16G8U-QWR4tNfY2jiOk-ZUEMVwBkq9R9iXZqXrxPsGOE/edit?usp=sharing).

### New package: `pkg/servercounter`
This package is structured around the `ServerCounter` interface:

```go
// A ServerCounter counts the number of available proxy servers.
type ServerCounter interface {
	CountServers() int
}
```

Currently provided are `StaticServerCounter`, which simply returns a stored int, and `CachedServerCounter`, which wraps another `ServerCounter` and provides catching functionality with a configurable cache expiry time.

### Proxy server
This PR wires the `ServerCounter` into the proxy server without changing its behavior by creating a `StaticServerCounter` with the command-line provided count.

### Proxy agent
This PR wires the `ServerCounter` into the proxy agent without changing its behavior by creating a `StaticServerCounter` with a default count of 0.

This PR also adds the `respectReceivedServerCount` field to the agent `ClientSet` struct. This field toggles whether the server count received from the proxy server should overwrite the proxy agent's server counter. Since I didn't want to change the behavior of the agent yet, this is set to `true` by default.

### Should `CountServers()` return an `error`?
Sometimes the `ServerCounter` will fail to get an updated server count. In this case, we could have `CountServers()` return `(int, error)` instead.

I opted against this because we can handle errors in the `ServerCounter` itself. This simplifies things on the caller side and allows for including a fallback server count in the `ServerCounter`.

### Should `ServerCounter` include an `UpdateCount(int)` method?
It might be useful for clients to (a) temporarily override the server count or (b) request an update to any cached server counts. They could do this by calling `UpdateCount()` with a non-negative value to temporarily override the count or with -1 to reset the override and request a count update.

I opted against this because the client can accomplish (a) using a `StaticServerCounter` (downside: this overwrites the previous `ServerCounter` if stored in the same field), and (b) seems like it would have limited use as the `ServerCounter` should be completely responsible for managing the server count.

A possible workaround is to add the `UpdateCount()` method to *just* `CachedServerCounter` rather than the entire interface. This allows for finer control of caching while maintaining a simple API for the general case.

### Should `respectReceivedServerCount` be `true` or `false` by default?
My take: have it `true` by default unless an alternative server count method is provided in the commandline args to the proxy agent. This ensures backwards compatibility with old proxy servers.


carreter:
@avrittrohwer for visibility

pkg/agent/clientset.go
```diff
@@ -63,6 +64,8 @@ type ClientSet struct {
 	warnOnChannelLimit bool
 
 	syncForever bool // Continue syncing (support dynamic server count).
+
+	respectReceivedServerCount bool // Respect server count received from proxy server rather than relying on the agent's own server counter
```
cheftako, line 68: Should chat on this. During certain events with multiple KASs/Konnectivity servers we can get different server counts from different servers.

carreter: Yeah, @avrittrohwer also raised this issue. One potential solution would be to cache the last n received server counts and take the highest value. Ultimately this gets solved by having proxy agents count the leases themselves.

PR #635, opened 2024-06-20
New proxy server counting logic for agent and server

carreter:
Accidentally close #634, reopening it here.

An initial pass at the groundwork for #358 and #273 (i.e. allowing a dynamic proxy server count on both the server and the agent in a backwards-compatible manner). Currently a draft, depends on #632. Design doc [here](https://docs.google.com/document/d/16G8U-QWR4tNfY2jiOk-ZUEMVwBkq9R9iXZqXrxPsGOE/edit?usp=sharing).

### New package: `pkg/servercounter`
This package is structured around the `ServerCounter` interface:

```go
// A ServerCounter counts the number of available proxy servers.
type ServerCounter interface {
	CountServers() int
}
```

Currently provided are `StaticServerCounter`, which simply returns a stored int, and `CachedServerCounter`, which wraps another `ServerCounter` and provides catching functionality with a configurable cache expiry time.

### Proxy server
This PR wires the `ServerCounter` into the proxy server without changing its behavior by creating a `StaticServerCounter` with the command-line provided count.

### Proxy agent
This PR wires the `ServerCounter` into the proxy agent without changing its behavior by creating a `StaticServerCounter` with a default count of 0.

This PR also adds the `respectReceivedServerCount` field to the agent `ClientSet` struct. This field toggles whether the server count received from the proxy server should overwrite the proxy agent's server counter. Since I didn't want to change the behavior of the agent yet, this is set to `true` by default.

### Should `CountServers()` return an `error`?
Sometimes the `ServerCounter` will fail to get an updated server count. In this case, we could have `CountServers()` return `(int, error)` instead.

I opted against this because we can handle errors in the `ServerCounter` itself. This simplifies things on the caller side and allows for including a fallback server count in the `ServerCounter`.

### Should `ServerCounter` include an `UpdateCount(int)` method?
It might be useful for clients to (a) temporarily override the server count or (b) request an update to any cached server counts. They could do this by calling `UpdateCount()` with a non-negative value to temporarily override the count or with -1 to reset the override and request a count update.

I opted against this because the client can accomplish (a) using a `StaticServerCounter` (downside: this overwrites the previous `ServerCounter` if stored in the same field), and (b) seems like it would have limited use as the `ServerCounter` should be completely responsible for managing the server count.

A possible workaround is to add the `UpdateCount()` method to *just* `CachedServerCounter` rather than the entire interface. This allows for finer control of caching while maintaining a simple API for the general case.

### Should `respectReceivedServerCount` be `true` or `false` by default?
My take: have it `true` by default unless an alternative server count method is provided in the commandline args to the proxy agent. This ensures backwards compatibility with old proxy servers.


carreter:
@avrittrohwer MVP server lease counter with unit tests is done, just need to make e2e tests!

carreter:
I think it's working now, at least on the agent side! It successfully detects the 2 apiserver leases when I configure it with the proper label selector, and establishes connections accordingly.

image

The code still needs some cleaning up though...

carreter:
Consider the code under `pkg/servercounter` ready for review. Remaining tasks:

- [ ] e2e tests
- [ ] clean up commandline options
- [ ] linting
- [ ] more logging/metrics

pkg/servercounter/lease_counter.go
```diff
@@ -0,0 +1,73 @@
+package servercounter
+
+import (
+	"fmt"
+	"time"
+
+	coordinationv1api "k8s.io/api/coordination/v1"
+	"k8s.io/apimachinery/pkg/labels"
```
avritt-techx, line 8: stylistic nitpick: put named imports below unnamed in a separate block:

```
"k8s.io/apimachinery/pkg/labels"
"k8s.io/klog/v2"

coordinationv1api "k8s.io/api/coordination/v1"
coordinationv1listers "k8s.io/client-go/listers/coordination/v1"
```

pkg/servercounter/lease_counter.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	return &ServerLeaseCounter{
+		lister:        lister,
+		selector:      selector,
+		fallbackCount: fallbackCount,
+	}, nil
+}
+
+func (lc *ServerLeaseCounter) CountServers() int {
+	leases, err := lc.lister.List(lc.selector)
```
avritt-techx, line 32: Could you add a comment here briefly describing the rational for not using a LIST and WATCH pattern (added complexity)?  It would help future readers of the code understand why we didn't do it initially

pkg/servercounter/counter.go
```diff
@@ -0,0 +1,14 @@
+package servercounter
```
avritt-techx, line 1: Please add package comments and doc comments for every public type/function: https://tip.golang.org/doc/comment

pkg/servercounter/lease_counter.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		klog.Errorf("could not list leases to update server count, using fallback count (%v): %v", lc.fallbackCount, err)
+		return lc.fallbackCount
```
avritt-techx, line 35: We need to consider the implications of the List failure.  In general we can't assume kube-apiserver will always send a successful response.  If we get a single List failure, should we immediately return the fallback value?  What behavior would that cause in the agent sync loop?

pkg/servercounter/lease_counter.go
```diff
@@ -0,0 +1,73 @@
...
+
+	if count != lc.fallbackCount {
+		klog.Infof("found %v valid leases, updating fallback count", count)
+		lc.fallbackCount = count
+	}
+
+	return count
+}
+
+func IsLeaseValid(lease *coordinationv1api.Lease) bool {
```
avritt-techx, line 60: should this be public?  Do we expect importers of this package to use this function?

pkg/servercounter/lease_counter.go
```diff
@@ -0,0 +1,73 @@
...
+	var lastRenewTime time.Time
+	if lease.Spec.RenewTime != nil {
+		lastRenewTime = lease.Spec.RenewTime.Time
+	} else if lease.Spec.AcquireTime != nil {
+		lastRenewTime = lease.Spec.AcquireTime.Time
+	} else {
+		klog.Warningf("lease %v has neither a renew time or an acquire time, marking as expired: %v", lease.Name, lease)
```
avritt-techx, line 67: should we just return false here?

pkg/servercounter/lease_counter.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	duration := time.Duration(*lease.Spec.LeaseDurationSeconds) * time.Second
```
avritt-techx, line 70: because time.Second is already a time.Duration, I think you can just to `*lease.Spec.LeaseDurationSeconds * time.Second`

pkg/servercounter/lease_counter.go
avritt-techx: Oh I didn't read the block before writing this comment

```
        if count != lc.fallbackCount {
		klog.Infof("found %v valid leases, updating fallback count", count)
		lc.fallbackCount = count
	}
```

 I think the behavior is fine. Could you add a comment here briefly explaining the agent behavior in the event of List failures however?

pkg/servercounter/lease_counter.go
carreter: Done.

pkg/servercounter/counter.go
carreter: Done!

pkg/servercounter/lease_counter.go
carreter: Added to the function doc comment.

pkg/servercounter/lease_counter.go
carreter: Thanks for catching that! Done.

pkg/servercounter/lease_counter.go
carreter: Funny quirk of Go: when you do `10 * time.Second`, the compiler infers that `10` is actually a `time.Duration`. You have to explicitly cast an `int` to a `time.Duration` if it's a variable.

pkg/servercounter/lease_counter.go
carreter: Made it private.

pkg/servercounter/lease_counter.go
carreter: Done.

pkg/servercounter/lease_counter_test.go
```diff
@@ -0,0 +1,249 @@
...
+			Labels: template.labels,
+		},
+		Spec: coordinationv1.LeaseSpec{},
+	}
+
+	if template.durationSecs != 0 {
+		lease.Spec.LeaseDurationSeconds = &template.durationSecs
+	}
+	if template.timeSinceAcquire != time.Duration(0) {
+		acquireTime := metav1.NewMicroTime(time.Now().Add(-template.timeSinceAcquire))
```
avrittrohwer, line 34: Reading how this is used in the tests, I don't think there is an issue using time.Now().  However if we were doing anything more complicated it would be better to:

1. Have a `var timeNow = time.Now` declaration in the package.  This variable is used in place of directly calling `time.Now()`.
2. Implement a `type controlledTime struct {}` in the test file that allows for manual manipulation of the flow of time.
3. In tests, override `timeNow` to be a `controlledTime`.

This would allow for determinism in unit tests.  Using time.Now() in unit tests is a common gotcha that causes flakeyness

pkg/servercounter/lease_counter_test.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+func (lister *fakeLeaseLister) List(selector labels.Selector) ([]*coordinationv1.Lease, error) {
+	lister.calls = append(lister.calls, selector)
+
+	if lister.err != nil {
+		return nil, lister.err
+	}
+
+	res := make([]*coordinationv1.Lease, 0)
```
avrittrohwer, line 130: nitpick: if not initializing to a capacity > 0, use `[]*coordinationv1.Lease{}`

pkg/servercounter/lease_counter_test.go
carreter: Ended up going with #1.

pkg/servercounter/lease_counter_test.go
carreter: Done!

pkg/servercounter/lease_counter_test.go
```diff
@@ -0,0 +1,249 @@
...
+				t.Fatalf("server counter creation failed: %v", err)
+			}
+
+			got := counter.CountServers()
+			if tc.want != got {
+				t.Errorf("incorrect server count (got: %v, want: %v)", got, tc.want)
+			}
+		})
+	}
+}
```
avrittrohwer, line 251: can you add a test that ensures the fallback value is updated on successful List, and the updated fallback count is used when the next List returns an error?

pkg/servercounter/lease_counter_test.go
avrittrohwer: sorry I could have explained better, these are all steps for a way to get around using time.Now(), not separate options.

Basically the issue with using time.Now() is that we don't control the test execution timing.  The key interaction is in lease_counter.go `return lastRenewTime.Add(duration).After(time.Now())`.  In the unit tests this is a possible (however unlikely) source of nondeterminism. 

Using the "freshly renewed lease is valid" case as an example with the current code:

1. timeNow is set to some fixed time.
2. newLeaseFromTemplate() is called, which creates a `Lease{ Spec { LeaseDurationSeconds: 1000, AcquireTime: timeNow - 1000s, RenewTime: timeNow - 1s }}`
3. (however unlikely) process running the test gets preempted for 1001 seconds.
4. isLeaseValid() is called.  `lastRenewTime` is set to `timeNow - 1s`.  `duration` is set to 1000. 
 `lastRenewTime.Add(duration).After(time.Now())` returns false.  The test fails.

We can control this nondeterminism by not directly using time.Now() in both lease_counter.go and lease_counter_test.go.  In lease_counter.go, we declare a `var timeNow` function that is initially set to `time.Now` (the function, not a value from calling the function).  We use the `timeNow` function alias to get the current time.  In non-test code, this is equivalent to directly using `time.Now` to get the current time.

In lease_counter_test.go, we add a `type controlledTime struct { t time.Time }` and implement methods `Now() time.Time` and probably something like `Advance(d time.Duration)`.  In each test we construct a controlledTime and set the package `timeNow` to be the `Now` function on the controlledTime (note that this means we can not run tests cases concurrently, if we did we would have multiple tests writing to `timeNow` which would cause nondeterminism).  In the test we would manually call `Advance()` to simulate the passage of time.  Looking back at our example, this would look something like:

1. `t := controlledTime{ time.Unix(1000, 0) }`
2. `timeNow = t.Now`
3. newLeaseFromTemplate() is called, which creates a `Lease{ Spec { LeaseDurationSeconds: 1000, AcquireTime: time.Unix(0, 0), RenewTime: time.Unix(999 0) }}`
4. t.Advance(time.Second)
5. (however unlikely) process running the test gets preempted for 1001 seconds.
6. isLeaseValid() is called.  `lastRenewTime` is set to `time.Unix(999, 0)`.  `duration` is set to 1000. 
 `lastRenewTime.Add(duration).After(timeNow() /* returns time.Unix(1001, 0) */)` returns true.  The test succeeds

pkg/servercounter/lease_counter_test.go
carreter: Done.

pkg/servercounter/lease_counter_test.go
```diff
@@ -0,0 +1,249 @@
...
+	"testing"
+	"time"
+
+	coordinationv1 "k8s.io/api/coordination/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	coordinationv1listers "k8s.io/client-go/listers/coordination/v1"
+)
+
+type leaseTemplate struct {
```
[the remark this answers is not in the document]
carreter: Thank you!!!

pkg/servercounter/lease_counter_test.go
carreter: Ohhhh, understood. Doing this now!

carreter:
Currently blocked by #639, which adds an e2e framework with which the lease-based counting system can be tested.

carreter:
Going to close this PR and open a new one as the commit history is incredibly messy.

PR #639, opened 2024-07-08
Bump go 1.22.2->1.22.5 and add E2E tests

carreter:
Made some E2E tests that:
- spin up a `kind` cluster
- deploy konnectivity servers and agents
- get metrics to confirm all servers are connected to all agents

Currently do this for two scenarios: 1 agent + 1 server and 3 agents + 3 servers.

Had to run `go mod tidy` and `go mod vendor` along the way to pull in some dependencies, couldn't figure out how to do this without bumping the go patch version. Hope this is ok, I can try and rework it if not!


carreter:
@avrittrohwer is on vacation this week, is anyone else available to review?

carreter:
Blocks #635 FYI

carreter:
@tallclair for visibility. Working on cleaning up the PR to see if we can avoid upgrading the go version and minimize the vendored dependencies!

carreter:
Seems like `sigs.k8s.io/e2e-fromework v0.4.0` requires go 1.22.3 so it is impossible to downgrade:

```
$ go mod tidy -go 1.22.2
go: sigs.k8s.io/e2e-framework@v0.4.0 requires go@1.22.3, but 1.22.2 is requested
```

Here are the new packages that are getting pulled into `vendor/`:

```
$ git diff master -- vendor/modules.txt | grep -e "+# " | sed -e "s/\+# //" -e "/.*=>.*/d"
github.com/evanphx/json-patch/v5 v5.9.0
github.com/gorilla/websocket v1.5.0
github.com/imdario/mergo v0.3.15
github.com/moby/spdystream v0.2.0
github.com/mxk/go-flowrate v0.0.0-20140419014527-cca7078d478f
github.com/vladimirvivien/gexe v0.2.0
sigs.k8s.io/controller-runtime v0.18.2
sigs.k8s.io/e2e-framework v0.4.0
sigs.k8s.io/yaml v1.4.0
```

@cheftako Where can I check if these are ok to include?

go.mod
```diff
@@ -1,7 +1,6 @@
 module sigs.k8s.io/apiserver-network-proxy
 
-go 1.22
-toolchain go1.22.4
```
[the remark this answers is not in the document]
carreter: `go mod tidy` did this automatically. Reverted.

Makefile
```diff
@@ -80,6 +80,10 @@ test:
 test-integration: build
 	go test -mod=vendor -race ./tests -agent-path $(PWD)/bin/proxy-agent
 
+.PHONY: test-e2e
+test-e2e: docker-build
+	go test -mod=vendor ./e2e -agent-image ${AGENT_FULL_IMAGE}-$(TARGETARCH):${TAG} -server-image ${SERVER_FULL_IMAGE}-$(TARGETARCH):${TAG}
```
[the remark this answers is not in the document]
carreter: Nope, just didn't know about it. Added the `-race` flag.

e2e/templates_test.go
```diff
@@ -0,0 +1,65 @@
...
+	Args     []KeyValue
+}
+
+var (
+	serverTemplates = template.Must(template.ParseGlob("templates/server/*"))
+	agentTemplates  = template.Must(template.ParseGlob("templates/agent/*"))
+)
+
+// renderServerTemplate renders a template from e2e/templates/server into a kubernetes object.
+func renderServerTemplate(file string, params any) (client.Object, *schema.GroupVersionKind, error) {
```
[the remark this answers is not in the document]
carreter: Sounds like this would be significantly easier to use. Will do!

e2e/templates_test.go
```diff
@@ -0,0 +1,65 @@
+package e2e
+
+import (
```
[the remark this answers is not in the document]
carreter: Done!

e2e/singleserver_singleagent_test.go
```diff
@@ -0,0 +1,138 @@
...
+func TestSingleServer_SingleAgent_StaticCount(t *testing.T) {
+	serverServiceHost := "konnectivity-server.kube-system.svc.cluster.local"
+	agentServiceHost := "konnectivity-agent.kube-system.svc.cluster.local"
+	adminPort := 8093
+
+	serverStatefulSetCfg := StatefulSetConfig{
+		Replicas: 1,
+		Image:    *serverImage,
+		Args: []KeyValue{
+			{Key: "log-file", Value: "/var/log/konnectivity-server.log"},
```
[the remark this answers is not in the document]
carreter: Done!

e2e/main_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	testenv = env.New()
+	kindClusterName := "kind-test"
+	kindCluster := kind.NewCluster(kindClusterName)
+
+	testenv.Setup(
+		envfuncs.CreateCluster(kindCluster, kindClusterName),
+		envfuncs.LoadImageToCluster(kindClusterName, *agentImage),
+		envfuncs.LoadImageToCluster(kindClusterName, *serverImage),
+		func(ctx context.Context, cfg *envconf.Config) (context.Context, error) {
```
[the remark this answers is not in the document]
carreter: Done!

e2e/templates/agent/statefulset.yaml
```diff
@@ -0,0 +1,82 @@
+apiVersion: apps/v1
+kind: StatefulSet
```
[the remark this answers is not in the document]
carreter: StatefulSets give the pods a unique identifier we can reach them with. 

```go
metricsFamilies, err := getMetrics(fmt.Sprintf("%v-%v:%v/metrics", agentPod.Name, agentServiceHost, adminPort))
```

e2e/main_test.go
```diff
@@ -0,0 +1,128 @@
...
+			return ctx, nil
+		},
+	)
+
+	testenv.Finish(envfuncs.DestroyCluster(kindClusterName))
+
+	os.Exit(testenv.Run(m))
+}
+
+func getMetrics(url string) (map[string]*io_prometheus_client.MetricFamily, error) {
```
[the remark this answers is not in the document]
carreter: Created a `getMetricsGaugeValue()` func

e2e/multiserver_multiagent_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+		var agentPods *corev1.PodList
+		err := client.Resources().List(ctx, agentPods, resources.WithLabelSelector("k8s-app=konnectivity-agent"))
+		if err != nil {
+			t.Fatalf("couldn't get agent pods: %v", err)
+		}
+
+		for _, agentPod := range agentPods.Items {
+
+			metricsFamilies, err := getMetrics(fmt.Sprintf("%v-%v:%v/metrics", agentPod.Name, agentServiceHost, adminPort))
```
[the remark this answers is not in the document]
carreter: Done!

carreter:
@avrittrohwer Addressed your comments. Working on including this in the GitHub workflows now.

carreter:
Excluded the e2e tests from the main test runners (fast-test and test) in the Makefile. Hopefully that fixes this.

avrittrohwer:
Hmm, should we expect to see the kind-e2e job in the checks?  I don't see it

carreter:
> Hmm, should we expect to see the kind-e2e job in the checks? I don't see it

I always forget how this works. Might not show up until it gets merged?

Makefile
```diff
@@ -67,19 +67,23 @@ mock_gen:
 # Unit tests with faster execution (nicer for development).
 .PHONY: fast-test
 fast-test:
-	go test -mod=vendor -race ./...
+	go test -mod=vendor -race $(shell go list ./... | grep -v e2e)
```
avrittrohwer, line 70: do you know if this would exclude any file with e2e in the name?  Could we make this more specific to files within the e2e/ directory

avrittrohwer:
> > Hmm, should we expect to see the kind-e2e job in the checks? I don't see it
> 
> I always forget how this works. Might not show up until it gets merged?

Could you merge this into your fork's main branch then open up a PR again your fork so the new action runs?

Makefile
carreter: It would, yes. Updated the pattern to match `/e2e$` and `/e2e/.*` instead.

carreter:
Seems like `kind-e2e` is included as part of the `e2e` workflow but needs maintainer approval before running:

image

There's some minor bugs in the workflow I'm ironing out, will ping when they're sorted!

carreter:
Workflow is working now! I will make it so it runs against the same k8s version matrix as the existing e2e tests.

carreter:
@avrittrohwer Github action for e2e tests is now working against multiple k8s versions. Unfortunately I can't seem to get `make docker-build` to load the image built by the build step, and it rebuilds it for each test run. Any clue how to fix this?

avrittrohwer:
@cheftako @jkh52 could either of you approve the new e2e action check so we can validate this PR?

carreter:
@avrittrohwer FYI, it's working on my fork of the repo: https://github.com/carreter/apiserver-network-proxy/actions/runs/10012999446

avrittrohwer:
> @avrittrohwer Github action for e2e tests is now working against multiple k8s versions. Unfortunately I can't seem to get `make docker-build` to load the image built by the build step, and it rebuilds it for each test run. Any clue how to fix this?

Is this because the test-e2e make target is reliant on docker-build target: `test-e2e: docker-build`?

e2e/main_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+	testenv.Setup(
+		envfuncs.CreateCluster(kindCluster, kindClusterName),
+		envfuncs.LoadImageToCluster(kindClusterName, *agentImage),
+		envfuncs.LoadImageToCluster(kindClusterName, *serverImage),
+		func(ctx context.Context, cfg *envconf.Config) (context.Context, error) {
+			client := cfg.Client()
+
+			// Render agent RBAC templates.
+			agentServiceAccount, _, err := renderAgentTemplate("serviceaccount.yaml", struct{}{})
```
ipochi: we should reuse kind templates in examples directory. That way we won't have to update templates at two different places. Also it would always mean that if e2e passes.

e2e/main_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/envfuncs"
+	"sigs.k8s.io/e2e-framework/support/kind"
+)
+
+var (
+	testenv     env.Environment
+	agentImage  = flag.String("agent-image", "", "The proxy agent's docker image.")
+	serverImage = flag.String("server-image", "", "The proxy server's docker image.")
+)
```
ipochi: not a blocking change but why not put it together in a config struct ?

e2e/singleserver_singleagent_test.go
```diff
@@ -16,10 +17,11 @@ import (
 )
 
 func TestSingleServer_SingleAgent_StaticCount(t *testing.T) {
-	serviceHost := "konnectivity-server.kube-system.svc.cluster.local"
+	serverServiceHost := "konnectivity-server.kube-system.svc.cluster.local"
+	agentServiceHost := "konnectivity-agent.kube-system.svc.cluster.local"
 	adminPort := 8093
 
-	serverDeploymentCfg := DeploymentConfig{
+	serverStatefulSetCfg := StatefulSetConfig{
```
ipochi, line 20: why change from deployment to statefulset ?

e2e/singleserver_singleagent_test.go
```diff
@@ -0,0 +1,130 @@
...
+			{Key: "uds-name", Value: "/etc/kubernetes/konnectivity-server/konnectivity-server.socket"},
+			{Key: "delete-existing-uds-file"},
+			{Key: "cluster-cert", Value: "/etc/kubernetes/pki/apiserver.crt"},
+			{Key: "cluster-key", Value: "/etc/kubernetes/pki/apiserver.key"},
+			{Key: "server-port", Value: "8090"},
+			{Key: "agent-port", Value: "8091"},
+			{Key: "health-port", Value: "8092"},
+			{Key: "admin-port", Value: strconv.Itoa(adminPort)},
+			{Key: "keepalive-time", Value: "1h"},
+			{Key: "mode", Value: "grpc"},
```
ipochi: we should also test e2e for http-connect mode.

Perhaps this can be retrieved as environment variable and the e2e github workflow run twice once with grpc and second time with http-connect  ?

similar to how you have done for k8s matrix

.github/workflows/e2e.yaml
```diff
@@ -62,24 +62,30 @@ jobs:
...
+      uses: actions/download-artifact@v4
       with:
         name: konnectivity-server
-    - uses: actions/download-artifact@v4
+    - name: Download prebuilt konnectivity-agent image
+      uses: actions/download-artifact@v4
       with:
         name: konnectivity-agent
-    - name: Install kind
+    - name: Load prebuilt konnectivity images
```
ipochi, line 86: can you please tell me what this does ? 

Does it not build a new konnectivity server/agent image per PR which is then used by the e2e tests ?

ipochi (changes requested):
Kudos for making the effort to add e2e tests. Left my comments.

I am generally not ok to add new set of templates, I think we should reuse the examples/kind templates the benefit would we two fold.

e2e/main_test.go
carreter: How do you mean? I did it this way so it could be passed in easily from the command line (e.g. by the `Makefile`).

e2e/singleserver_singleagent_test.go
carreter: StatefulSets allow for easier addressing of specific pods that belong to a service.

e2e/main_test.go
carreter: Good point. In that case, where should the logic for rendering the templates go?

ipochi: I see your point, we add a script in the examples/kind directory that renders  and creates the cluster for folks to try out from the documentation.

but in e2e test, we take the template as is and render and apply in code.

e2e/singleserver_singleagent_test.go
carreter: Added in the latest commit.

e2e/main_test.go
carreter: Exactly. Do you still want me to merge the two of them, or is it ok to keep them separate?

carreter:
Latest commit significantly speeds up the `kind-e2e` runner by not rebuilding the docker images on each iteration of the matrix.

e2e/main_test.go
ipochi: we can do that as a separate PR if this is causing blockers. Preference is to use the examples/kind.

carreter: Separate PR would be best as this blocks #635. I'll open an issue.

carreter:
@avrittrohwer @cheftako @ipochi @tallclair how are we feeling about this PR? If there are no major changes to be made, I'll start working on e2e tests for #635. Just want to be sure the e2e framework won't significantly change from under me before I do so!

e2e/multiserver_multiagent_test.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+func TestMultiServer_MultiAgent_StaticCount(t *testing.T) {
+	serverServiceHost := "konnectivity-server.kube-system.svc.cluster.local"
+	agentServiceHost := "konnectivity-agent.kube-system.svc.cluster.local"
+	adminPort := 8093
+	replicas := 3
+
+	serverStatefulSetCfg := StatefulSetConfig{
+		Replicas: 3,
```
avrittrohwer, line 22: nit: use `replicas` instead of `3` here and below in the agent config

avrittrohwer (approved):

e2e/multiserver_multiagent_test.go
carreter: Done!

ipochi:
> @avrittrohwer @cheftako @ipochi @tallclair how are we feeling about this PR? If there are no major changes to be made, I'll start working on e2e tests for #635. Just want to be sure the e2e framework won't significantly change from under me before I do so!

@carreter I'm good.

ipochi (approved):
Thanks for this effort !!

avrittrohwer:
https://github.com/kubernetes/test-infra/pull/33097 is merged

.github/workflows/e2e.yaml
cheftako: @aojea Could I get you to take a look at this?

carreter: I was able to fix this by making a new entry in the `Makefile` that doesn't have a dependency on `docker-build`.

e2e/singleserver_singleagent_test.go
cheftako: With Imran here. Agent should not need to maintain state across instances. So from a Kubernetes perspective it should be stateless and hence should run in a deployment not a statefulset. Worried that running it in statefulset in a test could mask a problem.

carreter: Hmm, do you know of a way to get metrics from a specific pod? That's the reason I was using a StatefulSet.

carreter:
As per @cheftako and @ipochi 's suggestion, I have changed the StatefulSet out for a Deployment and address the pods directly by their IP.

e2e/main_test.go
avrittrohwer, line 1: Could you add a short README.md to the e2e directory that goes over what maint_test.go sets up (RBAC and service) and what each specific test is expected to set up?

carreter: On it!

carreter: Done, could you take a look?

e2e/README.md
avrittrohwer, line 1: nit: I don't see 'KinD' capitalization used on the linked docs, keep it 'kind' for consistency

e2e/main_test.go
avrittrohwer: LGTM!

go.mod
```diff
@@ -21,7 +21,9 @@ require (
 	k8s.io/client-go v0.30.2
 	k8s.io/component-base v0.30.2
 	k8s.io/klog/v2 v2.130.0
-	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.0
+	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.29.0
```
jkh52, line 24: pleaser restore v0.0.0 to avoid confusion. Note there is a replace directive below.

carreter: Done!

e2e/metrics_assertions_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+func getMetricsGaugeValue(url string, name string) (int, error) {
+	resp, err := http.Get(url)
+	if err != nil {
+		return 0, fmt.Errorf("could not get metrics: %w", err)
+	}
+
+	metricsParser := &expfmt.TextParser{}
+	metricsFamilies, err := metricsParser.TextToMetricFamilies(resp.Body)
+	defer resp.Body.Close()
```
cheftako, line 24: Not critical but I think the defer should really be right after the http.Get error check.

e2e/metrics_assertions_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/prometheus/common/expfmt"
+	"sigs.k8s.io/e2e-framework/klient/k8s/resources"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+)
+
+func getMetricsGaugeValue(url string, name string) (int, error) {
+	resp, err := http.Get(url)
+	if err != nil {
+		return 0, fmt.Errorf("could not get metrics: %w", err)
```
cheftako, line 19: Debuggability is easier if you include the url in with the error.

e2e/metrics_assertions_test.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func assertServersAreConnected(expectedConnections int, adminPort int) func(context.Context, *testing.T, *envconf.Config) context.Context {
+	return func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+		client := cfg.Client()
+
+		var serverPods *corev1.PodList
+		err := client.Resources().List(ctx, serverPods, resources.WithLabelSelector("k8s-app=konnectivity-server"))
+		if err != nil {
+			t.Fatalf("couldn't get server pods: %v", err)
```
cheftako, line 69: Again debuggability goes up if you have a reference to relevant request (server pods) with the error.

e2e/metrics_assertions_test.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+
+		for _, serverPod := range serverPods.Items {
+			numConnections, err := getMetricsGaugeValue(fmt.Sprintf("%v:%v/metrics", serverPod.Status.PodIP, adminPort), "konnectivity_network_proxy_server_ready_backend_connections")
+			if err != nil {
+				t.Fatalf("couldn't get agent metric 'konnectivity_network_proxy_server_ready_backend_connections' for pod %v", serverPod.Name)
```
cheftako, line 75: nout found is the obvious error but there could be other reasons we got an error here and we did not check the  error type. I would suggest logging it.

e2e/metrics_assertions_test.go
```diff
@@ -0,0 +1,85 @@
...
+			}
+
+			if numConnections != expectedConnections {
+				t.Errorf("incorrect number of connected agents (want: %v, got: %v)", expectedConnections, numConnections)
```
cheftako, line 79: %v can mask some interesting errors in a case like this. They may be unequal because they are of different types but their toString method generates the same string. If you believe they should both be ints then it is better to use %d.

e2e/static_count_test.go
```diff
@@ -0,0 +1,129 @@
...
+			{"logtostderr", "true"},
+			{"log-file-max-size", "0"},
+			{"uds-name", "/etc/kubernetes/konnectivity-server/konnectivity-server.socket"},
+			{Key: "delete-existing-uds-file"},
+			{"cluster-cert", "/etc/kubernetes/pki/apiserver.crt"},
+			{"cluster-key", "/etc/kubernetes/pki/apiserver.key"},
+			{"server-port", "8090"},
+			{"agent-port", "8091"},
+			{"health-port", "8092"},
+			{"admin-port", strconv.Itoa(adminPort)},
```
cheftako, line 88: Not wrong but its not clear to me why your type converting what appears to be a constant?

e2e/metrics_assertions_test.go
carreter: Done.

e2e/metrics_assertions_test.go
carreter: Done.

e2e/metrics_assertions_test.go
carreter: Added a reference to the label selector being used to discover the pods.

e2e/metrics_assertions_test.go
carreter: Done.

e2e/metrics_assertions_test.go
carreter: Done.

e2e/static_count_test.go
carreter: Compiler complains that adminPort is an int if I don't conver it.

Issue #630, opened 2024-06-11
http-connect: incorrect error mesage from server to client because of hijack and error message written to http.Error or w.Write


ipochi:
Below is one of the example for the situation where incorrec error message is sent to the client from the server.

If we try to use curl or proxy-test-client to proxy a request without any agents connected to konnectivity-server, then the appropriate error message should be shown in [tunnel.go#L81](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/1ea67f9b4b9f67ea2dd2bf640ace4aa77165a4e0/pkg/server/tunnel.go#L81) i.e `http.Error(w, fmt.Sprintf("currently no tunnels available: %v", err), http.StatusInternalServerError)`

However, we never get that error printed on konnectivity-server, instead we get the below error message to the client.

```I0612 00:38:17.782494   27345 tunnel.go:42] "Received request for host" method="CONNECT" host="example.com:80" userAgent="curl/8.6.0"
I0612 00:38:17.782517   27345 tunnel.go:44] "TLS" commonName="proxy-client"
I0612 00:38:17.782595   27345 tunnel.go:78] Set pending(rand=6529093877251815469) to &{0x140001b2000 0x14000184200 {} 0x103002480 true false false false {{} 0} {0 0} <nil> {0x14000206540 map[] true true} map[] false 0 -1 200 false false [] {{} 0} [84 117 101 44 32 49 49 32 74 117 110 32 50 48 50 52 32 49 57 58 48 56 58 49 55 32 71 77 84] [0 0 0 0 0 0 0 0 0 0] [50 48 48] 0x140001750a0 {{} 0}}
2024/06/12 00:38:17 http: response.WriteHeader on hijacked connection from sigs.k8s.io/apiserver-network-proxy/pkg/server.(*Tunnel).ServeHTTP (tunnel.go:81)
2024/06/12 00:38:17 http: response.Write on hijacked connection from fmt.Fprintln (print.go:305)
```

Why ? because we hijack the connection in [#L58](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/1ea67f9b4b9f67ea2dd2bf640ace4aa77165a4e0/pkg/server/tunnel.go#L58). Since the connection is hijacked after the hijacker.Hijack() call, any attempts to use http.Error or w.Write to send an error message to the client will fail, resulting in the log message about the hijacked connection.

If, I make a simple change and write the error directly to `conn`, then I get the correct response back to the client

```
[...]

PR #640, opened 2024-07-16
http-connect: error to be written on connection

ipochi:
Incorrect error mesage from server to client because of hijack and error message being written to http.Error or w.Write, instead the error message should be written to the connection

fixes: #630


ipochi:
@cheftako @jkh52 Can this get merged ?

ipochi:
@cheftako @jkh52 friendly ping on this.

PR #643, opened 2024-07-26
Lease-based server counting logic for agent and lease controller for server

carreter:
Stacked PR on top of #639.

Implements a lease-based system that allows proxy agents to dynamically discover the number of proxy servers they should be connected to. This logic is only enabled when the `--server-lease-selector` flag is provided.

## New command-line flags for agent

- `--count-server-leases`: Enables the lease counting system. Disabled by default.
- `--kubeconfig`: Path to the kubeconfig used to create a client to watch for leases. Uses inClusterConfig by default.

## New command-line flags for server

-  `--enable-lease-controller`: Enables the controller that publishes, renews, and garbage collects leases. Disabled by default.


cmd/agent/app/options/options.go
```diff
@@ -80,6 +81,14 @@ type GrpcProxyAgentOptions struct {
 
 	SyncForever    bool
 	XfrChannelSize int
+
+	// Providing a label selector enables updating the server count by counting the
+	// number of valid leases matching the selector.
+	ServerLeaseSelector string
+	// Lease informer resync period.
+	InformerResync time.Duration
```
avrittrohwer, line 89: not used after your refactor

examples/kind-multinode/templates/k8s/konnectivity-agent-ds.yaml
```diff
@@ -53,7 +78,8 @@ spec:
           "--sync-forever",
           "--probe-interval=5s",
           "--service-account-token-path=/var/run/secrets/tokens/konnectivity-agent-token",
-          "--agent-identifiers=ipv4=${HOST_IP}"
+          "--agent-identifiers=ipv4=${HOST_IP}",
+          "--server-count-lease-selector=apiserver.kubernetes.io/identity=kube-apiserver",
```
avrittrohwer, line 82: If the lease controller logic lands before this PR is submitted, update the example to use the KNP leases

pkg/agent/lease_counter.go
```diff
@@ -0,0 +1,103 @@
...
+	start := timeNow()
+	defer func() {
+		latency := timeNow().Sub(start)
+		metrics.Metrics.ObserveLeaseListLatency(latency)
+	}()
+	leases, err := lc.leaseClient.List(ctx, metav1.ListOptions{LabelSelector: lc.selector.String()})
+	if err != nil {
+		klog.Errorf("could not list leases to update server count, using fallback count (%v): %v", lc.fallbackCount, err)
+
+		apiStatus, ok := err.(apierrors.APIStatus)
```
avrittrohwer, line 57: From the example: https://pkg.go.dev/errors#example-As

I think the more idiomatic usage would be:

```
var statusErr *apierrors.APIStatus
if errors.As(err, &statusErr) {
...
} else { 
...
}
```

pkg/agent/lease_counter.go
```diff
@@ -0,0 +1,103 @@
...
+
+	count := 0
+	for _, lease := range leases.Items {
+		if isLeaseValid(lease) {
+			count++
+		} else {
+		}
+	}
+
+	if count == 0 {
```
avrittrohwer, line 78: hm, what corner case is this covering?

examples/kind-multinode/templates/k8s/konnectivity-agent-ds.yaml
carreter: Sounds good!

cmd/agent/app/options/options.go
carreter: Removed.

pkg/agent/lease_counter.go
carreter: Fixed.

pkg/agent/lease_counter.go
carreter: My thought was that this could cover a corner case where something happens to the leasing system and all the leases disappear for some reason. Upon further reflection, I've removed this line - best to just return the actual count unless we encounter an error.

carreter:
This is ready for review now!

cmd/agent/app/options/options.go
```diff
@@ -122,6 +129,8 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.BoolVar(&o.WarnOnChannelLimit, "warn-on-channel-limit", o.WarnOnChannelLimit, "Turns on a warning if the system is going to push to a full channel. The check involves an unsafe read.")
 	flags.BoolVar(&o.SyncForever, "sync-forever", o.SyncForever, "If true, the agent continues syncing, in order to support server count changes.")
 	flags.IntVar(&o.XfrChannelSize, "xfr-channel-size", 150, "Set the size of the channel for transferring data between the agent and the proxy server.")
+	flags.StringVar(&o.ServerLeaseSelector, "server-lease-selector", o.ServerLeaseSelector, "Providing a label selector enables updating the server count by counting the number of valid leases matching the selector.")
```
cheftako, line 132: We should probably provide a warning here. Maybe it can be a reference to a section in the OSS docs. Reads should not require a webhook call from the KAS. However writing a lease can require an admission webhook call. This means that attempts to write the lease can fail because there is no tunnel from the KAS to cluster. There may be no tunnel to the cluster because the server count is 0. The server count is 0 because nothing can write a lease.

cheftako: Wondering if we should always assume a lower bound of 1.

cmd/agent/app/options/options.go
```diff
@@ -198,6 +207,17 @@ func (o *GrpcProxyAgentOptions) Validate() error {
 	if err := validateAgentIdentifiers(o.AgentIdentifiers); err != nil {
 		return fmt.Errorf("agent address is invalid: %v", err)
 	}
+	if o.ServerLeaseSelector != "" {
+		if _, err := labels.Parse(o.ServerLeaseSelector); err != nil {
+			return fmt.Errorf("invalid server count lease selector: %w", err)
+		}
```
cheftako, line 213: We should probably have a check that if o.ServerLeaseSelector != "" then o.KubeconfigPath != "".

cmd/agent/app/server.go
```diff
@@ -133,6 +135,31 @@ func (a *Agent) runProxyConnection(o *options.GrpcProxyAgentOptions, drainCh, st
...
+		var k8sClient *kubernetes.Clientset
+		config, err := clientcmd.BuildConfigFromFlags("", o.KubeconfigPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to load kubernetes client config: %v", err)
+		}
+		k8sClient, err = kubernetes.NewForConfig(config)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create kubernetes clientset: %v", err)
+		}
+		serverLeaseSelector, err := labels.Parse(o.ServerLeaseSelector)
```
cheftako, line 149: This should never happen. Maybe we should be caching the result of the parse we called during validation?

cmd/agent/app/server.go
```diff
@@ -133,6 +135,31 @@ func (a *Agent) runProxyConnection(o *options.GrpcProxyAgentOptions, drainCh, st
...
+			return nil, fmt.Errorf("invalid server count lease selector: %w", err)
+		}
+		serverLeaseCounter := agent.NewServerLeaseCounter(
+			k8sClient,
+			serverLeaseSelector,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create server lease counter: %w", err)
+		}
+		cc.ServerLeaseCounter = serverLeaseCounter
```
cheftako, line 160: lower bound to 1?

cmd/agent/app/server.go
```diff
@@ -133,6 +135,31 @@ func (a *Agent) runProxyConnection(o *options.GrpcProxyAgentOptions, drainCh, st
...
+		serverLeaseSelector, err := labels.Parse(o.ServerLeaseSelector)
+		if err != nil {
+			return nil, fmt.Errorf("invalid server count lease selector: %w", err)
+		}
+		serverLeaseCounter := agent.NewServerLeaseCounter(
+			k8sClient,
+			serverLeaseSelector,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create server lease counter: %w", err)
```
cheftako, line 158: We should make sure this gets logged. Worried this is opening an failure scenario where KAS cannot call into the cluster; wedging the cluster.

cmd/server/app/options/options.go
```diff
@@ -293,6 +304,25 @@ func (o *ProxyRunOptions) Validate() error {
...
 
+	// Validate leasing parameters. All must be empty or have a value.
+	if o.LeaseLabelSelector != "" || o.LeaseDuration != 0 || o.LeaseGCInterval != 0 || o.LeaseRenewalInterval != 0 || o.LeaseNamespace != "" {
+		if o.LeaseLabelSelector == "" {
+			return fmt.Errorf("LeaseLabelSelector cannot be empty when leasing system is enabled")
+		}
+		if o.LeaseNamespace == "" {
+			return fmt.Errorf("LeaseNamespace cannot be empty when leasing system is enabled")
+		}
+		if o.LeaseDuration == 0 {
```
cheftako, line 315: Are we ok with it being negative?

cmd/server/app/options/options.go
```diff
@@ -293,6 +304,25 @@ func (o *ProxyRunOptions) Validate() error {
...
+			return fmt.Errorf("LeaseDuration cannot be zero when leasing system is enabled")
+		}
+		if o.LeaseGCInterval == 0 {
```
cheftako, line 318: Ditto

cmd/server/app/options/options.go
```diff
@@ -293,6 +304,25 @@ func (o *ProxyRunOptions) Validate() error {
...
+			return fmt.Errorf("LeaseGCInterval cannot be zero when leasing system is enabled")
+		}
+		if o.LeaseRenewalInterval == 0 {
```
cheftako, line 321: Ditto

cmd/server/app/server.go
```diff
@@ -142,6 +143,16 @@ func (p *Proxy) Run(o *options.ProxyRunOptions, stopCh <-chan struct{}) error {
 		defer frontendStop()
 	}
 
+	if o.LeaseLabelSelector != "" || o.LeaseDuration != 0 || o.LeaseGCInterval != 0 || o.LeaseRenewalInterval != 0 {
```
cheftako, line 146: This seems redundant after the validation check. Not wrong but we can probably just use o.LeaseLabelSelector != "". Better yet you can add a helper function isLeaseCountingEnabled() which implements this and makes the code more readable. (Especially given that your missing o.LeaseNamespace != "" which suggests the helper would help keep the checks consistent)

cmd/server/app/server.go
```diff
@@ -142,6 +143,16 @@ func (p *Proxy) Run(o *options.ProxyRunOptions, stopCh <-chan struct{}) error {
...
+		leaseLabels, err := labels.ConvertSelectorToLabelsMap(o.LeaseLabelSelector)
+		if err != nil {
+			return fmt.Errorf("could not parse lease label selector: %w", err)
```
cheftako, line 149: Having the unparseable string (using %q) in the error message makes the oncalls life better.

cmd/agent/app/options/options.go
carreter: Current behavior pretty much assumes a lower bound of 1, as a connection is always attempted at first regardless of the server count and the server count is only taken in to consideration if it is != 0:

https://github.com/carreter/apiserver-network-proxy/blob/server-counter-e2e/pkg/agent/clientset.go#L191-L217

cmd/agent/app/options/options.go
carreter: Done.

cmd/agent/app/server.go
carreter: Could we just swallow the error instead? Would require less passing state around.

cmd/agent/app/server.go
carreter: Added a klog expression.

cmd/server/app/options/options.go
carreter: Probably not. Switched the expressions to be `<= 0`

cmd/server/app/server.go
carreter: Added helper method!

cmd/server/app/server.go
carreter: Done!

cmd/agent/app/options/options.go
tallclair: I disagree. If you pass an empty kubeconfig to `clientcmd.BuildConfigFromFlags` then it defaults to the in-cluster config, which is probably what most users want.

pkg/agent/lease_counter.go
```diff
@@ -0,0 +1,81 @@
...
+	leaseClient   coordinationv1.LeaseInterface
+	selector      labels.Selector
+	fallbackCount int
+}
+
+// NewServerLeaseCounter creates a server counter that counts valid leases that match the label
+// selector and provides the fallback count (initially 0) if this fails.
+func NewServerLeaseCounter(k8sClient kubernetes.Interface, labelSelector labels.Selector) *ServerLeaseCounter {
+	return &ServerLeaseCounter{
+		leaseClient:   k8sClient.CoordinationV1().Leases(""),
```
tallclair, line 31: What does an empty namespace do here? If this is just using the agent namespace, can we remove the leases-namespace flag?

cmd/server/app/options/options.go
```diff
@@ -138,7 +145,11 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
...
 	flags.StringSliceVar(&o.CipherSuites, "cipher-suites", o.CipherSuites, "The comma separated list of allowed cipher suites. Has no effect on TLS1.3. Empty means allow default list.")
 	flags.IntVar(&o.XfrChannelSize, "xfr-channel-size", o.XfrChannelSize, "The size of the two KNP server channels used in server for transferring data. One channel is for data coming from the Kubernetes API Server, and the other one is for data coming from the KNP agent.")
-
+	flags.DurationVar(&o.LeaseDuration, "lease-duration", o.LeaseDuration, "The duration of the KNP server lease. Lease system off by default")
+	flags.DurationVar(&o.LeaseRenewalInterval, "lease-renewal-interval", o.LeaseRenewalInterval, "The interval between KNP server lease renewal calls. Lease system off by default")
+	flags.DurationVar(&o.LeaseGCInterval, "lease-gc-interval", o.LeaseGCInterval, "The interval between KNP server garbage collection calls. Lease system off by default")
+	flags.StringVar(&o.LeaseLabelSelector, "lease-label-selector", o.LeaseLabelSelector, "The label selector for KNP server leases. Lease system off by default")
```
tallclair, line 151: This is just the set of labels to add, right? I think drop the word `selector`

tallclair: Optionally, you can then use the configuration map option, like so: https://github.com/kubernetes/kubernetes/blob/master/cmd/kubelet/app/options/options.go#L305-L306

pkg/server/leases/controller.go
```diff
@@ -0,0 +1,49 @@
+package leases
```
tallclair, line 1: nit: missing license header on all new files

pkg/server/leases/controller.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func NewController(k8sClient kubernetes.Interface, holderIdentity string, leaseDurationSeconds int32, renewInterval time.Duration, gcCheckPeriod time.Duration, leaseName, leaseNamespace string, leaseLabels map[string]string) *Controller {
+	acquireController := lease.NewController(
+		clock.RealClock{},
+		k8sClient,
+		holderIdentity,
+		leaseDurationSeconds,
+		func() {
+			klog.Errorf("repeat heartbeat error in lease acquisition controller for lease %v", leaseName)
```
tallclair, line 27: nit: think this already logs on every failure, no need to log here (can just provide nil for the func)

pkg/agent/lease_counter.go
```diff
@@ -0,0 +1,81 @@
...
+
+// Count counts the number of leases in the apiserver matching the provided
+// label selector.
+//
+// In the event that no valid leases are found or lease listing fails, the
+// fallback count is returned. This fallback count is updated upon successful
+// discovery of valid leases.
+func (lc *ServerLeaseCounter) Count(ctx context.Context) int {
+	// Since the number of proxy servers is generally small (1-10), we opted against
+	// using a LIST and WATCH pattern and instead list all leases on each call.
```
tallclair, line 45: I think it's still worth considering using an informer here. It simplifies the caching, and if you plumb through the events could let the agents be more responsive to server changes.

tallclair (commented):
Didn't get through the full review. Will pick back up tomorrow.

pkg/util/leases.go
```diff
@@ -0,0 +1,59 @@
...
+	} else {
+		klog.Warningf("lease %v has neither a renew time or an acquire time, marking as expired: %v", lease.Name, lease)
+		return false
+	}
+
+	duration := time.Duration(*lease.Spec.LeaseDurationSeconds) * time.Second
+
+	return lastRenewTime.Add(duration).After(timeNow()) // renewTime+duration > time.Now()
+}
+func NewLeaseFromTemplate(template LeaseTemplate) *coordinationv1api.Lease {
```
tallclair, line 29: Looks like this is only used for testing? If so, move it to the `testing` package: https://github.com/kubernetes-sigs/apiserver-network-proxy/tree/master/pkg/testing

pkg/agent/clientset.go
```diff
@@ -179,16 +184,17 @@ func (cs *ClientSet) resetBackoff() *wait.Backoff {
 }
 
 // sync makes sure that #clients >= #proxy servers
-func (cs *ClientSet) sync() {
+func (cs *ClientSet) sync(ctx context.Context) {
```
tallclair, line 187: The lease system lets us make this function a lot smarter, but the changes here don't really take advantage of it. Are you planning to follow up with those changes?

Specifically, I'm thinking of:
1. When a new lease is added, agents should try to connect to it. This is related to the informers comment elsewhere. This also means we don't need to continuously poll any more.
2. clientset should ensure that there is a server connection for each valid lease. This could avoid some edge cases where the agent has the right number of connections, but isn't connected to the right servers. (I'm not sure how this could happen in practice, but maybe if the loadbalancer & KCP instances get out of sync).

pkg/agent/clientset.go
```diff
@@ -179,16 +184,17 @@ func (cs *ClientSet) resetBackoff() *wait.Backoff {
...
 	backoff := cs.resetBackoff()
 	var duration time.Duration
 	for {
-		if err := cs.connectOnce(); err != nil {
+		if err := cs.connectOnce(ctx); err != nil {
 			if dse, ok := err.(*DuplicateServerError); ok {
 				clientsCount := cs.ClientsCount()
-				klog.V(4).InfoS("duplicate server", "serverID", dse.ServerID, "serverCount", cs.serverCount, "clientsCount", clientsCount)
-				if cs.serverCount != 0 && clientsCount >= cs.serverCount {
+				serverCount := cs.ServerCount(ctx)
```
tallclair, line 195: connectOnce already counted the servers. Is there some way to plumb that value through to here? This can also be avoided with the suggestions above.

pkg/agent/metrics/metrics.go
```diff
@@ -97,6 +101,15 @@ func newAgentMetrics() *AgentMetrics {
 		},
 		[]string{},
 	)
+	serverCount := prometheus.NewGaugeVec(
```
tallclair, line 104: I don't think there's any reason to use a vector metric if there aren't any labels? I could be wrong, but working with vector metrics is more expensive.

pkg/agent/lease_counter.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	metrics.Metrics.ObserveLeaseList(200, "")
+
+	count := 0
+	for _, lease := range leases.Items {
+		if util.IsLeaseValid(lease) {
+			count++
+		} else {
+		}
```
tallclair, line 73:
```suggestion
		}
```

cmd/agent/app/server.go
```diff
@@ -133,6 +135,32 @@ func (a *Agent) runProxyConnection(o *options.GrpcProxyAgentOptions, drainCh, st
...
+		}
+		serverLeaseSelector, err := labels.Parse(o.ServerLeaseSelector)
+		if err != nil {
+			return nil, fmt.Errorf("invalid server count lease selector: %w", err)
+		}
+		serverLeaseCounter := agent.NewServerLeaseCounter(
+			k8sClient,
+			serverLeaseSelector,
+		)
+		if err != nil {
```
tallclair, line 157: This is the error from the lease selector. I think you can just remove this block?

cmd/agent/app/server.go
tallclair: There's already special handling to ensure the agent keeps trying to connect when the server count is 0, so I don't think bounding to 1 is needed.

tallclair (commented):
I didn't have a chance to look at the tests.

cmd/agent/app/options/options.go
cheftako: Good point

cmd/agent/app/server.go
cheftako: I don't agree. If we exit runProxyConnection because of a lease error, you wedge the cluster. Happy to chat more but I think if we're not careful feel are going to do some real damage to our reliability here.

cmd/server/app/server.go
```diff
@@ -142,6 +143,16 @@ func (p *Proxy) Run(o *options.ProxyRunOptions, stopCh <-chan struct{}) error {
...
 	}
 
+	if o.IsLeaseCountingEnabled() {
+		leaseLabels, err := labels.ConvertSelectorToLabelsMap(o.LeaseLabelSelector)
+		if err != nil {
+			return fmt.Errorf("could not parse lease label selector (%q): %w", o.LeaseLabelSelector, err)
+		}
+		leaseController := leases.NewController(k8sClient, o.ServerID, int32(o.LeaseDuration.Seconds()), o.LeaseRenewalInterval, o.LeaseGCInterval, fmt.Sprintf("konnectivity-proxy-server-%v", o.ServerID), o.LeaseNamespace, leaseLabels)
+		klog.V(1).Infoln("Starting lease acquisition and garbage collection controller.")
+		leaseController.Run(ctx)
```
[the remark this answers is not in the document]
cheftako: +1. We also need to use a timeout on the lease to ensure that a server which went down to quickly to cleanup doesn't stick around in the count forever. However it would be good to notice faster than that.

pkg/agent/clientset.go
```diff
@@ -36,9 +36,12 @@ type ClientSet struct {
 	clients map[string]*Client // map between serverID and the client
 	// connects to this server.
 
-	agentID     string // ID of this agent
-	address     string // proxy server address. Assuming HA proxy server
-	serverCount int    // number of proxy server instances, should be 1
+	agentID string // ID of this agent
+	address string // proxy server address. Assuming HA proxy server
+
+	serverCounter           *ServerLeaseCounter // counts number of proxy server leases
```
cheftako, line 42: I think this is misleading. I don't believe this is actually a ServerLeaseCounter. It seems like a ServerCounterManager. One of the options in the ServerCounterManager is lease but thats just 1 of the options.

cheftako: Or maybe it should be misleading. I think it would be better to isolate the logic about server counts in 1 place rather than having various conditionals littering the server count throughout.

pkg/agent/clientset.go
```diff
@@ -211,20 +217,32 @@ func (cs *ClientSet) sync() {
...
 	if err != nil {
 		return err
 	}
-	if cs.serverCount != 0 && cs.serverCount != serverCount {
-		klog.V(2).InfoS("Server count change suggestion by server",
-			"current", cs.serverCount, "serverID", c.serverID, "actual", serverCount)
-
-	}
-	cs.serverCount = serverCount
+	cs.lastReceivedServerCount = receivedServerCount
```
cheftako, line 245: We should not be losing the server count changed warning.

pkg/agent/clientset.go
cheftako: Alternately the variable should be named leaseCounter if we don't want to restructure

pkg/agent/clientset.go
```diff
@@ -211,20 +217,32 @@ func (cs *ClientSet) sync() {
...
+	}
+
+	metrics.Metrics.SetServerCount(serverCount)
+	return serverCount
+}
+
+func (cs *ClientSet) connectOnce(ctx context.Context) error {
+	serverCount := cs.ServerCount(ctx)
+
+	if !cs.syncForever && serverCount != 0 && cs.ClientsCount() >= serverCount {
```
cheftako, line 238: ServerCount can currently return 0. If that happens consistently across agents we can wedge the cluster.

tallclair: I thought that's what the `serverCount != 0` here is trying to prevent? It's saying that even if `ClientsCount() == serverCount` then if the `serverCount == 0` you should keep trying to connect anyway.

cmd/agent/app/options/options.go
carreter: Reverted!

cmd/agent/app/server.go
carreter: Lease counter now always returns at least 1.

cmd/agent/app/options/options.go
```diff
@@ -122,6 +129,8 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.BoolVar(&o.WarnOnChannelLimit, "warn-on-channel-limit", o.WarnOnChannelLimit, "Turns on a warning if the system is going to push to a full channel. The check involves an unsafe read.")
 	flags.BoolVar(&o.SyncForever, "sync-forever", o.SyncForever, "If true, the agent continues syncing, in order to support server count changes.")
 	flags.IntVar(&o.XfrChannelSize, "xfr-channel-size", 150, "Set the size of the channel for transferring data between the agent and the proxy server.")
+	flags.StringVar(&o.ServerLeaseSelector, "server-lease-selector", o.ServerLeaseSelector, "Providing a label selector enables updating the server count by counting the number of valid leases matching the selector.")
+	flags.StringVar(&o.KubeconfigPath, "kubeconfig", o.KubeconfigPath, "absolute path to the kubeconfig file")
```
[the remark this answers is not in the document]
carreter: Capitalized and removed the bit about it being absolute, I don't think it is necessary.

cmd/server/app/options/options.go
```diff
@@ -138,7 +145,11 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
...
+	flags.StringVar(&o.LeaseNamespace, "lease-namespace", o.LeaseNamespace, "The namespace to which KNP server leases will be published. Lease system off by default")
```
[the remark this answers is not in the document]
carreter: @tallclair @cheftako Do you have a sense of what good default values for these would be? I was thinking:
- lease duration: 30 secs
- renewal interval: 15 secs
- GC interval: 15 secs
- label selector: k8s-app=konnectivity-server
- namespace: kube-system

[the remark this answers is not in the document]
carreter: De-duping is probably the way to go.

pkg/agent/lease_counter.go
carreter: set it to always be kube-system.

cmd/server/app/options/options.go
carreter: Will likely drop this entirely in favor of sensible default values.

pkg/server/leases/controller.go
carreter: Done.

pkg/agent/lease_counter.go
carreter: Added a comment about this.

pkg/server/leases/gc_controller.go
```diff
@@ -0,0 +1,57 @@
...
+	for _, lease := range leases.Items {
+		if util.IsLeaseValid(lease) {
+			continue
+		}
+
+		// Optimistic concurrency: if a lease has a different resourceVersion than
+		// when we got it, it may have been renewed.
+		err := c.leaseInterface.Delete(ctx, lease.Name, *metav1.NewRVDeletionPrecondition(lease.ResourceVersion))
+		if errors.IsNotFound(err) {
+			klog.Infof("lease %v was already deleted", lease.Name)
```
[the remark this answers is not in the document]
carreter: Done.

pkg/server/leases/gc_controller.go
```diff
@@ -0,0 +1,57 @@
...
+		} else if err != nil {
+			klog.Errorf("could not delete lease %v: %v", lease.Name, err)
```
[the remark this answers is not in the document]
carreter: Done!

pkg/server/leases/gc_controller.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+}
+
+func (c *GarbageCollectionController) Run(ctx context.Context) {
+	wait.Until(c.gc, c.gcCheckPeriod, ctx.Done())
+}
+
+func (c *GarbageCollectionController) gc() {
+	ctx := context.Background()
+	leases, err := c.leaseInterface.List(ctx, metav1.ListOptions{LabelSelector: c.labelSelector})
```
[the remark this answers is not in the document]
carreter: Done!

pkg/server/server.go
```diff
@@ -37,10 +37,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
-
```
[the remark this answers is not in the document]
carreter: Done.

pkg/util/leases.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"time"
+
+	"github.com/google/uuid"
+	coordinationv1api "k8s.io/api/coordination/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
+)
+
+var timeNow = time.Now
```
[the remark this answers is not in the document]
carreter: Will use that next time!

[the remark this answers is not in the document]
carreter: Actually, might be best to use this now as it's a bit of a headache to set up the controlled time so many times.

cmd/agent/app/server.go
carreter: Oops, good catch.

pkg/agent/clientset.go
carreter: Changed to `leaseCounter`.

pkg/agent/clientset.go
carreter: Lease count always returns at least 1 now.

pkg/agent/clientset.go
carreter: Added it back in!

pkg/agent/metrics/metrics.go
carreter: I was just copying the previously existing metrics, which are all GaugeVecs without labels. What should I be using instead?

pkg/agent/clientset.go
carreter: Done.

pkg/util/leases.go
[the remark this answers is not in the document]
carreter: Wired in the clock utility you suggested - thank you for the FYI!

cmd/server/app/server.go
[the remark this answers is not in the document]
carreter: Added this! Let me know if what I wrote works.

pkg/agent/lease_counter.go
konryd: I see that one of the previous versions of this PR did use informer. Did you have trouble getting it to work? (from a quick look - is it possible that that version was missing a call to Start()? )

https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/643/commits/7a6b3f9d13fcdfe14de674742d5ffec42a7bf29e

I think Informer is preferrable even with a Lister-based approach (so: the logic looks like periodic listing, but we're hitting the cache instead of apiserver).

carreter: @avrittrohwer asked me to switch it over to a direct list call in order to make observability into the success rate of the calls easier. I couldn't figure out a good way to hook up the Informer to publish metrics. It seems like people would generally prefer the Informer method, I can do this if it is desired!

cmd/server/app/server.go
```diff
@@ -42,17 +42,23 @@ import (
...
 	"sigs.k8s.io/apiserver-network-proxy/pkg/util"
 	"sigs.k8s.io/apiserver-network-proxy/proto/agent"
 )
 
 var udsListenerLock sync.Mutex
 
-const ReadHeaderTimeout = 60 * time.Second
+const (
+	ReadHeaderTimeout    = 60 * time.Second
+	LeaseDuration        = 30 * time.Second
```
cheftako, line 57: GKE uses num agents ~= log(num nodes). So for 10K node clusters I don't expect a problem on GKE. However the ANP sample manifests uses a daemonset so num agents == num nodes. 10K agents renewing leases every 30 seconds would be an additional ~150 QPS of writes against the KAS. At some point we may want to A) use an alternate to daemonsets in our examples B) document the need for an alternate to dameonsets in our docs C) provide a configuration for the lease duration. Not saying it needs to be part of this PR but one of these should be a follow up AI.

cheftako: NM just realized this is server code not agent code. Deleting.

pkg/agent/clientset.go
```diff
@@ -211,23 +217,40 @@ func (cs *ClientSet) sync() {
...
+
+	if serverCount != cs.lastServerCount {
+		klog.Warningf("change detected in proxy server count (was: %d, now: %d)", cs.lastServerCount, serverCount)
+		cs.lastServerCount = serverCount
 	}
-	if cs.serverCount != 0 && cs.serverCount != serverCount {
-		klog.V(2).InfoS("Server count change suggestion by server",
-			"current", cs.serverCount, "serverID", c.serverID, "actual", serverCount)
 
+	metrics.Metrics.SetServerCount(serverCount)
```
cheftako, line 236: Ideally we should record the metrics where get obtain them. The primary reason I say this is that for there are no leases case this will record a metric of 1 not 0.

pkg/agent/clientset.go
cheftako: I can't seem to find it.

pkg/agent/metrics/metrics.go
```diff
@@ -136,6 +174,9 @@ func (a *AgentMetrics) Reset() {
 	a.endpointConnections.Reset()
 	a.streamPackets.Reset()
 	a.streamErrors.Reset()
+	a.serverCount.Reset()
```
cheftako, line 177: Accumulating values clearly need to be reset. Less clear to me that polled values (like serverCount) make sense to be reset. Reading the value after resetting and getting the reset value doesn't mean that new servers haven't accumulated, it just means you haven't polled since the reset.

pkg/server/leases/controller.go
```diff
@@ -0,0 +1,81 @@
...
+	coordinationv1 "k8s.io/api/coordination/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/component-helpers/apimachinery/lease"
+	"k8s.io/klog/v2"
+	"k8s.io/utils/clock"
+)
+
+const shutdownDeleteGrace = 10
```
cheftako, line 31: pkg/server/leases/controller.go:31:7: const `shutdownDeleteGrace` is unused (unused)
const shutdownDeleteGrace = 10

e2e/lease_count_test.go
```diff
@@ -0,0 +1,327 @@
...
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+type leaseController struct {
+	labels        map[string]string
+	validLeases   []*coordinationv1.Lease
+	expiredLeases []*coordinationv1.Lease
+}
+
+func NewLeaseController(labels map[string]string) *leaseController {
```
cheftako, line 22: e2e/lease_count_test.go:22:51: unexported-return: exported func NewLeaseController returns unexported type *e2e.leaseController, which can be annoying to use (revive)
func NewLeaseController(labels map[string]string) *leaseController {
                                                  ^
                                                  
Is there a reason NewLeaseController needs to be public? If not its probably easier to just make it not public.

pkg/agent/metrics/metrics.go
avrittrohwer: You can use promethus.NewGauge() https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#NewGauge

pkg/agent/metrics/metrics.go
avrittrohwer, line 1: I think it would be good to have new metrics written by the server around the lease calls that it makes

carreter:
/retest

Still need to address review comments, will do that by tomorrow morning.

e2e/static_count_test.go
```diff
@@ -1,129 +1,104 @@
...
-			{"uds-name", "/etc/kubernetes/konnectivity-server/konnectivity-server.socket"},
-			{Key: "delete-existing-uds-file"},
-			{"cluster-cert", "/etc/kubernetes/pki/apiserver.crt"},
-			{"cluster-key", "/etc/kubernetes/pki/apiserver.key"},
-			{"server-port", "8090"},
-			{"agent-port", "8091"},
-			{"health-port", "8092"},
-			{"admin-port", strconv.Itoa(adminPort)},
-			{"keepalive-time", "1h"},
-			{"mode", *connectionMode},
```
cheftako, line 90: e2e/main_test.go:38:2: var `connectionMode` is unused (unused)
	connectionMode = flag.String("mode", "grpc", "Connection mode to use during e2e tests.")

pkg/agent/clientset.go
carreter: Will not have time to do this in this PR, but I will open issues for these two things!

pkg/agent/clientset.go
carreter: Hmm, don't we want the metric to report what the agent *thinks* the server count is, though? The agent will never believe the count is 0.

Happy to add a separate metric for the lease counting system!

pkg/agent/clientset.go
carreter: https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/643/files#diff-f79c206ef287efc120666a024e5a23b09183ad983123a5cbd3fde793b261d0f8R228-R231

pkg/agent/lease_counter.go
carreter: Will switch over to an Informer and hook in metrics in the Informer per some suggestions @avrittrohwer made!

pkg/agent/metrics/metrics.go
carreter: Gotcha, I just assumed that all the metrics should be reset when this method is called.

e2e/main_test.go
```diff
@@ -156,21 +156,172 @@ func renderAndApplyManifests(ctx context.Context, cfg *envconf.Config) (context.
...
+				t.Errorf("logs for pod %q: %v", pod.Name, logs)
+			}
+			t.Fatalf("waiting for deletion of pods for deployment %q failed, dumped pod logs", deployment.Name)
+		}
+
+		return ctx
+	}
+}
+
+func sleepFor(duration time.Duration) func(context.Context, *testing.T, *envconf.Config) context.Context {
```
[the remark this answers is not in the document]
carreter: Removed.

examples/kind-multinode/templates/k8s/konnectivity-agent-ds.yaml
carreter: Done.

pkg/agent/lease_counter.go
carreter: Done.

pkg/agent/lease_counter.go
carreter: Put this back in to ensure that the agent always thinks there is at least 1 server.

pkg/agent/metrics/metrics.go
carreter: Done.

pkg/agent/metrics/metrics.go
carreter: Done.

pkg/server/leases/controller.go
carreter: Should be done now. Let me know if I missed any.

pkg/server/leases/controller.go
carreter: Removed.

carreter:
I believe I addressed all the review comments. Let me know if I missed anything.

pkg/agent/clientset.go
```diff
@@ -237,15 +257,15 @@ func (cs *ClientSet) connectOnce() error {
...
+	return serverCount, nil
 }
 
 func (cs *ClientSet) Serve() {
 	labels := runpprof.Labels(
 		"agentIdentifiers", cs.agentIdentifiers,
 		"serverAddress", cs.address,
 	)
-	go runpprof.Do(context.Background(), labels, func(context.Context) { cs.sync() })
+	go runpprof.Do(context.Background(), labels, func(ctx context.Context) { cs.sync() })
```
avrittrohwer, line 268: ctx not used

pkg/agent/clientset.go
avrittrohwer: Should we remove the `serverCount != 0` check then?

pkg/agent/lease_counter.go
```diff
@@ -0,0 +1,147 @@
...
+
+// Count counts the number of leases in the apiserver matching the provided
+// label selector.
+//
+// In the event that no valid leases are found or lease listing fails, the
+// fallback count is returned. This fallback count is updated upon successful
+// discovery of valid leases.
+func (lc *ServerLeaseCounter) Count() int {
+	// Since the number of proxy servers is generally small (1-10), we opted against
+	// using a LIST and WATCH pattern and instead list all leases on each call.
```
avrittrohwer, line 67: We're still using an informer under the hood so this comment is misleading

pkg/agent/lease_counter.go
```diff
@@ -0,0 +1,147 @@
...
+	// TODO: Switch to an informer-based system and use events rather than a polling loop.
```
avrittrohwer, line 68: don't need the metric writes here anymore

pkg/agent/lease_counter.go
```diff
@@ -0,0 +1,147 @@
...
+		lc.fallbackCount = count
+	}
+
+	return count
+}
+
+func NewLeaseInformerWithMetrics(client kubernetes.Interface, namespace string, resyncPeriod time.Duration) cache.SharedIndexInformer {
+	return cache.NewSharedIndexInformer(
+		&cache.ListWatch{
+			ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
```
avrittrohwer, line 106: write lease latencies metric here

pkg/server/leases/controller.go
```diff
@@ -0,0 +1,79 @@
...
+		gcController:      gcController,
+		acquireController: acquireController,
+	}
+}
+
+// Run starts the garbage collection and lease acquisition controllers. Non-blocking.
+func (c *Controller) Run(ctx context.Context) {
+	go c.gcController.Run(ctx)
+	go c.acquireController.Run(ctx)
+
```
avrittrohwer, line 70: nit: blank line

cmd/agent/app/server.go
```diff
@@ -133,6 +142,30 @@ func (a *Agent) runProxyConnection(o *options.GrpcProxyAgentOptions, drainCh, st
...
+		}
+		k8sClient, err = kubernetes.NewForConfig(config)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create kubernetes clientset: %v", err)
+		}
+		leaseInformer := agent.NewLeaseInformerWithMetrics(k8sClient, LeaseNamespace, LeaseInformerResync)
+		go leaseInformer.Run(stopCh)
+		cache.WaitForCacheSync(stopCh, leaseInformer.HasSynced)
+		leaseLister := coordinationv1lister.NewLeaseLister(leaseInformer.GetIndexer())
+		serverLeaseSelector, _ := labels.Parse("k8s-app=konnectivity-server")
```
ipochi, line 160: Question: Is this label meant to be hardcoded ? If I remember correctly the idea was to pass the labels as cli arguments ?

cmd/server/app/server.go
```diff
@@ -149,6 +155,22 @@ func (p *Proxy) Run(o *options.ProxyRunOptions, stopCh <-chan struct{}) error {
...
+	if o.EnableLeaseController {
+		leaseController := leases.NewController(
+			k8sClient,
+			o.ServerID,
+			int32(LeaseDuration.Seconds()),
+			LeaseRenewalInterval,
+			LeaseGCInterval,
+			fmt.Sprintf("konnectivity-proxy-server-%v", o.ServerID),
+			LeaseNamespace,
+			map[string]string{"k8s-app": "konnectivity-server"},
```
ipochi, line 167: same question here. Perhaps the label is passed to as cli flags ? of course non-blocking as we cna modify the behaviour in a separate PR if it means getting this merged sooner.

pkg/agent/lease_counter.go
ipochi: Perhaps a question for the meeting but I'm having a hard time following the thoughts on why must an agent think there is at least 1 server.

cmd/agent/app/server.go
```diff
@@ -38,14 +38,23 @@ import (
...
+	"k8s.io/utils/clock"
 	"sigs.k8s.io/apiserver-network-proxy/cmd/agent/app/options"
 	"sigs.k8s.io/apiserver-network-proxy/pkg/agent"
 	"sigs.k8s.io/apiserver-network-proxy/pkg/util"
 )
 
-const ReadHeaderTimeout = 60 * time.Second
+const (
+	ReadHeaderTimeout   = 60 * time.Second
+	LeaseNamespace      = "kube-system"
```
ipochi, line 55: I believe, this too should be passed down as a cli flag

carreter: I'll open up a separate issue for this along w/ passing the label selector as a CLI flag as well.

pkg/agent/clientset.go
carreter: Removed the name argument.

pkg/agent/lease_counter.go
carreter: Removed.

pkg/agent/lease_counter.go
carreter: Added!

pkg/server/leases/controller.go
carreter: Removed.

cmd/agent/app/server.go
carreter: We ended up deciding to hard code it as this is the most common use case. Will put up an issue to have an optional flag to configure this.

e2e/templates/agent/deployment.yaml
```diff
@@ -18,15 +18,18 @@ spec:
       containers:
       - name: konnectivity-agent-container
         image: {{ .Image }}
+        securityContext:
+          allowPrivilegeEscalation: false
+          runAsUser: 0
```
cheftako, line 23: Why are we running as the root user?

e2e/templates/server/deployment.yaml
```diff
@@ -14,16 +14,27 @@ spec:
...
+        operator: "Exists"
+      - operator: "Exists"
+      nodeSelector:
+        node-role.kubernetes.io/control-plane: ""
       containers:
       - name: konnectivity-server-container
         image: {{ .Image }}
+        securityContext:
+          allowPrivilegeEscalation: false
+          runAsUser: 0
```
cheftako, line 30: Ditto

examples/kind-multinode/rendered/konnectivity-server.yaml
```diff
@@ -0,0 +1,127 @@
...
+      hostNetwork: true
+      containers:
+      - name: konnectivity-server-container
+        image: gcr.io/carreter-dev2/proxy-server-amd64:01e3cce3bd4c8e4cff2366d1169c0326a2f1af02
+        resources:
+          requests:
+            cpu: 1m
+        securityContext:
+          allowPrivilegeEscalation: false
+          runAsUser: 0
```
cheftako, line 64: Especially in our examples I don't think we should run as root.

e2e/templates/agent/deployment.yaml
carreter: I was copying the pre-existing example in `kind/`. I believe it is necessary for access to the kubeconf in `/etc/kubernetes`.

pkg/server/leases/gc_controller.go
```diff
@@ -70,6 +71,8 @@ func (c *GarbageCollectionController) gc(ctx context.Context) {
 			klog.V(4).Infof("Lease %v was already deleted", lease.Name)
```
avrittrohwer, line 71: Could you add HTTP-level counter and latency around this delete call?

pkg/agent/lease_counter.go
```diff
@@ -0,0 +1,142 @@
...
+				start := time.Now()
+				defer func() {
+					latency := time.Now().Sub(start)
+					metrics.Metrics.ObserveLeaseListLatency(latency)
```
avrittrohwer, line 100: please add a http_response_code field to the latency metric

pkg/server/leases/gc_controller.go
```diff
@@ -0,0 +1,78 @@
...
+		pc:             pc,
+	}
+}
+
+func (c *GarbageCollectionController) Run(ctx context.Context) {
+	wait.UntilWithContext(ctx, c.gc, c.gcCheckPeriod)
+}
+
+func (c *GarbageCollectionController) gc(ctx context.Context) {
+	leases, err := c.leaseInterface.List(ctx, metav1.ListOptions{LabelSelector: c.labelSelector})
```
avrittrohwer, line 56: Please add HTTP-level counter and latency metrics around this List call

pkg/agent/lease_counter.go
carreter: Done.

pkg/server/leases/gc_controller.go
carreter: Done.

pkg/server/leases/gc_controller.go
carreter: Done.

pkg/agent/lease_counter.go
avrittrohwer: Discussed in the bi-weekly meeting, but general feedback was that we don't envision a case where agents will run without servers, so making the floor 1 is more defensive to faults in the lease counting mechanism

e2e/templates/server/deployment.yaml
carreter: This error occurs if the server doesn't run as root:

```
│ E0823 21:30:49.548374       1 main.go:48] error: failed to load kubernetes client config: error loading config file "/etc/kubernetes/admin.conf": open /etc/kubernetes/admin.conf: permission denied            │
```

pkg/server/leases/gc_controller_test.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+func TestGarbageCollectionController(t *testing.T) {
+	testCases := []struct {
+		name           string
+		template       proxytesting.LeaseTemplate
+		selector       string
+		expectDeletion bool
+	}{
+		{
```
avrittrohwer, line 38: nit: mixing time.Minute and DurationSecs makes me have to do mental math, we could use time.Second instead

pkg/server/metrics/metrics.go
```diff
@@ -148,6 +154,48 @@ func newServerMetrics() *ServerMetrics {
 			"reason",
 		},
 	)
+	culledLeases := prometheus.NewCounter(prometheus.CounterOpts{
```
avrittrohwer, line 157: is this metric the same as lease_delete_total with http_status_code 200?

avrittrohwer (approved):

pkg/server/metrics/metrics.go
carreter: Pretty much, yeah.

cheftako:
Nice work! Thank you!

avrittrohwer:
Congrats on the work @carreter !  Very proud of your internship :)

pkg/agent/lease_counter.go
```diff
@@ -0,0 +1,144 @@
...
+
+	// Ensure returned count is always at least 1.
+	if count == 0 {
+		klog.Warningf("No valid leases found, assuming server count of 1")
+		count = 1
+	}
+
+	if count != lc.fallbackCount {
+		lc.fallbackCount = count
+	}
```
tallclair, line 88: nit: prefer
```suggestion
		lc.fallbackCount = count
```

tallclair (commented):
Looks like I was too slow on my review :)

Dropping these comments in for posterity. Nothing important, we can do a follow-up cleanup. I think the biggest question I had was whether we really need the dedicated lease request metrics.

PR #655, opened 2024-09-16
Use protobuf encoding for core K8s APIs in apiserver-network-proxy

tosi3k:
For core K8s API objects like Pods, Nodes, etc., we can use protobuf encoding which, compared to the default JSON encoding, reduces CPU consumption related to (de)serialization, reduces overall latency of the API calls, reduces memory footprint and the work performed by the GC and results in quicker propagation of objects to event handlers of shared informers.

Core system components of K8s default their serialization method to protobuf for 8 years already: https://github.com/kubernetes/kubernetes/pull/25738.

Some benchmarks comparing JSON vs. protobuf showcasing how the latter data format (de)serializes faster and uses less memory:

- https://medium.com/@akresling/go-benchmark-json-v-protobuf-4ec3c62ec8d4
- https://www.codingexplorations.com/blog/performance-comparison-protobuf-marshaling-vs-json-marshaling-in-go
- https://medium.com/@david_turner/protocol-buffers-just-how-fast-are-they-9ec19ea580db


jkh52:
The only failure is with lint, and this PR is not likely the culprit:

```
level=info msg="[runner] linters took 15.706254297s with stages: goanalysis_metalinter: 15.075948509s"
cmd/agent/app/server.go:104:2: undefined: klog (typecheck)
	klog.V(1).Infoln("Shutting down agent.")
	^
cmd/agent/app/server.go:126:2: undefined: klog (typecheck)
	klog.V(2).InfoS("Received first signal", "signal", s)
	^
cmd/agent/app/server.go:129:2: undefined: klog (typecheck)
	klog.V(2).InfoS("Received second signal", "signal", s)
	^
pkg/server/backend_manager.go:64:1: missing return (typecheck)
}
^
```

jkh52:
> I like the change. I have concern that there might be edge cases where this _might_ cause problems. I do not know of any but wonder for instance if we have adequately tested this with the system running in HTTP-CONNECT mode rather than GRPC. I think it would be good to at least have a flag to allow someone to revert to the legacy behavior if they do suspect a problem.

+1 to a flag, but the default should be protobuf (I don't see much risk here).

tosi3k:
Thanks - added appropriate flag to both the agent and the server with defaulting to protobuf.

PTAL :)

jkh52:
/lgtm
/approve

(The lint error should hopefully be resolved soon, and this PR will become merge-able.)

tosi3k:
@jkh52 given it's an unrelated linter failure, maybe we can force submit the PR? :)

Although I suppose it could wait for Monday given Friday is the worst day for releasing / submitting anything, especially in controversial ways like force submitting.

PR #658, opened 2024-10-03
Fix lint: move to Github Action

jkh52:
Update golangci-lint from 1.54.0 to v1.56.2, per needed for go1.22 per https://github.com/golangci/golangci-lint/issues/4273.

Introduce a Github Action based linter.

Fix resulting findings.

### DIFF
--- a/.github/workflows/golangci-lint.yaml
+++ b/.github/workflows/golangci-lint.yaml
@@ -1,0 +1,29 @@
+name: golangci-lint
+on:
+  push:
+    branches:
+      - master
+      - release-0.31
+  pull_request:
+
+permissions:
+  contents: read
+
+env:
+  GO_VERSION: v1.22
+  GOLANGCI_LINT_VERSION: v1.56
+
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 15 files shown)


jkh52:
Now, golangci-lint appears to succeed but then "hangs". Any ideas, @liangyuanpeng or @cheftako ?

jkh52:
The lint authors recommend using Github Actions: https://golangci-lint.run/welcome/install/#ci-installation

This works for me, see latest snapshot.

jkh52:
Although `pull-apiserver-network-proxy-make-lint-master` continues to fail, see the successful `golangci-lint / lint (pull_request)`.

/retest

jkh52:
@cheftako friendly ping.

It would be nice to get lint working again against mainline. I would have liked to simply adjust Makefile but I never solved a remaining mystery why it hangs when invoked from kubekins / test-infra. At the same linter version it succeeds very quickly via Github Action. Migrating lint to Github Actions might be superior going forward, even if we did not have that mystery.

liangyuanpeng:
I first tried to reproduce it in the local /github action, but with no luck, I am trying to test it by upgrading golangci-lint to 1.59.0 in https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/664

PR #659, opened 2024-10-03
Updated clinet.go to 0.31.1

cheftako:
Trying to unblock dependabot and enable the 0.31 release. Trying to upgrade dependencies in stages.

$ go get k8s.io/client-go@v0.31.1
$ go mod tidy
$ go mod vendor


jkh52:
/lgtm

I'm good with force submitting this, if the only failure is lint.

PR #670, opened 2024-10-25
make lease label and lease namesapce configurable

ipochi:
Currently the lease controller system has hard coded values for lease namespace and lease labels.
This PR adds cli flags to make those values configurable.

### DIFF
--- a/cmd/agent/app/options/options.go
+++ b/cmd/agent/app/options/options.go
@@ -85,6 +85,10 @@ type GrpcProxyAgentOptions struct {
 	// Enables updating the server count by counting the number of valid leases
 	// matching the selector.
 	CountServerLeases bool
+	// Namespace where lease objects are managed.
+	LeaseNamespace string
+	// Labels on which lease objects are managed.
+	LeaseLabel string
 	// Path to kubeconfig (used by kubernetes client for lease listing)
 	KubeconfigPath string
 	// Content type of requests sent to apiserver.
@@ -132,6 +136,8 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.BoolVar(&o.SyncForever, "sync-forever", o.SyncForever, "If true, the agent continues syncing, in order to support server count changes.")
 	flags.IntVar(&o.XfrChannelSize, "xfr-channel-size", 150, "Set the size of the channel for transferring data between the agent and the proxy server.")
 	flags.BoolVar(&o.CountServerLeases, "count-server-leases", o.CountServerLeases, "Enables lease counting system to determine the number of proxy servers to connect to.")
+	flags.StringVar(&o.LeaseNamespace, "lease-namespace", o.LeaseNamespace, "Namespace where lease objects are managed.")
+	flags.StringVar(&o.LeaseLabel, "lease-label", o.LeaseLabel, "The labels on which the lease objects are managed.")
 	flags.StringVar(&o.KubeconfigPath, "kubeconfig", o.KubeconfigPath, "Path to the kubeconfig file")
 	flags.StringVar(&o.APIContentType, "kube-api-content-type", o.APIContentType, "Content type of requests sent to apiserver.")
 	return flags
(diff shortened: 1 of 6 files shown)


cmd/agent/app/options/options.go
```diff
@@ -159,6 +165,9 @@ func (o *GrpcProxyAgentOptions) Print() {
 	klog.V(1).Infof("AgentIdentifiers set to %s.\n", util.PrettyPrintURL(o.AgentIdentifiers))
 	klog.V(1).Infof("WarnOnChannelLimit set to %t.\n", o.WarnOnChannelLimit)
 	klog.V(1).Infof("SyncForever set to %v.\n", o.SyncForever)
+	klog.V(1).Infof("CountServerLeases set to %v.\n", o.CountServerLeases)
```
cheftako, line 168: Good catch

cmd/agent/app/options/options.go
```diff
@@ -132,6 +136,8 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.BoolVar(&o.SyncForever, "sync-forever", o.SyncForever, "If true, the agent continues syncing, in order to support server count changes.")
 	flags.IntVar(&o.XfrChannelSize, "xfr-channel-size", 150, "Set the size of the channel for transferring data between the agent and the proxy server.")
 	flags.BoolVar(&o.CountServerLeases, "count-server-leases", o.CountServerLeases, "Enables lease counting system to determine the number of proxy servers to connect to.")
+	flags.StringVar(&o.LeaseNamespace, "lease-namespace", o.LeaseNamespace, "Namespace where lease objects are managed.")
```
cheftako, line 139: [Nice To Have]: The agent needs to be using the querying the same lease namespace and label as the server is writing to. Would be nice if we either got them from the server or  at least validated that they are the same. Might be good to eventually add this as metadata on the handshake.

cmd/server/app/options/options.go
```diff
@@ -321,6 +330,13 @@ func (o *ProxyRunOptions) Validate() error {
 			}
 		}
 	}
+	// Validate labels provided.
+	if o.EnableLeaseController {
+		_, err := util.ParseLabels(o.LeaseLabel)
```
cheftako, line 335: rather than parsing the labels more than once, would it make sense to store the result of the parsed label at this point?

pkg/util/util.go
```diff
@@ -0,0 +1,27 @@
...
+	labels := make(map[string]string)
+
+	if len(labelStr) == 0 {
+		return labels, fmt.Errorf("empty string provided")
+	}
+	pairs := strings.Split(labelStr, ",")
+
+	for _, pair := range pairs {
+		keyValue := strings.SplitN(pair, "=", 2)
+		if len(keyValue) != 2 {
```
cheftako, line 19: If you use SplitN then keyValue will only be != 2 if there are no '=' in the input. In this case I think it would be safer to use Split. This will allow "foo=bar=err" to return {"foo", "bar", "err"} and cause an error for length 3. As it stands SplitN will return {"foo", "bar=err"} which has length 2.

cheftako:
Nice change. I would like to see the Split vs SplitN change. The rest we can treat as feature requests.

cmd/server/app/options/options.go
ipochi: Didn't want to pollute the `ProxyRunOptions` struct with another label field, especially we are using once in validate and once to setup the server.

ipochi:
@cheftako I believe this is good to get merged.

pkg/util/util_test.go
```diff
@@ -0,0 +1,74 @@
...
+		{
+			input: "app=myapp,env=prod,version=1.0",
+			expectedOutput: map[string]string{
+				"app":     "myapp",
+				"env":     "prod",
+				"version": "1.0",
+			},
+			shouldError: false,
+		},
+		{
```
cheftako, line 22: Would be nice to have a test case like "foo=var=baz"

PR #675, opened 2024-11-15
KNP agent should pick the larger of the two server counts

konryd:
This way, when agent and server both switch from server-provided counts to counting leases, we will maintain compatibility in both directions:
  * response-reliant agent with lease-enabled server: would work, as servers still send down the (static, flag-provided) server count in the response header
  * lease-counting agent with non-lease-enabled server: would work, as agent would fall back on the number provided by the servers

Thanks to this compatibility, enabling lease-counting in a fleet of clusters is much more practical (there's no need for two separate, interdependent rollouts).

### DIFF
--- a/pkg/agent/clientset.go
+++ b/pkg/agent/clientset.go
@@ -30,6 +30,12 @@ import (
 	"sigs.k8s.io/apiserver-network-proxy/pkg/agent/metrics"
 )
 
+const (
+	fromResponses = "KNP server response headers"
+	fromLeases = "KNP lease count"
+	fromFallback = "fallback to 1"
+)
+
 // ClientSet consists of clients connected to each instance of an HA proxy server.
 type ClientSet struct {
 	mu      sync.Mutex         //protects the clients.
@@ -68,6 +74,7 @@ type ClientSet struct {
 	xfrChannelSize     int
 
 	syncForever bool // Continue syncing (support dynamic server count).
+	serverCountSource string
 }
 
 func (cs *ClientSet) ClientsCount() int {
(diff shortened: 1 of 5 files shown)


pkg/agent/clientset.go
```diff
@@ -218,15 +218,25 @@ func (cs *ClientSet) sync() {
...
-	} else {
-		serverCount = cs.lastReceivedServerCount
+		countFromLeases = cs.leaseCounter.Count()
+	}
+	countFromResponses := cs.lastReceivedServerCount
+
+	serverCount := countFromLeases
+	countSource := "KNP server lease count"
+	if countFromResponses > serverCount {
+		serverCount = countFromResponses
```
cheftako, line 230: Not sure this is what we want. This clearly works when we are scaling up the number of KNP servers. However I believe this prevents us from easily scaling down the number of KNP servers. I thought was one of the goals of the lease system.

konryd: That depends on how the static `--server-count` flag is set. With these changes it can:
  * be set to 1 to rely on leases alone
  * be set to some-default-value to provide a lower bound as a safety

The default lower bound is critical when we're gradually replacing agents and servers in parallel: 
We will still have a portion of the servers not recording their lease and a portion of agents assuming too-low count (because they only look at leases).

pkg/agent/clientset.go
```diff
@@ -218,15 +221,45 @@ func (cs *ClientSet) sync() {
...
+
 	var serverCount int
-	if cs.leaseCounter != nil {
-		serverCount = cs.leaseCounter.Count()
-	} else {
-		serverCount = cs.lastReceivedServerCount
+	var countSourceLabel string
+
+	switch cs.serverCountSource {
+	case "":
```
cheftako, line 232: Golang does not do fall through (https://go.dev/ref/spec#Expression_switches "the first one that equals the switch expression triggers execution of the statements of the associated case; the other cases are skipped") So I believe empty string is going to silently do nothing, which I don't think is what we want. 

Did you mean the following?
case "", "default":

konryd: Thanks! Fixed.

PR #677, opened 2024-11-20
Update golang to 1.22.7

cheftako:
Pending dependency changes require golang 1.22.7.
Upgrading to allow those dependencies to go through.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/apiserver-network-proxy
 
-go 1.22.5
+go 1.22.7
 
 require (
 	github.com/google/uuid v1.6.0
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/apiserver-network-proxy
 
-go 1.22.5
+go 1.22.7
```
jkh52, line 3: This file lost a "toolchain" directive fairly recently, should we restore it?

jkh52:
/lgtm
/hold

Hold for the toolchain question. Feel free to punt.

go.mod
cheftako: I'd like to get it back here and in the matching test-infra repo. However right now I'd like to unblock dependabot.

PR #684, opened 2025-01-08
Updated golang version to 1.23.0

cheftako:
go mod edit -go 1.23.0
go mod vendor
make clean
make certs
make gen
make build


tallclair:
/lgtm

Test failures are a bit concerning though

PR #685, opened 2025-01-08
🐛 Respect log level from the command-line arguments

makhov:
Currently, the log level is hardcoded and konnectivity is usually quite noisy

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -36,9 +36,11 @@ func main() {
 	flags.AddFlagSet(o.Flags())
 	local := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	klog.InitFlags(local)
-	err := local.Set("v", "4")
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "error setting klog flags: %v", err)
+	if local.Lookup("v") == nil {
+		err := local.Set("v", "4")
+		if err != nil {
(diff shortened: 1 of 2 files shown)


makhov:
Great to see it merged, thank you! Do you have any plans for a new release with the fix?

ipochi:
@makhov @cheftako 

This PR introduced a condition (`if local.Lookup("v") == nil`) before setting the default klog verbosity level. This condition was incorrect because `klog.InitFlags()` always defines the `-v` flag, meaning the default verbosity was never being applied. This caused the effective log level to fall back to klog's default of `0` instead of the intended `4`.

to test: 

1. build proxy-server image on release-0.30 branch
2. run the image passing v=4 or v=0 etc or any other level and you'll find the the explicit value of v is respected.

to test that this PR introduced an error

1. build on latest master or 0.31, 0.32, or 0.33 branches (where this fix is backported)
2.  DO NOT pass any flag, according to code, it should take up the default value of 4. Yet, it doesn't it takes the default of klog called during InitFlags() is 0 and hence  `local.Lookup("v") == nil` is never true.

makhov:
@ipochi thanks fot noticing! Looks like I accidentally messed up an order of thigs. As I [described in the PR comments](https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/782#pullrequestreview-3243168897), I'm not a big fun of reverting the change. Let me just fix it.

PR #687, opened 2025-01-10
Added debug to apparent stop channel called early.

cheftako:


pkg/agent/client.go
```diff
@@ -538,10 +538,19 @@ func (a *Client) remoteToProxy(connID int64, eConn *endpointConn) {
 	}
 
 	for {
+		select {
+		case <-a.stopCh:
+			return
+		default:
+		}
+		timeout := time.Now().Add(15 * time.Second)
+		eConn.conn.SetReadDeadline(timeout)
```
tallclair, line 547: I think what we want to do here is close the connection if stopch is closed. Something like this?

```go
go func() {
  <-a.stopCh
  eConn.conn.Close()
}
```

cheftako: This function has a defer cleanup() call on line 533. When we return it should get called. That cleanup makes a protected call to https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/c1a1e092d30133901cddf3ada0b41678f95ca0e7/pkg/agent/client.go#L386. That will then close() the connection.

tallclair: Is it a problem to call `close()` multiple times? I wasn't suggesting that we were leaving the connection open, but rather calling close should be a way to quickly react to the stop channel, rather than polling for it every 15 seconds, which is what I understood your change to be doing?

cheftako: I don't believe close in golang is re-entrant. In fact I think the reason we had to protect the clean up with call once in that code path is specifically because call causes a panic if called twice.

pkg/agent/client.go
```diff
@@ -538,10 +538,19 @@ func (a *Client) remoteToProxy(connID int64, eConn *endpointConn) {
...
 		n, err := eConn.conn.Read(buf[:])
 		klog.V(5).InfoS("received data from remote", "bytes", n, "connectionID", connID)
 
-		if err == io.EOF {
+		if err == os.ErrDeadlineExceeded {
+			continue
```
tallclair, line 552: Can you add a unit test for this case? Maybe put the timeout in a global variable that can be overridden for a faster test. In particular, I want to make sure that a streaming connection that doesn't send any data for a window longer than the timeout won't be broken.

pkg/agent/client.go
tallclair: maybe just call `eConn.cleanup()` then?

cmd/agent/app/server.go
```diff
@@ -99,6 +99,7 @@ func (a *Agent) Run(o *options.GrpcProxyAgentOptions, drainCh, stopCh <-chan str
 	defer a.adminServer.Close()
 
 	<-stopCh
+
```
tallclair, line 102: nit: revert this?

pkg/agent/client.go
cheftako: We called `defer eConn.cleanup()` on line 533. So returning should mean it gets called.

pkg/agent/client.go
cheftako: Would you be good using the delayedServer in the proxy_test for this? Currently it waits between 500ms and 2 seconds. I could bump that up to waiting between 5 and 10 seconds. Then as you suggest shortening the timeout for the read on the test from 15 seconds down to 1 second.

cheftako: Done

pkg/agent/client.go
cheftako: tallclair is going to try the separate goroutine suggestion.

cheftako:
Original blocking issue seems to have at least temporarily gone away.
Putting a hold on this until the issue comes back.
/hold

cheftako:
/unhold
Original problem is back. Attempts to extend timeout aren't working. Now we are seeing the problem on TestProxyDial_RequestCancelled_Concurrent_GRPC. These are blocking on the same read call.

PR #696, opened 2025-01-25
tests: tweak timings so we cancel during HTTP processing

justinsb:
There are 3 possibilities:
* We cancel before the HTTP request (possibly even before the dial has completed)
* We cancel during the delay in our slow server
* We cancel after request processing

Because our intervals were [0, 1s] and [0.5s, 2s] I think we
previously could have hit all three cases.

Likely there is a bug lurking in one of these three cases,
but let's start by making the delays such that we are
always in the second situation (assuming dialing localhost
takes < 1s)

We now delay the HTTP response by 3 seconds, and cancel
after a random value in the interval [1s, 2s)

### DIFF
--- a/tests/proxy_test.go
+++ b/tests/proxy_test.go
@@ -105,17 +105,26 @@ type delayedServer struct {
 	maxWait time.Duration
 }
 
-func newDelayedServer() *delayedServer {
+// randomDuration returns a random duration in the [min, max) interval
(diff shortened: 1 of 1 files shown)


justinsb:
OK, a full green run.  But maybe it's just luck...

/retest

justinsb:
Still green, but just in case ...

/test pull-apiserver-network-proxy-test-master

PR #697, opened 2025-01-26
Fixed comment for connect() return.

cheftako:
Agent connect() does not return the server id.
It returned the number of servers we should connect to.

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -174,8 +174,8 @@ func newAgentClient(address, agentID, agentIdentifiers string, cs *ClientSet, op
 	return a, serverCount, nil
 }
 
-// Connect makes the grpc dial to the proxy server. It returns the serverID
-// it connects to.
+// Connect makes the grpc dial to the proxy server. It returns the count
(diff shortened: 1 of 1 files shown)


pkg/agent/client.go
```diff
@@ -174,8 +174,8 @@ func newAgentClient(address, agentID, agentIdentifiers string, cs *ClientSet, op
 	return a, serverCount, nil
 }
 
-// Connect makes the grpc dial to the proxy server. It returns the serverID
-// it connects to.
+// Connect makes the grpc dial to the proxy server. It returns the count
+// of the number of servers this agent should connected to.
```
justinsb, line 178: Sorry to nit pick, but:

```suggestion
// of the number of servers this agent should connect to.
```

(Feel free to just make the change locally and repush, it's just as easy as accepting, rebasing etc; if that's what we do in this repo)

justinsb:
Actually I think I'm going to go through and add lots of comments around serverCount, so let's just merge regardless of my nit.

/approve
/lgtm

PR #699, opened 2025-01-28
Updating baseimage to pick up assorted security fixes.

cheftako:
Going from Debian11 to Debian12 base image.


justinsb:
Thanks for doing this.

I was confused myself so looked it up, and the static images contain no binaries:

https://github.com/GoogleContainerTools/distroless/blob/main/base/README.md

> Statically compiled applications (Go) that do not require libc can use the gcr.io/distroless/static image, which contains:
> 
> ca-certificates
> A /etc/passwd entry for a root user
> A /tmp directory
> tzdata

But of course the ca-certificates do occasionally change.

/approve
/lgtm

PR #700, opened 2025-01-29
refactor: simplify ClientSet and add comments

justinsb:


pkg/agent/clientset.go
```diff
@@ -96,15 +96,12 @@ func (cs *ClientSet) HealthyClientsCount() int {
 
 }
 
-func (cs *ClientSet) hasIDLocked(serverID string) bool {
-	_, ok := cs.clients[serverID]
-	return ok
-}
-
+// HasID returns true if the ClientSet has a client to the specified serverID.
 func (cs *ClientSet) HasID(serverID string) bool {
```
jkh52, line 100: HasID looks unused, can be trimmed?

jkh52:
/lgtm
/approve

/hold

HOLD only for the optional nit and allow @cheftako to look.

pkg/agent/clientset.go
justinsb: So because it is exported, I wasn't sure that it was safe to remove - it _could_ be used from another project maybe?  But maybe it's only `konnectivity-client` that we consider part of our "API contract"?

Do we run an e2e where we just check that e.g. kube-apiserver still builds - I think that's probably our primary client, and I think I'd be very comfortable removing exported functions as long as apiserver still compiles.

On the other hand, I do agree, it's probably safe to remove, but maybe lots of small safe cleanups are better than one big (and potentially risky) cleanup?

jkh52: > So because it is exported, I wasn't sure that it was safe to remove - it _could_ be used from another project maybe? But maybe it's only `konnectivity-client` that we consider part of our "API contract"?

Your point is right, but historically pkg/agent has always been specific to konnectivity-agent binary.

> Do we run an e2e where we just check that e.g. kube-apiserver still builds - I think that's probably our primary client, and I think I'd be very comfortable removing exported functions as long as apiserver still compiles.

apiserver doesn't depend on pkg/agent. This repo has .github/workflows/e2e.yaml checking apiserver for k8s versions against locally-built konnectivity-server and konnectivity-agent, but mistakes made in konnectivity-client aren't fully detected until attempts to release to k/k.

> 
> On the other hand, I do agree, it's probably safe to remove, but maybe lots of small safe cleanups are better than one big (and potentially risky) cleanup?

I would still include it, but no objections to keeping as-is, this PR is strictly improvement.

jkh52:
Let's get this merged.

/unhold

PR #701, opened 2025-01-29
chore: fix field comments on ClientSet

justinsb:
Not following the go conventions here means the wrong comments
appear in IDEs.

### DIFF
--- a/pkg/agent/clientset.go
+++ b/pkg/agent/clientset.go
@@ -38,42 +38,67 @@ const (
 
 // ClientSet consists of clients connected to each instance of an HA proxy server.
 type ClientSet struct {
-	mu      sync.Mutex         //protects the clients.
-	clients map[string]*Client // map between serverID and the client
-	// connects to this server.
+	// mu guards access to the clients map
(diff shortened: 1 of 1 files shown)


pkg/agent/clientset.go
```diff
@@ -38,42 +38,67 @@ const (
...
+	mu sync.Mutex
+
+	// clients is a map between serverID and the client
+	// connected to this server.
+	clients map[string]*Client
+
+	// agentID is "our ID" - the ID of this agent.
+	agentID string
+
+	// Address is the proxy server address.  Assuming HA proxy server
```
cheftako, line 51: TODO to improve this. Its the address we use to contact the ANP Server. Normally this would be a LB address in front of the ANP Servers.

pkg/agent/clientset.go
```diff
@@ -38,42 +38,67 @@ const (
...
+	address string
+
+	// leaseCounter counts number of proxy server leases
```
cheftako, line 54: TODO: This only used if the relevant feature is enabled.

PR #714, opened 2025-02-12
Advance e2e test matrix (remove v1.28 and add v1.31).

jkh52:


liangyuanpeng (commented):
Seems like i missed it, LGTM

Issue #707, opened 2025-02-06
Create 0.32.x release (branch and tags)


batistein:
Draft of  possible tasks:

- [ ] Clear blockers
- [ ] Cut release-0.32 branch
- [ ] Create tag 0.32.0
- [ ] Promote images

jkh52:
@batistein I gathered from slack that you're interested in helping, is that right? I think you can send most PRs according to [RELEASE.md](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/RELEASE.md), but tag and release branch creation takes privileges and I can handle those.

@cheftako or @ipochi (or others), do you have anything specific that should be merged to master before cutting release-0.32 branch and 0.32.0 tag?

lucasrattz:
> [@batistein](https://github.com/batistein) I gathered from slack that you're interested in helping, is that right? I think you can send most PRs according to [RELEASE.md](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/RELEASE.md), but tag and release branch creation takes privileges and I can handle those.
> 
> [@cheftako](https://github.com/cheftako) or [@ipochi](https://github.com/ipochi) (or others), do you have anything specific that should be merged to master before cutting release-0.32 branch and 0.32.0 tag?

@batistein and I will work on this as soon as the 0.32 branch and tag are available. 

I noticed the container image for 0.31.2 is not yet on the OCI repository, so we'll take care of promoting it too.

jkh52:
I sent #713, #714, and #715 that I think make sense before branching `release-0.32`.

jkh52:
/assign @cheftako 

Hey @cheftako, I'm still waiting on #715 which I think should make 0.32 release. Can you review and assign back when you're ready for that to proceed?

Also - I see some cherry picks (#717, #720). Do those mean we need to create 0.29 and 0.31 patch tags?

jkh52:
/assign @batistein 

I created the 0.32 branch and first tag.

lucasrattz:
> /assign [@batistein](https://github.com/batistein)
> 
> I created the 0.32 branch and first tag.

I'll open the PR for releasing the container images.

PR #715, opened 2025-02-12
Updated golang version to 1.23.6

jkh52:
Build the server and agent binaries with 1.23.6.

Refresh golangci-lint.yaml.

### DIFF
--- a/.github/workflows/golangci-lint.yaml
+++ b/.github/workflows/golangci-lint.yaml
@@ -4,13 +4,14 @@ on:
     branches:
       - master
       - release-0.31
+      - release-0.32
   pull_request:
 
 permissions:
   contents: read
 
 env:
-  GO_VERSION: v1.23
+  GO_VERSION: v1.23.6
   GOLANGCI_LINT_VERSION: v1.60.1
 
 jobs:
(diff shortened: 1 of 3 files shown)


jkh52:
/assign @cheftako 

PTAL, I wasn't sure if you had a specific reason to stop at 1.23.0, when there were minor versions available with fixes (https://go.dev/doc/devel/release#go1.23.0)

cheftako:
I think there are a few other places which need to change at the same time. 
Eg.
https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/.github/workflows/golangci-lint.yaml

jkh52:
> I think there are a few other places which need to change at the same time. Eg. https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/.github/workflows/golangci-lint.yaml

I'm not sure that one is a problem, the logs in this PR for that lint step show it picking up 1.23.6. Opened #719 separately.

Do you see any other places missed?

jkh52:
PTAL, updated the golangci-lint.yaml

jkh52:
Friendly ping.

It would be nice to include this in #707.

PR #719, opened 2025-02-20
golangci-lint refresh.

jkh52:
- Adapt latest example from https://github.com/golangci/golangci-lint-action
- Prepare for release-0.32 branch


jkh52:
Folded into #715

PR #731, opened 2025-04-15
fix: k8s client setup if agent service account auth is not used

ipochi:
Currently the setting up of k8s client is broken if service account
authentication is not used between server and agent.

This condition  `if o.AgentNamespace != "" {` acts as a gatekeeper for
setting the k8s client which worked fine previously as server never
needed to talk to apiserver apart from authenticating agents using
service account token.

However when lease controller logic was added, it meant that setting up
k8s client was required if lease controller was enabled but
authentication was done using mTLS instead of service account
authentication.

This fixes that.

Closing https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/728  in favour of this.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -20,7 +20,7 @@ env:
 jobs:
   build:
     name: build
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
     - name: Check out code
       uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


ipochi:
e2e test failure is related to https://github.com/actions/runner-images/issues/11985

/cc @cheftako

cheftako:
Change looks good. Lets see what we need to do to get the tests healthy.

liangyuanpeng:
ipv6 runners have a kernel bug https://github.com/actions/runner-images/issues/11985 , this is the reason of CI failed.

ipochi:
@cheftako @jkh52 

after updating the github actions to run on 24.04, the issue involved with runners is resolved.
PR to update the runner: https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/734

PR #732, opened 2025-04-25
Fix error messages for test-client in UDS mode

antoninbas:
As an example, in UDS mode, certificates should not be provided, but the error message reporting this was stating the opposite.

### DIFF
--- a/cmd/test-client/main.go
+++ b/cmd/test-client/main.go
@@ -167,13 +167,13 @@ func (o *GrpcProxyClientOptions) Validate() error {
 	}
 	if o.proxyUdsName != "" {
 		if o.proxyHost != "" {
-			return fmt.Errorf("please do set proxy host when using UDS")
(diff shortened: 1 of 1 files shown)


cmd/test-client/main.go
```diff
@@ -167,13 +167,13 @@ func (o *GrpcProxyClientOptions) Validate() error {
 	}
 	if o.proxyUdsName != "" {
 		if o.proxyHost != "" {
-			return fmt.Errorf("please do set proxy host when using UDS")
+			return fmt.Errorf("please set proxy host to \"\" when using UDS")
```
jkh52, line 170: I see these zero values must be set due to defaulting, and it matches the [README instructions](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/README.md?plain=1#L157)

Thanks for improving!

PR #734, opened 2025-04-29
chore: bump github action runson to ubuntu-24.04

ipochi:
There is an error in github actions runner that runs on 22.04. This issue hasn't been rectified yet but the issue is not reported on 24.04

issue details: https://github.com/actions/runner-images/issues/11985

So this bumps the action to run on ubuntu 24.04


ipochi:
@cheftako

PR #749, opened 2025-06-04
Fix panic when mtls is disabled

JanHolger:
When mTLS is disabled (no --server-ca-cert) previously the server panic'ed due to an index error when trying to access the first peer certificate since go does not even parse client certificates in this case. As this was only accessed for logging the common name of the client, I've added a second check to check if there is a peer certificate present to the existing check that already checked if tls is enabled.


jkh52:
/ok-to-test

Looks like a very old bug, thank you for fixing!

PR #754, opened 2025-06-27
Support proxy_server_ready_backends metric

### LINKED ISSUE #294: [BUG] The backend ServerMetric is not correct

maqiuyujoyce:
Fixes #294
Prior art: #295

Added a new metric to track the total number of backends based on proxy strategy. To avoid circular import, I extracted out ProxyStrategy related types, functions and tests into a separate package.

I didn't overwrite the existent `proxy_server_ready_backend_connections` metric because of potential backwards compatibility concern brought up here: https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/295/files#r872839718 Instead, I marked the problematic `proxy_server_ready_backend_connections` metric deprecated in the help message.

Tested locally following [this example](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/README.md#grpc-client-using-mtls-proxy-with-dial-back-agent) and got the following result when supporting both default and desthost backend managers:
```
# HELP konnectivity_network_proxy_server_ready_backend_connections Number of konnectivity agent connected to the proxy server
# TYPE konnectivity_network_proxy_server_ready_backend_connections gauge
konnectivity_network_proxy_server_ready_backend_connections 1
# HELP konnectivity_network_proxy_server_ready_backend_connections_total Total number of konnectivity agent connected to the proxy server
# TYPE konnectivity_network_proxy_server_ready_backend_connections_total gauge
konnectivity_network_proxy_server_ready_backend_connections_total{proxy_strategy="default"} 1
konnectivity_network_proxy_server_ready_backend_connections_total{proxy_strategy="destHost"} 1
```

Added unit tests for the new metric as well.

### DIFF
--- a/pkg/server/metrics/metrics.go
+++ b/pkg/server/metrics/metrics.go
@@ -112,7 +112,7 @@ func newServerMetrics() *ServerMetrics {
 			Namespace: Namespace,
 			Subsystem: Subsystem,
 			Name:      "ready_backend_connections",
-			Help:      "Number of konnectivity agent connected to the proxy server",
+			Help:      "Number of konnectivity agent connected to the proxy server. DEPRECATED, please use ready_backend_connections_total",
 		},
 		[]string{},
 	)
@@ -295,8 +295,8 @@ func (s *ServerMetrics) HTTPConnectionInc() { s.httpConnections.Inc() }
 // HTTPConnectionDec decrements a finished HTTP CONNECTION connection.
 func (s *ServerMetrics) HTTPConnectionDec() { s.httpConnections.Dec() }
 
-// SetBackendCount sets the number of backend connection.
-func (s *ServerMetrics) SetBackendCount(count int) {
+// SetBackendCountDeprecated sets the number of backend connection.
+func (s *ServerMetrics) SetBackendCountDeprecated(count int) {
 	s.backend.WithLabelValues().Set(float64(count))
 }
 
--- a/pkg/server/metrics/metrics.go
+++ b/pkg/server/metrics/metrics.go
@@ -51,6 +51,7 @@ type ServerMetrics struct {
 	grpcConnections      *prometheus.GaugeVec
 	httpConnections      prometheus.Gauge
 	backend              *prometheus.GaugeVec
+	totalBackendCount    *prometheus.GaugeVec
 	pendingDials         *prometheus.GaugeVec
 	establishedConns     *prometheus.GaugeVec
 	fullRecvChannels     *prometheus.GaugeVec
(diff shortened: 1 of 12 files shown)


pkg/server/metrics/metrics.go
```diff
@@ -284,6 +299,11 @@ func (s *ServerMetrics) SetBackendCount(count int) {
 	s.backend.WithLabelValues().Set(float64(count))
 }
 
+// SetTotalBackendCount sets the total number of backend connection.
+func (s *ServerMetrics) SetTotalBackendCount(managerName string, count int) {
+	s.totalBackendCount.WithLabelValues(managerName).Set(float64(count))
```
cheftako, line 304: Labels can generate cardinality issues. I don't think this is likely to be a problem but we can reassure/protect ourselves by changing managerName from a string (unbounded cardinality) to an enumeration with an observation on limiting the number of options.

maqiuyujoyce: Done.

Based on @jkh52 's suggestion above, I used proxy strategy enum as the label. (Thank you @jkh52 !)
It introduced the circular dependency, so I extracted out ProxyStrategy related types and functions into a separate package.

maqiuyujoyce:
> I think there will remain a bug in `ready_backend_connections` (it gets clobbered when the server has more than one strategy). We could mark it deprecated (pkg/agent metrics has a help text example of this) if we think that will never get fixed.

Done.

pkg/server/metrics/metrics.go
```diff
@@ -110,10 +112,21 @@ func newServerMetrics() *ServerMetrics {
...
-			Help:      "Number of konnectivity agent connected to the proxy server",
+			Help:      "Number of konnectivity agent connected to the proxy server. DEPRECATED, please use ready_backend_connections_total",
 		},
 		[]string{},
 	)
+	totalBackendCount := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: Namespace,
+			Subsystem: Subsystem,
+			Name:      "ready_backend_connections_total",
```
jkh52, line 123: Final concern: the _total suffix here seems to contradict prometheus naming conventions (it implies a counter metric, not gauge).

Maybe just "ready_backends" or similar?

Issue #739, opened 2025-05-18
Agent maintains stale server count metadata leading to incomplete connections in rolling update scenario (non-lease controller mode)


ipochi:
Observed a bug in the interaction between the apiserver-network-proxy agent and server components when using a non-lease-controller-based approach to propagate server count metadata. This leads to agents not establishing connections to all available servers, especially during a rolling update where servers scale from N to M.

Environment: Proxy Server using non-lease-controller mode to communicate server count.

Problem:

The agent determines how many proxy server connections to establish based on the serverCount metadata received from the server it connects to. During a rolling update, older server pods may still exist and return stale metadata (e.g., old serverCount=2, even though actual server count has increased to 6). Here's what happens in such a case:

1. Agent initially receives server count 2, hence it connects to 2 servers (e.g., server A and server B).
2. The rollout begins; new servers are added (servers C, D, E, etc.).
3. The agent connects to more servers (e.g., C, D, E) — now having 5 total connections.
4. Later, the agent might get connected again to an old pod(e.g., server A, not yet terminated) as the agent tries the loadbalancer in front of proxy-server. In such a case, the agent receives outdated metadata: serverCount=2.
5. This overwrites the agent's internal view of server count (from 6 → 2).
6. The agent sees that it has 3 active connections (C, D, E) as eventually A and B connections will terminate as those pods will terminate. Agent believes it has met or exceeded the required number (3 ≥ 2).
7. No further connections are attempted, despite the presence of new servers (e.g., A', B', F).
8. As a result, the agent ends up with stale/incomplete connections, and does not recover until a restart.

Sample error log
```
I0504 19:34:38.561148       1 client.go:210] "Connect to server" serverID="67afde44-3666-4d1d-ba11-f15ed3a45668"
I0504 19:34:38.561167       1 clientset.go:213] "Server count change suggestion by server" current=2 serverID="67afde44-3666-4d1d-ba11-f15ed3a45668" actual=6
I0504 19:34:38.561174       1 clientset.go:222] "sync added client connecting to proxy server" serverID="67afde44-3666-4d1d-ba11-f15ed3a45668"
I0504 19:34:38.561198       1 client.go:321] "Start serving" serverID="67afde44-3666-4d1d-ba11-f15ed3a45668" agentID="0af9065f-f093-4687-b578-cc836b2088df"
I0504 19:34:39.405847       1 client.go:528] "remote connection EOF" connectionID=12
[...]

PR #774, opened 2025-09-11
fix: avoid stale server count update [backport to v0.30 release branch]

ipochi:
fixes : #739

we only accept the server count from a server that we have successfully added to our active client set, implicitly ignoring stale data from duplicate connection attempts.


cheftako:
Should we also be cherry picking this to 0.31, 0.32 and 0.33?

Issue #780, opened 2025-09-19
log level broken by a previous PR


ipochi:
A prior change in #685  introduced a condition (`if local.Lookup("v") == nil`) before setting the default klog verbosity level. This condition was incorrect because `klog.InitFlags()` always defines the `-v` flag, meaning the default verbosity was never being applied. This caused the effective log level to fall back to klog's default of `0` instead of the intended `4`.

Reverting the code should solve this issue.

makhov:
@ipochi thanks fot noticing! Looks like I accidentally messed up an order of thigs. As I [described in the PR comments](https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/782#pullrequestreview-3243168897), I'm not a big fun of reverting the change. Let me just fix it.

ipochi:
@makhov 

I'm a bit confused as to what was wrong with the code before #785 was merged.

The code did as expected:

If you don't pass the v flag, it's 4
If you do, that will be set.

makhov:
Actually nothing. Just my assumptions were wrong, sorry. 
On the other hand, I've heard a lot of complains that konnectivity is too noisy, maybe it makes sense to reconsider the default value

PR #782, opened 2025-09-19
Backport to release-0.33 Revert "Merge pull request #685 from makhov/respect-log-level-from-args"

ipochi:
This reverts commit 4ce5f068f3d2acf9e906da7ea9ea430ac877d798.

This commit reverts a previous modification to the flag handling logic in the agent and server's main() functions.

A prior change introduced a condition (`if local.Lookup("v") == nil`) before setting the default klog verbosity level. This condition was incorrect because `klog.InitFlags()` always defines the `-v` flag, meaning the default verbosity was never being applied. This caused the effective log level to fall back to klog's default of `0` instead of the intended `4`.

This change restores the original, correct behavior by unconditionally setting the verbosity to `4`. This ensures logs are visible by default while still allowing the user to override the setting via the command line.

https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/685#issuecomment-3309950955

Fixes: #780

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -36,11 +36,9 @@ func main() {
 	flags.AddFlagSet(o.Flags())
 	local := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	klog.InitFlags(local)
-	if local.Lookup("v") == nil {
-		err := local.Set("v", "4")
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "error setting klog flags: %v", err)
-		}
+	err := local.Set("v", "4")
+	if err != nil {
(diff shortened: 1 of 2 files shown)


makhov (changes requested):
I don't like the idea of reverting this. Now we just have a default level 0, which I'm personaly quite ok with. Reverting this will mean that we have a hardcoded level 4.

The proper fix sill be just to change the order of the check and InitFlags. I'll try to do it asap

makhov:
The problem is that it won't be respected. It's just always 4.

makhov:
You right. Sorry, I misunderstood the current flow.

PR #795, opened 2025-10-27
implement server agent draining

ipochi:
Currently, the server only logs if it receives a drain request from the agent. Ideally, in that scenario, the server should mark it and not select it for newer requests. This PR implements that.

### DIFF
--- a/pkg/server/backend_manager.go
+++ b/pkg/server/backend_manager.go
@@ -49,6 +49,25 @@ type Backend struct {
 	// cached from conn.Context()
 	id     string
 	idents header.Identifiers
+
+	// draining indicates if this backend is draining and should not accept new connections
+	draining bool
+	// mu protects draining field
+	mu sync.RWMutex
+}
+
+// IsDraining returns true if the backend is draining
+func (b *Backend) IsDraining() bool {
+	b.mu.RLock()
+	defer b.mu.RUnlock()
+	return b.draining
+}
+
+// SetDraining marks the backend as draining
+func (b *Backend) SetDraining() {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	b.draining = true
 }
 
 func (b *Backend) Send(p *client.Packet) error {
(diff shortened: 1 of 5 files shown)


ipochi:
> I'm happy to see interest in completing this feature! When I had been looking into this, here's the questions I ran into:
> 
> 1. Whether the server should keep agent draining state by raw connection, or by agent identifiers. The current BackendStorage makes the latter much more complicated, I agree with the former (like this PR does). But note that currently, the agent sync loop will continue to issue new syncs, and a server could get into conflicting drain state for a given agent. So without an accompanying agent change (to stop syncing new connections after entering draining state), it is possible for the server to continue picking these agents.
> 
> That potential agent change is probably an improvement but needs careful risk analysis.
> 
> 2. Fallback logic in case no non-draining agent is found, continue using one that is draining. I think the argument for this is strong, because it's "no worse" than the current server behavior, and may avoid degrading some clusters (those that are limping along with very few agents, e.g. 1).

Thank you for the feedback @jkh52 

Your point on 1, hadn't crossed my mind. It seems logical thing to do to stop the agent sync loop if the agent is draining. I'll give more thought on this whether the server should do more so that it doesn't get into a conflicted state or the agent should stop teh sync loops.

On Point 2, I think its valid point that continue to use the draining agent if there are no non-draining agents. Its akin falling back on current behaviour.

pkg/server/backend_manager.go
```diff
@@ -49,6 +49,25 @@ type Backend struct {
 	// cached from conn.Context()
 	id     string
 	idents header.Identifiers
+
+	// draining indicates if this backend is draining and should not accept new connections
+	draining bool
+	// mu protects draining field
```
cheftako: Consider using `sync/atomic` (e.g., `atomic.Bool` or `int32`) for the `draining` field. This would allow `IsDraining()` to be lock-free, reducing the overhead inside the O(N) loop in `GetRandomBackend`, which runs under the global `s.mu` lock.

pkg/server/backend_manager.go
```diff
@@ -346,9 +365,36 @@ func (s *DefaultBackendStorage) GetRandomBackend() (*Backend, error) {
...
+
+		// Keep track of first draining backend as fallback
+		if firstDrainingBackend == nil {
+			firstDrainingBackend = backend
+		}
+	}
+
+	// All agents are draining, use one as fallback
+	if firstDrainingBackend != nil {
+		agentID := firstDrainingBackend.id
```
cheftako: Inconsistency in log levels: `GetRandomBackend` logs the fallback event at `V(2)`, whereas `DestHostBackendManager` logs a similar fallback event at `V(4)`. It would be better to align them (likely `V(2)` given the significance).

cheftako (commented):
The changes look pretty good and follow the design goals, with the fallback mechanism in backend selection serving as a good safeguard. However, the complexity of `GetRandomBackend` has increased from O(1) to O(N), which is worth noting for performance at scale. Please verify the fallback behavior in `DestHostBackendManager` to ensure it doesn't create hotspots. Most importantly, please add unit tests for `GetRandomBackend` (and `DestHostBackendManager.Backend`) to verify that it prefers non-draining backends and correctly falls back to a draining backend if that is the only option.

PR #815, opened 2026-03-20
feat: add --tls-min-version flag to proxy server

xuezhaojun:
## Summary

- Add `--tls-min-version` flag to the proxy server, allowing dynamic configuration of the minimum TLS version
- Previously the minimum TLS version was hardcoded to TLS 1.2 in `getTLSConfig()`
- Accepted values: `VersionTLS10`, `VersionTLS11`, `VersionTLS12`, `VersionTLS13` (defaults to `VersionTLS12` when not specified, maintaining backward compatibility)

## Motivation

The proxy server currently hardcodes `tls.VersionTLS12` as the minimum TLS version. This makes it impossible for operators to enforce stricter TLS policies (e.g., TLS 1.3 only) or to align with platform-wide TLS security profiles. Adding this flag allows the minimum TLS version to be configured at startup, similar to the existing `--cipher-suites` flag.

## Changes

| File | Change |
|------|--------|
| `pkg/util/net.go` | Add `TLSVersions` map and `GetTLSVersion()` helper function |
| `cmd/server/app/options/options.go` | Add `TLSMinVersion` field, `--tls-min-version` flag, validation, and print |
| `cmd/server/app/server.go` | Update `getTLSConfig()` to accept and use dynamic TLS min version |
| `cmd/server/app/options/options_test.go` | Add default value test and validation tests for valid/invalid TLS versions |

## Test plan

- [x] `TestDefaultServerOptions` passes — verifies `TLSMinVersion` defaults to `""`
- [x] `TestValidate` passes — covers `VersionTLS12`, `VersionTLS13`, and invalid version
- [x] All existing tests pass with no regressions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cmd/server/app/options/options.go
+++ b/cmd/server/app/options/options.go
@@ -17,6 +17,7 @@ limitations under the License.
 package options
 
 import (
+	"crypto/tls"
 	"fmt"
 	"os"
 	"time"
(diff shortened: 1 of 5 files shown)


tesshuflower:
re-opened this as: https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/820

Issue #816, opened 2026-03-26
Agent reconnection regression in v0.33.0: slow retry on DuplicateServerError causes permanent NotReady nodes


IvanHunters:
## Problem Description

In konnectivity agent v0.33.0, agents fail to reconnect after connection loss, leaving Kubernetes nodes in permanent `NotReady` state.

## Root Cause

In the `sync()` method, when `DuplicateServerError` occurs during reconnection with `clientsCount < serverCount`, the code adds a significant delay:

```go
if dse, ok := err.(*DuplicateServerError); ok {
    clientsCount := cs.ClientsCount()
    if serverCount != 0 && clientsCount >= serverCount {
        duration = backoff.Step()
    } else {
        backoff = cs.resetBackoff()
        duration = wait.Jitter(backoff.Duration, backoff.Jitter)  // ~1s delay
    }
}
```

## Why This Breaks Reconnection

The agent maintains connections to multiple proxy server pods (e.g., 5 pods in a DaemonSet). When a client disconnects:

1. `RemoveClient(serverID_1)` removes the client from the map
2. `connectOnce()` attempts to create a new client
3. `newAgentClient()` uses **DNS load balancing** to connect to a random proxy pod
4. Probability of connecting to the correct pod = **1/5 (20%)**
5. If wrong pod → `DuplicateServerError` (serverID already exists in map)
6. **~1s delay** before next retry
7. Repeat steps 3-6 until correct pod is randomly selected

**Impact:**
- Average reconnection time: ~5 seconds per client
- If multiple clients disconnect simultaneously (e.g., VM restart), cascading failures occur
- Reconnection attempts accumulate, system never fully recovers

## Reproduction Steps

1. Deploy Kubernetes cluster with konnectivity v0.33.0
2. Wait for all agents to connect (5/5 connections established)
3. Restart one worker node VM (causes agent pod disconnect)
4. Observe: Agent never reconnects, node stays `NotReady` permanently

**Expected:** Agent reconnects within seconds  
**Actual:** Agent continuously gets `DuplicateServerError`, each retry delayed by ~1s

## Evidence

Logs showing the issue:
```
transport is closing
dial error: connection error: desc = "error reading server preface: EOF"
cannot connect once
duplicate server (repeating every ~1s)
```

No "Agent connected" message appears after initial connection loss.

## Impact

This regression affects any Kubernetes cluster using:
- Konnectivity v0.33.0
[...]

PR #817, opened 2026-03-26
fix(agent): remove reconnection delay on DuplicateServerError

IvanHunters:
## Description

This PR fixes a regression in v0.33.0 where agents fail to reconnect after connection loss, leaving Kubernetes nodes in permanent `NotReady` state.

## Problem

In v0.33.0, when `DuplicateServerError` occurs during reconnection with `clientsCount < serverCount`, the code adds a ~1s delay:

```go
} else {
    backoff = cs.resetBackoff()
    duration = wait.Jitter(backoff.Duration, backoff.Jitter)  // ~1s delay
}
```

Since `newAgentClient()` uses DNS load balancing, it connects to a random proxy pod. When a client disconnects, the reconnection attempt has only ~20% chance (1/5 pods) of selecting the correct pod. Wrong selections trigger `DuplicateServerError`, and the 1s delay significantly slows reconnection.

**Impact:** Average reconnection time increases from ~500ms to ~5s per client. With multiple simultaneous disconnects (e.g., VM restart), cascading failures prevent full recovery.

## Solution

Remove the else block that adds delay. Let `duration` remain 0 (default value), enabling immediate retry through `time.Sleep(0)`. This allows fast reconnection via DNS load balancing without creating a tight loop (the loop already includes gRPC dial time and network round trips).

## Context

The release-0.33 branch already contains one fix from v0.34.0 (moving `lastReceivedServerCount` update to after successful `AddClient()`). This PR adds the second critical fix - removing the delay on reconnection attempts.

Together, these changes restore the fast reconnection behavior from v0.28.6 and align with the complete fix in v0.34.0.

## Testing

Verified in production cluster:
- Before fix: Nodes remain NotReady permanently after VM restart (v0.33.0)
- After fix: Expected behavior matches v0.34.0 - agents reconnect within seconds

Related: #816

### DIFF
--- a/pkg/agent/clientset.go
+++ b/pkg/agent/clientset.go
@@ -226,10 +226,10 @@ func (cs *ClientSet) sync() {
 				klog.V(4).InfoS("duplicate server", "serverID", dse.ServerID, "serverCount", serverCount, "clientsCount", clientsCount)
 				if serverCount != 0 && clientsCount >= serverCount {
 					duration = backoff.Step()
-				} else {
-					backoff = cs.resetBackoff()
-					duration = wait.Jitter(backoff.Duration, backoff.Jitter)
 				}
+				// When clientsCount < serverCount, we need a new connection.
+				// Leave duration at 0 to retry immediately without delay,
+				// allowing fast reconnection via DNS load balancing.
 			} else {
 				klog.ErrorS(err, "cannot connect once")
 				duration = backoff.Step()
--- a/pkg/agent/clientset.go
+++ b/pkg/agent/clientset.go
@@ -100,6 +100,14 @@ type ClientSet struct {
 	// and the agent figures out from these observations how many
 	// agent-to-proxy-server connections it should maintain.
 	serverCountSource string
+
+	// syncImmediatelyOnDuplicate controls the retry delay when a reconnection
+	// attempt lands on a proxy server we are already connected to
+	// (DuplicateServerError) while we still need more connections. When true,
+	// the agent retries immediately instead of waiting a sync interval, which
+	// speeds up reconnection behind DNS load balancing at the cost of more
+	// frequent connection attempts. Defaults to false (the previous behavior).
+	syncImmediatelyOnDuplicate bool
 }
 
 func (cs *ClientSet) ClientsCount() int {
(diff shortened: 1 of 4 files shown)


pkg/agent/clientset.go
```diff
@@ -226,10 +226,10 @@ func (cs *ClientSet) sync() {
 				klog.V(4).InfoS("duplicate server", "serverID", dse.ServerID, "serverCount", serverCount, "clientsCount", clientsCount)
 				if serverCount != 0 && clientsCount >= serverCount {
 					duration = backoff.Step()
-				} else {
-					backoff = cs.resetBackoff()
-					duration = wait.Jitter(backoff.Duration, backoff.Jitter)
 				}
+				// When clientsCount < serverCount, we need a new connection.
```
cheftako: This has potential negative impact if something else is going on and now we generate a ton of connection requests. Happy to have this added as an option but I think it would be safer to add a flag for this behavior. We should also default to the existing behavior so we are backward compatible.

cheftako:
Can you rebase on the latest?
Also the PR cannot be merged until you [sign the CLA](https://github.com/kubernetes/community/blob/main/CLA.md#how-do-i-sign)

pkg/agent/clientset.go
ipochi: I second this. With sync-forever enabled I've seen the agent connecting to a new proxy-server instantly.

IvanHunters:
Thanks for the reviews, @cheftako and @ipochi.

I've reworked the change based on your feedback:

- The delay removal is now **opt-in** behind a new `--sync-immediately-on-duplicate` flag, defaulting to `false`, so the existing behavior is unchanged and backward compatible.
- To address the concern about generating too many connection requests: when enabled, immediate retries only happen on a known connection deficit (`serverCount != 0 && clientsCount < serverCount`) and are **bounded to `serverCount` consecutive attempts** before falling back to the jittered sync interval. The counter resets on any progress, so a permanent deficit or an unknown (zero) server count cannot become an unbounded zero-delay reconnect loop.
- The delay decision is extracted from `sync()` into a `nextResyncDuration` helper, now covered by unit tests (per-branch backward compatibility, the bounded immediate-retry budget, and preserved exponential backoff).

I've also rebased onto the latest `release-0.33` and squashed to a single commit. I'll get the CLA signed.

PR #821, opened 2026-04-24
server: avoid selecting draining backends

ipochi:
Track non-draining backend candidates so selection avoids scanning and
does not route new requests to draining backends when healthy choices
exist.

```
go test ./pkg/server -count=1
go test ./pkg/server -race -run Test(DefaultBackendManager_GetRandomBackend_LazilyEvictsDrainingCandidate|DestHostBackendManager_Backend_DrainingFallback|DestHostBackendManager_Backend_DistributesAcrossNonDrainingCandidates|DestHostBackendManager_NonDrainingCacheTracksCurrentBackends)$'
-count=1
```


cheftako:

PR #822, opened 2026-04-28
konnectivity-client: extract newSingleStreamTunnel helper

ipochi:
This is a pure refactor with no behavioral change. Factor Proxy() stream setup and serve() goroutine launch out of
CreateSingleUseGrpcTunnelWithContext into an internal newSingleStreamTunnel helper that takes a ProxyServiceClient and a closeFn.

The helper invokes closeFn from the serve() goroutine on exit via the clientConn interface instead of unconditionally calling grpcConn.Close(). For the existing single-use path, closeFn is c.Close, preserving today's behavior: tunnelCtx still bounds both Proxy() establishment and the serve() goroutine, and the *grpc.ClientConn is still closed when serve() returns.

This separation prepares for a future reusable tunnel API that will share one *grpc.ClientConn across many per-dial streams. In that path, closeFn will cancel only the per-dial stream context rather than closing the shared conn. No such caller is added in this commit.

Add a closerFunc adapter so a func() error satisfies the existing clientConn interface without introducing a new abstraction.

### DIFF
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -45,6 +45,32 @@ type Tunnel interface {
 	Done() <-chan struct{}
 }
 
+// ReusableTunnel is a Tunnel whose DialContext may be called many times.
+// Each returned net.Conn is independent. The caller is responsible for
+// calling Close to release the underlying gRPC ClientConn.
+//
+// A ReusableTunnel owns a single underlying *grpc.ClientConn whose lifetime
+// matches the tunnel's; per-dial work happens on new gRPC Proxy streams over
+// that shared connection rather than new gRPC ClientConn connections.
+//
+// Done semantics differ subtly from a single-use Tunnel: Done is closed only
+// after Close has been called and all in-flight per-dial child streams have
+// drained. It does not fire on remote unreachability alone; callers detect
+// transport failure via DialContext errors (typed via GetDialFailureReason)
+// and decide whether to Close and rebuild.
+type ReusableTunnel interface {
+	Tunnel
+	// Close releases the tunnel's resources, including the underlying
+	// gRPC ClientConn. In-flight DialContext calls are cancelled.
+	//
+	// Close blocks until all per-dial child streams have drained and Done
+	// has fired. Concurrent Close calls all observe the same post-condition:
+	// every caller returns only after teardown is complete. The first
+	// caller's return value carries any error from closing the underlying
+	// ClientConn; subsequent callers return nil.
+	Close() error
+}
+
 type dialResult struct {
 	err    *dialFailure
 	connid int64
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -179,20 +179,43 @@ func CreateSingleUseGrpcTunnelWithContext(createCtx, tunnelCtx context.Context, 
 		return nil, err
 	}
 
-	grpcClient := client.NewProxyServiceClient(c)
+	proxyClient := client.NewProxyServiceClient(c)
 
-	stream, err := grpcClient.Proxy(tunnelCtx)
+	tunnel, err := newSingleStreamTunnel(tunnelCtx, proxyClient, c.Close)
 	if err != nil {
 		c.Close()
 		return nil, err
 	}
 
-	tunnel := newUnstartedTunnel(stream, c)
+	return tunnel, nil
+}
+
+// newSingleStreamTunnel creates a Tunnel backed by a single Proxy stream on
+// proxyClient. closeFn is invoked from the serve() goroutine on exit (via the
+// clientConn.Close interface), in place of the previous unconditional
+// grpcConn.Close().
+//
+// tunnelCtx bounds both Proxy() RPC establishment and the resulting stream's
+// serve goroutine, matching the existing CreateSingleUseGrpcTunnelWithContext
+// semantics. Callers wanting to bound only establishment must do so themselves.
+func newSingleStreamTunnel(tunnelCtx context.Context, proxyClient client.ProxyServiceClient, closeFn func() error) (Tunnel, error) {
+	stream, err := proxyClient.Proxy(tunnelCtx)
+	if err != nil {
+		return nil, err
+	}
+
+	tunnel := newUnstartedTunnel(stream, closerFunc(closeFn))
 
 	go tunnel.serve(tunnelCtx)
 
 	return tunnel, nil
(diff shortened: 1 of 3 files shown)


cheftako:

konnectivity-client/pkg/client/client.go
```diff
@@ -179,21 +205,229 @@ func CreateSingleUseGrpcTunnelWithContext(createCtx, tunnelCtx context.Context,
...
+// reconnect, or load balancing. Callers detect transport failure via
+// DialContext errors (typed via GetDialFailureReason) and decide whether
+// to Close and rebuild. For TLS cert reload or load balancing across proxy
+// replicas, configure the underlying gRPC ClientConn with appropriate
+// DialOptions, such as dynamic TransportCredentials or
+// grpc.WithDefaultServiceConfig.
+//
+// MaxConcurrentStreams on both client and server bounds the per-tunnel
+// parallelism; phase 1 does not change defaults.
+func CreateGRPCTunnel(ctx context.Context, address string, opts ...grpc.DialOption) (ReusableTunnel, error) {
```
cheftako, line 284: nit: The existing single-use constructors use `Grpc` casing (e.g. `CreateSingleUseGrpcTunnel`, `CreateSingleUseGrpcTunnelWithContext`, and internal type `reusableGrpcTunnel`). While `GRPC` is standard Go initialism style, `CreateReusableGrpcTunnel` or keeping `CreateGRPCTunnel` is fine — just noting the slight casing inconsistency with the existing exported constructor names.

konnectivity-client/pkg/client/client.go
```diff
@@ -179,21 +205,229 @@ func CreateSingleUseGrpcTunnelWithContext(createCtx, tunnelCtx context.Context,
...
+		if res.err != nil {
+			streamCancel()
+			r.children.Done()
+			// Typed dialFailure so callers' shouldInvalidateTunnel can
+			// distinguish stream-setup failure from per-dial backend errors.
+			metrics.Metrics.ObserveDialFailure(metrics.DialFailureStreamSetup)
+			return nil, &dialFailure{res.err.Error(), metrics.DialFailureStreamSetup}
+		}
+		inner = res.t
+	case <-requestCtx.Done():
```
cheftako, line 364: Draining `resCh` here to ensure that a late-completing stream establishment goroutine gets its `inner.Done()` tracked before `children.Done()` fires is very clean. I verified locally with `go test -race` and test 5b (`TestReusableTunnel_LateEstablishmentWaited`) confirms this avoids races when `Close()` is called right after request context expiration.

konnectivity-client/pkg/client/client.go
```diff
@@ -179,21 +205,229 @@ func CreateSingleUseGrpcTunnelWithContext(createCtx, tunnelCtx context.Context,
...
+	// inner.DialContext honors requestCtx for the DIAL_REQ/DIAL_RSP exchange.
+	// Errors returned here are typed dialFailures from the existing path
+	// (DialFailureEndpoint, DialFailureContext, DialFailureTimeout, ...).
+	//
+	// Some inner.DialContext error paths (notably Send failure inside the
+	// existing grpcTunnel.dialContext) do not call closeTunnel themselves.
+	// We must cancel the stream context on every error to guarantee the
+	// per-dial stream's serve goroutine exits, so children.Done can fire.
+	conn, err := inner.DialContext(requestCtx, protocol, address)
+	if err != nil {
```
cheftako, line 402: Good catch calling `streamCancel()` on `inner.DialContext` errors here — since `grpcTunnel.dialContext` doesn't close the stream on `t.Send` failures, without this the serve goroutine would remain blocked in `Recv()` and `Close()` would hang on `children.Wait()`.

konnectivity-client/pkg/client/client.go
```diff
@@ -179,21 +205,229 @@ func CreateSingleUseGrpcTunnelWithContext(createCtx, tunnelCtx context.Context,
...
+		streamCancel()
+		return nil, err
+	}
+	return conn, nil
+}
+
+func (r *reusableGrpcTunnel) Done() <-chan struct{} { return r.done }
+
+func (r *reusableGrpcTunnel) Close() error {
+	r.acceptMu.Lock()
```
cheftako, line 412: The mutex synchronization around `closing` and `children.Add(1)` in `DialContext` vs `Close()` nicely handles the `sync.WaitGroup` `Add`/`Wait` ordering requirements.

konnectivity-client/pkg/client/metrics/metrics.go
```diff
@@ -64,6 +64,13 @@ const (
 	DialFailureTunnelClosed DialFailureReason = "tunnelclosed"
 	// DialFailureAlreadyStarted indicates that a single-use tunnel dialer was already used once.
 	DialFailureAlreadyStarted DialFailureReason = "tunnelstarted"
+	// DialFailureStreamSetup indicates that opening a new Proxy stream on a
+	// reusable tunnel's shared gRPC connection failed before any DIAL_REQ was
+	// sent. Distinct from DialFailureEndpoint (backend dial) and
+	// DialFailureContext (caller-side cancellation) so callers can selectively
+	// invalidate the shared transport on stream-setup failures while leaving
+	// it intact on backend or caller-class failures.
+	DialFailureStreamSetup DialFailureReason = "streamsetup"
```
cheftako, line 73: Adding `DialFailureStreamSetup` makes a lot of sense so callers (like the kube-apiserver egress connector) can distinguish connection/stream setup failures (which warrant invalidating the shared tunnel) from endpoint-level or context dial errors.

konnectivity-client/pkg/client/client.go
```diff
@@ -179,21 +205,229 @@ func CreateSingleUseGrpcTunnelWithContext(createCtx, tunnelCtx context.Context,
...
+	if r.closing {
+		r.acceptMu.Unlock()
+		// Already closing. Block until the first caller finishes draining
+		// so all callers observe the same post-condition: resources
+		// released, Done fired. Documented on ReusableTunnel.Close.
+		<-r.done
+		return nil
+	}
+	r.closing = true
+	r.acceptMu.Unlock()
```
cheftako, line 422: It would be hard for Unlock to fail to fire right now as only simple r.closing manipulations exist right now. However if anyone adds any logic inside the lock/unlock block that may become a problem. It would be nice to extract a setClosing method which uses a defer unlock to ensure we do not fail to unlock.

cheftako (approved):
Would be nice to follow up on the unlock in a follow up PR.

PR #823, opened 2026-04-28
konnectivity-client: add reusable gRPC tunnel

ipochi:
Add `ReusableTunnel` and `CreateGRPCTunnel` so callers can reuse one gRPC `ClientConn` and open one `Proxy` stream per `DialContext`, instead of creating a new gRPC connection
  per proxied dial.

 #### Changes

  - Add `ReusableTunnel` interface with explicit `Close` / `Done` drain semantics.
  - Add `reusableGrpcTunnel` implementation backed by one shared gRPC `ClientConn`.
  - Add `DialFailureStreamSetup` so callers can distinguish shared-transport stream setup failures from backend dial failures.
  - Refactor existing single-use tunnel setup through an internal helper.
  - Add reusable tunnel tests covering reuse, close/drain behavior, failure classification, races, and leaks.

  #### Notes

  This does not change the konnectivity wire protocol, server, agent, existing `Tunnel` interface, or single-use tunnel behavior. Reconnect and rebuild policy remain caller-
  owned.

### DIFF
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -45,6 +45,32 @@ type Tunnel interface {
 	Done() <-chan struct{}
 }
 
+// ReusableTunnel is a Tunnel whose DialContext may be called many times.
+// Each returned net.Conn is independent. The caller is responsible for
+// calling Close to release the underlying gRPC ClientConn.
+//
+// A ReusableTunnel owns a single underlying *grpc.ClientConn whose lifetime
+// matches the tunnel's; per-dial work happens on new gRPC Proxy streams over
+// that shared connection rather than new gRPC ClientConn connections.
+//
+// Done semantics differ subtly from a single-use Tunnel: Done is closed only
+// after Close has been called and all in-flight per-dial child streams have
+// drained. It does not fire on remote unreachability alone; callers detect
+// transport failure via DialContext errors (typed via GetDialFailureReason)
+// and decide whether to Close and rebuild.
+type ReusableTunnel interface {
+	Tunnel
+	// Close releases the tunnel's resources, including the underlying
+	// gRPC ClientConn. In-flight DialContext calls are cancelled.
+	//
+	// Close blocks until all per-dial child streams have drained and Done
+	// has fired. Concurrent Close calls all observe the same post-condition:
+	// every caller returns only after teardown is complete. The first
+	// caller's return value carries any error from closing the underlying
+	// ClientConn; subsequent callers return nil.
+	Close() error
+}
+
 type dialResult struct {
 	err    *dialFailure
 	connid int64
--- a/konnectivity-client/pkg/client/client.go
+++ b/konnectivity-client/pkg/client/client.go
@@ -179,20 +179,43 @@ func CreateSingleUseGrpcTunnelWithContext(createCtx, tunnelCtx context.Context, 
 		return nil, err
 	}
 
-	grpcClient := client.NewProxyServiceClient(c)
+	proxyClient := client.NewProxyServiceClient(c)
 
-	stream, err := grpcClient.Proxy(tunnelCtx)
+	tunnel, err := newSingleStreamTunnel(tunnelCtx, proxyClient, c.Close)
 	if err != nil {
 		c.Close()
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


cheftako:

konnectivity-client/pkg/client/client.go
```diff
@@ -179,21 +205,229 @@ func CreateSingleUseGrpcTunnelWithContext(createCtx, tunnelCtx context.Context,
...
 
+// closerFunc adapts a func() error to the clientConn interface (which only
+// requires Close() error).
+type closerFunc func() error
+
+func (c closerFunc) Close() error { return c() }
+
+// reusableGrpcTunnel implements ReusableTunnel with one shared *grpc.ClientConn
+// and one Proxy stream per DialContext call.
+type reusableGrpcTunnel struct {
```
cheftako, line 248: Please add compile-time interface checks here, e.g.:

```suggestion
var (
	_ ReusableTunnel = &reusableGrpcTunnel{}
	_ Tunnel         = &reusableGrpcTunnel{}
)

type reusableGrpcTunnel struct {
```

konnectivity-client/pkg/client/client.go
```diff
@@ -179,21 +205,229 @@ func CreateSingleUseGrpcTunnelWithContext(createCtx, tunnelCtx context.Context,
...
+	// (sync.WaitGroup requires that positive Adds happen-before any Wait
+	// when the counter is zero).
+	acceptMu sync.Mutex
+	closing  bool
+
+	children sync.WaitGroup
+	done     chan struct{}
+}
+
+// CreateGRPCTunnel creates a reusable Tunnel to a konnectivity proxy at
```
cheftako, line 269: nit: It might be helpful to briefly clarify in the docstring that `ctx` is only used to bound the initial `grpc.DialContext` connection establishment, since `reusableGrpcTunnel` creates an independent internal context for the tunnel's ongoing lifetime and relies on `Close()` for teardown.

konnectivity-client/pkg/client/client.go
```diff
@@ -179,21 +205,229 @@ func CreateSingleUseGrpcTunnelWithContext(createCtx, tunnelCtx context.Context,
...
+// reconnect, or load balancing. Callers detect transport failure via
+// DialContext errors (typed via GetDialFailureReason) and decide whether
+// to Close and rebuild. For TLS cert reload or load balancing across proxy
+// replicas, configure the underlying gRPC ClientConn with appropriate
+// DialOptions, such as dynamic TransportCredentials or
+// grpc.WithDefaultServiceConfig.
+//
+// MaxConcurrentStreams on both client and server bounds the per-tunnel
+// parallelism; phase 1 does not change defaults.
+func CreateGRPCTunnel(ctx context.Context, address string, opts ...grpc.DialOption) (ReusableTunnel, error) {
```
cheftako, line 284: nit: Naming question — the existing constructors in this package use `CreateSingleUseGrpcTunnel` / `CreateSingleUseGrpcTunnelWithContext` (mixed-case `Grpc`), whereas this introduces `CreateGRPCTunnel` (all-caps `GRPC`).

I know `GRPC` aligns better with standard Go acronym conventions, but I wanted to check if the naming divergence from the existing `CreateSingleUseGrpc*` helpers was intentional.

konnectivity-client/pkg/client/reusable_tunnel_test.go
```diff
@@ -0,0 +1,974 @@
...
+		t.Fatalf("tunnel Close: %v", err)
+	}
+	select {
+	case <-tun.Done():
+	case <-time.After(5 * time.Second):
+		t.Fatal("Done did not fire after Close")
+	}
+}
+
+// 12. Stream establishment failure is typed DialFailureStreamSetup and observed.
```
cheftako, line 827: I think the numbered test comments jump from test 9 (`TestReusableTunnel_DoneRequiresClose`) directly to 12 (`TestReusableTunnel_StreamSetupFailureTyped`), skipping 10 and 11 (and test 6 `TestReusableTunnel_CloseErrorReturnedToFirstCallerOnly` has no number prefix). Are we missing tests? Would be good to keep the numbering consistent so as not to cause confusion.

PR #825, opened 2026-05-01
Switch tests to current supported K8s versions.

cheftako:
Switch tests to current supported K8s versions.


justinsb (approved):
/approve
/lgtm

PR #830, opened 2026-05-01
Prevent dependabot from updating konnectivity-client

cheftako:
konnectivity-client should be in line with the oldest supported K/K. 
It should be ok for K/K to push newer libraries than ANP. 
That is what we expect to happen on newer K/K releases. 
ANP should not be forcing upgrades on K/K.


justinsb (approved):
/approve
/lgtm

PR #831, opened 2026-05-03
Upgrade to go v1.26

cheftako:
Prompt:
please upgrade the top level go.mod to use go version 1.26.0. Ensure all the tests pass. Do not make changes in
   konnectivity-client.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/apiserver-network-proxy
 
-go 1.24.7
+go 1.26.2
 
 require (
 	github.com/google/uuid v1.6.0
--- a/pkg/agent/metrics/metrics.go
+++ b/pkg/agent/metrics/metrics.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package metrics
+package metrics //nolint:revive
 
 import (
 	"strconv"
--- a/.github/workflows/golangci-lint.yaml
+++ b/.github/workflows/golangci-lint.yaml
@@ -11,8 +11,8 @@ permissions:
   contents: read
 
 env:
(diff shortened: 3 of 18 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/apiserver-network-proxy
 
-go 1.24.7
+go 1.26.0
```
justinsb: Nit: don't we have 1.26.2 now?

pkg/agent/metrics/metrics.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package metrics
```
justinsb, line 17: Not sure why we're doing this.. is it related to 1.26?

cheftako: [revive] linter with the error "var-naming: avoid package names that conflict with Go standard library package names"

[Variable naming](https://github.com/mgechev/revive/blob/HEAD/RULES_DESCRIPTIONS.md#var-naming) which leads to [package naming](https://github.com/mgechev/revive/blob/HEAD/RULES_DESCRIPTIONS.md#package-naming)

cheftako: Can add //nolint:revive

go.mod
cheftako: Fixed

justinsb (approved):
/approve
/lgtm

pkg/agent/metrics/metrics.go
cheftako: It's related to the GOLANGCI_LINT_VERSION: v2.9.0. 

2.9.0 was the minimum to support 1.26

Issue #855, opened 2026-05-28
HTTPConnect pending backend dials can accumulate while agents remain connected


inerplat:
## Summary

We observed a production failure mode where the apiserver-network-proxy server was still reporting connected backends, but HTTPConnect frontend connections and `pending_backend_dials` kept increasing. In that state, kube-apiserver egress traffic such as admission webhook calls started failing with 500/timeout responses.

This appears to happen when a server-to-agent backend stream stops making forward progress while the backend is still considered available. New HTTP CONNECT dials are accepted by the proxy server, added to the pending dial table, and then can remain pending without a bounded lifetime.

## What's happened

Screenshot 1 shows the HTTPConnect connection count for one kube-apiserver starting to increase monotonically around 2026-04-23, reaching about 1.55K in the displayed window:

<img width="923" height="303" alt="Image" src="https://github.com/user-attachments/assets/8159a6b7-f3f4-47ae-a582-93944a2b95a0" />

Screenshot 2 shows the same kube-apiserver's HTTPConnect connection count later reaching about 27.7K by 2026-05-19, while gRPC backend connections stayed around 100:

<img width="931" height="312" alt="Image" src="https://github.com/user-attachments/assets/7abd17b1-5d3e-4e56-b489-94fcb3492a4e" />

## Observed Behavior

- `konnectivity_network_proxy_server_http_connections` increased monotonically for one kube-apiserver.
- `konnectivity_network_proxy_server_pending_backend_dials` increased and did not promptly drain.
- `konnectivity_network_proxy_server_grpc_connections{service_method="Connect"}` and ready backend metrics stayed non-zero.
- Admission webhook requests through the kube-apiserver egress path failed with 500/timeout responses.
- Restarting the affected proxy component cleared the accumulated state.

## Reproduction Used To Validate The Failure Mode

We validated the failure mode with a natural backpressure scenario:

1. Run apiserver-network-proxy server in HTTPConnect mode.
2. Configure kube-apiserver egress traffic to go through konnectivity.
3. Create a Service backend that accepts HTTP headers but does not read the request body.
4. Send large POST requests to that Service through the kube-apiserver service proxy path.
5. In parallel, send admission webhook dry-run requests.
6. Observe HTTPConnect connections and pending backend dials increase while the backend still appears connected.

This reproduction models a class of operational cases where an in-cluster endpoint keeps the TCP connection alive but stops consuming request body data quickly enough.

## Why This Matters

[...]

PR #856, opened 2026-05-28
Add configurable timeout for pending backend dials

inerplat:
## Summary

This PR adds an opt-in `--backend-dial-timeout` option to the proxy server.

When configured, the proxy server bounds how long a frontend dial can remain pending while sending `DIAL_REQ` to the selected backend or waiting for `DIAL_RSP`. On timeout, the pending dial is removed and the frontend receives a bounded error.

The default value is `0`, so existing behavior is preserved unless operators explicitly enable the timeout.

## Motivation

In HTTPConnect deployments used by kube-apiserver egress-selector setups, we observed a failure mode where backend connections remained registered while new HTTP CONNECT dials accumulated in `pending_backend_dials`. This can surface as admission webhook failures when the server-to-agent path stops making forward progress but the backend is still considered available.

The new option gives operators a guardrail against unbounded pending dial accumulation without changing default behavior.

## Changes

- Add `--backend-dial-timeout` to proxy-server options.
- Keep the default disabled with `0`.
- Apply the timeout while sending frontend `DIAL_REQ` packets to the selected backend.
- Apply the timeout while waiting for a backend `DIAL_RSP` before the HTTPConnect tunnel is established.
- Remove pending dial entries on timeout.
- Add a `backend_dial_timeout` dial failure metric reason.
- Add a reduced HTTPConnect regression test for blocked backend dial send.

## Compatibility

The default behavior is unchanged because `--backend-dial-timeout` defaults to `0`.

Operators can opt in with a value such as:

`--backend-dial-timeout=30s`

## Testing

Ran:

`go test ./cmd/server/app/options ./pkg/server -count=1`

Added test:

`TestHTTPConnectTunnelBlockedBackendDialSendTimesOutAndCleansPendingDial`

The test starts the HTTPConnect tunnel handler, blocks backend `Send(DIAL_REQ)`, verifies the dial enters pending state, and then verifies the configured timeout returns HTTP 504 and cleans up the pending dial.

## Related Issue

https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/855

### DIFF
--- a/cmd/server/app/options/options.go
+++ b/cmd/server/app/options/options.go
@@ -122,6 +122,8 @@ type ProxyRunOptions struct {
 	NeedsKubernetesClient bool
 	// Graceful shutdown timeout duration
 	GracefulShutdownTimeout time.Duration
+	// Backend dial timeout duration for requests from the server to backend agents.
+	BackendDialTimeout time.Duration
 }
 
 func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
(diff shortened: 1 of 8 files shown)


inerplat:
@cheftako 
The golangci-lint failure has been fixed. I also found and fixed an additional issue where a timed-out backend dial send could leave the backend connection wedged.

Could you please retrigger the CI for the latest commit?

Thanks.

cmd/server/app/options/options.go
```diff
@@ -122,6 +122,8 @@ type ProxyRunOptions struct {
 	NeedsKubernetesClient bool
 	// Graceful shutdown timeout duration
 	GracefulShutdownTimeout time.Duration
+	// Backend dial timeout duration for frontend dial requests
```
cheftako: [Nice to have] Can we fix the comment to be less confusing? Eg (Dial timeout for requests from server to agent)

cmd/server/app/options/options.go
```diff
@@ -164,6 +166,7 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.StringVar(&o.LeaseNamespace, "lease-namespace", o.LeaseNamespace, "The namespace where lease objects are managed by the controller.")
 	flags.StringVar(&o.LeaseLabel, "lease-label", o.LeaseLabel, "The labels on which the lease objects are managed.")
 	flags.DurationVar(&o.GracefulShutdownTimeout, "graceful-shutdown-timeout", o.GracefulShutdownTimeout, "Timeout duration for graceful shutdown of the server. The server will wait for active connections to close before forcefully terminating. Set to 0 to disable graceful shutdown (default: 0).")
+	flags.DurationVar(&o.BackendDialTimeout, "backend-dial-timeout", o.BackendDialTimeout, "Timeout duration for sending DIAL_REQ packets to backend agent streams and waiting for DIAL_RSP. Set to 0 to disable timeout.")
```
cheftako: [Nice to have] Would be good to indicate that 0 is the default.

cheftako (approved):
Looks good

PR #866, opened 2026-06-22
Restore proper non-graceful shutdown behavior

twz123:
PR #794 attempted to retain the old non-graceful shutdown behavior if the graceful shutdown timeout was zero/not specified by converting the deferred close calls into an if block after the server received its shutdown signal. However, the PR didn't account for the fact that those deferred calls were executed in reverse order, having the opposite effect: the frontend server waited indefinitely for all frontend connections to close.

Reorder the close calls so that they are called in the same order as when they were called via the defer statements. Also, pass the `Run` method's context to the call to `frontendShutdown`, as this is the same context that was passed to the wrapped HTTP server's Shutdown method previously.

Fixes:

* #794


cheftako:
@ipochi Can you look at this?

ipochi:
I'm taking a  look at this one.

PR #868, opened 2026-06-23
Make lease labels a real label selector for the agent

twz123:
The lease labels are already converted into a label selector anyways, so allowing the full label selector syntax removes an artificial
constraint.

### DIFF
--- a/cmd/agent/app/options/options.go
+++ b/cmd/agent/app/options/options.go
@@ -17,6 +17,7 @@ limitations under the License.
 package options
 
 import (
+	"cmp"
 	"fmt"
 	"net"
 	"net/url"
@@ -27,6 +28,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/spf13/pflag"
 	"google.golang.org/grpc"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog/v2"
 
@@ -87,8 +89,8 @@ type GrpcProxyAgentOptions struct {
 	CountServerLeases bool
 	// Namespace where lease objects are managed.
 	LeaseNamespace string
-	// Labels on which lease objects are managed.
-	LeaseLabel string
+	// Label selector for leases to take into account when counting servers.
+	LeaseLabelSelector labels.Selector
 	// ServerCountSource describes how server counts should be combined.
 	ServerCountSource string
 	// Path to kubeconfig (used by kubernetes client for lease listing)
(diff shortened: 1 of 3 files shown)


cmd/agent/app/options/options.go
```diff
@@ -140,7 +142,10 @@ func (o *GrpcProxyAgentOptions) Flags() *pflag.FlagSet {
 	flags.IntVar(&o.XfrChannelSize, "xfr-channel-size", 150, "Set the size of the channel for transferring data between the agent and the proxy server.")
 	flags.BoolVar(&o.CountServerLeases, "count-server-leases", o.CountServerLeases, "Enables lease counting system to determine the number of proxy servers to connect to.")
 	flags.StringVar(&o.LeaseNamespace, "lease-namespace", o.LeaseNamespace, "Namespace where lease objects are managed.")
-	flags.StringVar(&o.LeaseLabel, "lease-label", o.LeaseLabel, "The labels on which the lease objects are managed.")
+	flags.VarP(labelSelectorFlag{&o.LeaseLabelSelector}, "lease-label-selector", "", "Label selector for leases to take into account when counting servers.")
```
cheftako, line 145: The default may be empty but it would be good to be consistent and pass in o.LeaseLabelSelector as the default value. If at some point we decide we should have a default, the mechanism for setting it is consistent with other fields.

cmd/agent/app/options/options_test.go
```diff
@@ -51,6 +50,11 @@ func TestDefaultServerOptions(t *testing.T) {
 	assertDefaultValue(t, "WarnOnChannelLimit", defaultAgentOptions.WarnOnChannelLimit, false)
 	assertDefaultValue(t, "SyncForever", defaultAgentOptions.SyncForever, false)
 	assertDefaultValue(t, "XfrChannelSize", defaultAgentOptions.XfrChannelSize, 150)
+	assertDefaultValue(t, "CountServerLeases", defaultAgentOptions.CountServerLeases, false)
+	assertDefaultValue(t, "LeaseNamespace", defaultAgentOptions.LeaseNamespace, "kube-system")
+	assertDefaultValue(t, "LeaseLabelSelector", defaultAgentOptions.LeaseLabelSelector.String(), "k8s-app=konnectivity-server")
```
cheftako, line 55: Please make sure your tests cover at least as much as our old tests. Eg https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/util/util_test.go#L23 has tests for "app=myapp,env=prod,invalid" and " key = value , another = test "

PR #871, opened 2026-07-01
Decouple send/recv in data->control in agent

cheftako:
APIServer Network Proxy (ANP) is responsible for tunneling traffic from the Kubernetes APIServer (KAS) to the cluster and responses back. To do this traffic goes from KAS to the ANP Server. From the ANP server traffic is sent to the ANP Agent. From the ANP Agent traffic is sent to the final destination. Responses are sent from the destination to the ANP Agent. From the ANP Agent the response is sent to the ANP Server. From the ANP Server the response is sent to the KAS. When going through the ANP Server and the ANP Agent it is important that the Send method to push the traffic on the next segment does not block the receive method from the last segment. This is achieved by putting a transfer buffer or channel between these methods. This channel or transfer buffer exists in the ANP Server in both directions. However it only exists in the ANP Agent when coming from the ANP Server and headed toward the final destination. The read from the destination can be found in pkg/agent/client.go at about line 553. The send to the ANP Server can be found at about line 574. Please add a channel to stop this send from blocking the read. A declaration of such a channel or transfer buffer can be found at pkg/agent/client.go line 385. The usage of the channel can be found at pkg/agent/client.go line 606.

### DIFF
--- a/pkg/agent/client.go
+++ b/pkg/agent/client.go
@@ -50,6 +50,7 @@ type endpointConn struct {
 	connID    int64
 	cleanFunc func()
 	dataCh    chan []byte
+	sendCh    chan []byte
 	cleanOnce sync.Once
 	warnChLim bool
 	dialDone  chan struct{}
@@ -59,8 +60,7 @@ func (e *endpointConn) cleanup() {
 	e.cleanOnce.Do(e.cleanFunc)
 }
 
-func (e *endpointConn) send(msg []byte) {
-	// TODO (cheftako@): Get perf test working and compare this solution with a lock based solution.
+func (e *endpointConn) sendViaDataChannel(msg []byte) {
 	defer func() {
 		// Handles the race condition where we write to a closed channel
 		if err := recover(); err != nil {
@@ -383,9 +383,11 @@ func (a *Client) Serve() {
 
 			connID := atomic.AddInt64(&a.nextConnID, 1)
 			dataCh := make(chan []byte, a.cs.xfrChannelSize)
+			sendCh := make(chan []byte, a.cs.xfrChannelSize)
 			dialDone := make(chan struct{})
 			eConn := &endpointConn{
 				dataCh:    dataCh,
+				sendCh:    sendCh,
 				dialDone:  dialDone,
 				warnChLim: a.warnOnChannelLimit,
 			}
@@ -476,8 +478,9 @@ func (a *Client) Serve() {
 					go runpprof.Do(context.Background(), labels, func(context.Context) { eConn.cleanup() })
 					return
 				}
-				go runpprof.Do(context.Background(), labels, func(context.Context) { a.remoteToProxy(connID, eConn) })
-				go runpprof.Do(context.Background(), labels, func(context.Context) { a.proxyToRemote(connID, eConn) })
+				go runpprof.Do(context.Background(), labels, func(context.Context) { a.remoteToSendChannel(connID, eConn) })
+				go runpprof.Do(context.Background(), labels, func(context.Context) { a.sendChannelToProxy(connID, eConn) })
+				go runpprof.Do(context.Background(), labels, func(context.Context) { a.dataChannelToRemote(connID, eConn) })
 			})
 
 		case client.PacketType_DATA:
@@ -490,7 +493,7 @@ func (a *Client) Serve() {
 
 			eConn, ok := a.connManager.Get(data.ConnectID)
 			if ok {
-				eConn.send(data.Data)
+				eConn.sendViaDataChannel(data.Data)
 			} else {
 				klog.V(2).InfoS("received DATA for unrecognized connection", "connectionID", data.ConnectID)
 				a.Send(&client.Packet{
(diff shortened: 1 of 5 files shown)


tests/proxy_test.go
```diff
@@ -562,7 +570,7 @@ func TestProxy_LargeResponse(t *testing.T) {
 	expectCleanShutdown(t)
 
 	ctx := context.Background()
-	length := 1 << 20 // 1M
+	length := 1 << 21 // 1M
```
justinsb: I think that is 2M :-)

pkg/agent/client.go
```diff
@@ -534,27 +545,23 @@ func (a *Client) Serve() {
...
-	resp := &client.Packet{
-		Type: client.PacketType_DATA,
-	}
 
 	for {
 		n, err := eConn.conn.Read(buf[:])
 		klog.V(5).InfoS("received data from remote", "bytes", n, "connectionID", connID)
 
 		if err == io.EOF {
-			klog.V(2).InfoS("remote connection EOF", "connectionID", connID)
```
justinsb, line 557: Should we keep this?

pkg/agent/client.go
```diff
@@ -50,17 +50,18 @@ type endpointConn struct {
 	connID    int64
 	cleanFunc func()
 	dataCh    chan []byte
+	sendCh    chan []byte
 	cleanOnce sync.Once
 	warnChLim bool
 	dialDone  chan struct{}
+	sendDone  chan struct{}
```
justinsb: We should start adding comments (somewhere) defining the contract.  I would personally put them here on the struct.

I think the contract is that sendDone is closed by the goroutine reading from sendCh, when it has read to the end of sendCh and sent all the packets.

pkg/agent/client.go
```diff
@@ -383,10 +384,15 @@ func (a *Client) Serve() {
 
 			connID := atomic.AddInt64(&a.nextConnID, 1)
 			dataCh := make(chan []byte, a.cs.xfrChannelSize)
+			sendCh := make(chan []byte, a.cs.xfrChannelSize)
 			dialDone := make(chan struct{})
+			sendDone := make(chan struct{})
```
justinsb, line 402: Super-nit: I would probably group sendCh and sendDone together.  I think one day we might want to create a type that wraps the channel to give behaviour that is less likely to panic / easier to use.

pkg/agent/client.go
```diff
@@ -451,6 +459,7 @@ func (a *Client) Serve() {
 					if err := a.Send(dialResp); err != nil {
 						klog.ErrorS(err, "could not send DIAL_RSP with error", "dialID", dialReq.Random, "connectionID", connID, "dialAddress", dialReq.Address)
 					}
+					close(eConn.sendDone)
```
justinsb, line 475: We _could_ drain sendCh and count the packets, but it would only be for diagnostic purposes.  (And to honor the contract)

pkg/agent/client.go
```diff
@@ -534,27 +545,23 @@ func (a *Client) Serve() {
 	}
 }
 
-func (a *Client) remoteToProxy(connID int64, eConn *endpointConn) {
+func (a *Client) remoteToSendChannel(connID int64, eConn *endpointConn) {
```
justinsb, line 579: The naming is quite difficult. Maybe kube-apiserver, control-plane proxy, target-side proxy,  target.  But even that I don't love.  Not for this PR, anyway

pkg/agent/client.go
```diff
@@ -566,23 +573,70 @@ func (a *Client) remoteToProxy(connID int64, eConn *endpointConn) {
...
+		// As the read side of the dataCh channel, we cannot close it.
+		// However serve() may be blocked writing to the channel,
+		// so we need to consume the channel until it is closed.
+		discardedPktCount := 0
+		for range eConn.sendCh {
+			// Ignore values as this indicates there was a problem
+			// with the remote connection.
+			discardedPktCount++
+		}
+		if discardedPktCount > 0 {
```
justinsb, line 644: I knew I had seen the "drain sendCh and log" code somewhere :-)

pkg/agent/client.go
```diff
@@ -566,23 +573,70 @@ func (a *Client) remoteToProxy(connID int64, eConn *endpointConn) {
...
+				},
+			},
+		}
 
-		resp.Payload = &client.Packet_Data{Data: &client.Data{
-			Data:      buf[:n],
-			ConnectID: connID,
-		}}
 		if err := a.Send(resp); err != nil {
 			klog.ErrorS(err, "could not send DATA", "connectionID", connID)
```
justinsb, line 661: Silently dropping errors here is going to cause problems ... I think we should either force-close the connection, or explicitly send an error back to the client.

We should at least break from the loop, rather than continuing to try to send the next packet; if it recovers that's going to be bad.

I suggest creating a bug rather than trying to fix here.

pkg/agent/client.go
```diff
@@ -566,23 +573,70 @@ func (a *Client) remoteToProxy(connID int64, eConn *endpointConn) {
...
+			return
+		}
+	}
+}
+
+func (a *Client) sendChannelToProxy(connID int64, eConn *endpointConn) {
+	defer close(eConn.sendDone)
+	defer func() {
+		if panicInfo := recover(); panicInfo != nil {
+			klog.V(2).InfoS("Exiting sendChannelToProxy with recovery", "panicInfo", panicInfo, "connectionID", connID)
```
justinsb: I think relying on `recover()` should be a V(0) or ErrorS.  And maybe in tests we run in "strict mode" and fail the test if we ever need recover?  One for the future though.

justinsb:
Code changes look good to me, though I would document the contract around sendCh & sendDone.

The others are super-nits or existing footguns, but I think we should probably work towards addressing them

tests/proxy_test.go
cheftako: :P

pkg/agent/client.go
cheftako: Yes. I incorrectly thought it was part of the debug logging I had added.

pkg/agent/client.go
cheftako: Documented.

pkg/agent/client.go
cheftako: Done and in a couple of other places as well.

pkg/agent/client.go
cheftako: Good catch. Fixed.

pkg/agent/client.go
cheftako: I like the "SendChannel" part as I think it correctly reflects 1/2 of whats going on. I would like find something better than "remote".

pkg/agent/client.go
cheftako: +1

pkg/agent/client.go
cheftako: Added a return. It should better indicate the dropped packets but honestly I think thats the only thing it will change. Recovery should be fine for TCP packets. Other traffic may be different.

pkg/agent/client.go
cheftako: Bumped to V(0)

PR #888, opened 2026-08-01
fix: scope removeEstablishedForBackendConn to the retiring backend

singhharsh1708:
While reading through the server I noticed `removeEstablishedForBackendConn` deletes the whole per-agent `established` map on the first matching connection (`delete(s.established, agentID)`), rather than just the connections belonging to the backend that's going away.

That's fine when there's one backend per agentID, but `backends` is a slice per agentID, so with an HA agent you can have several. In that case retiring one backend also evicts the connections served by the *other* backends under the same agentID. They get dropped from the routing map and aren't added to the returned slice, so the caller never sends them a `CLOSE_RSP` and their frontend connections leak.

The two sibling functions already do this the right way — `removeEstablished` and `removeEstablishedForStream` delete per-connID and only drop the agent entry once its inner map is empty. This just brings `removeEstablishedForBackendConn` in line with them.

I also added `TestRemoveEstablishedForBackendConnPreservesOtherBackends`: two connections under one agentID on two different backends, retire one, and check the other is still routable while only the retired backend's connection comes back. It fails on the current code and passes with the change.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -456,11 +456,14 @@ func (s *ProxyServer) removeEstablishedForBackendConn(agentID string, backend *B
 	if !ok {
 		return nil, fmt.Errorf("can't find agentID %s in the established", agentID)
 	}
-	for _, frontend := range established {
+	for connID, frontend := range established {
 		if frontend.backend == backend {
-			delete(s.established, agentID)
+			delete(established, connID)
 			ret = append(ret, frontend)
 		}
+	}
+	if len(established) == 0 {
+		delete(s.established, agentID)
 	}
 
 	metrics.Metrics.SetEstablishedConnCount(s.getCount(s.established))
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -464,6 +464,8 @@ func (s *ProxyServer) removeEstablishedForBackendConn(agentID string, backend *B
 	}
 	if len(established) == 0 {
 		delete(s.established, agentID)
+	} else {
+		klog.V(2).InfoS("Frontends established over other backend connections remain after backend removal", "agentID", agentID, "remaining", len(established))
 	}
 
 	metrics.Metrics.SetEstablishedConnCount(s.getCount(s.established))
--- a/pkg/server/server_test.go
+++ b/pkg/server/server_test.go
@@ -1087,3 +1087,29 @@ func dialClosePkt(dialID int64) *client.Packet {
 		},
 	}
 }
+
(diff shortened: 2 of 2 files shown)


pkg/server/server.go
```diff
@@ -456,12 +456,15 @@ func (s *ProxyServer) removeEstablishedForBackendConn(agentID string, backend *B
 	if !ok {
 		return nil, fmt.Errorf("can't find agentID %s in the established", agentID)
 	}
-	for _, frontend := range established {
+	for connID, frontend := range established {
 		if frontend.backend == backend {
-			delete(s.established, agentID)
+			delete(established, connID)
```
cheftako, line 461: golang garbage collector should clean up the ProxyClientConnections once the map[in64]*ProxyClientConnection is cleaned up. However this probably speeds up that clean up and makes it more deterministic.

pkg/server/server.go
```diff
@@ -456,12 +456,15 @@ func (s *ProxyServer) removeEstablishedForBackendConn(agentID string, backend *B
...
 			ret = append(ret, frontend)
 		}
 	}
+	if len(established) == 0 {
```
cheftako, line 465: This will potentially fail to delete the map where we were deleting it before. I buy warning if len() > 0 but I think we should delete if any instance of `if frontend.backend == backend` is true. Anything else it potentially introducing a memory leak.

cheftako (commented):
Potentially a good improvement but please take a look at the flagged edge case.

pkg/server/server.go
singhharsh1708: Right, and deleting the per-connID entries directly also makes that cleanup deterministic instead of waiting for the whole agent entry to be dropped.

pkg/server/server.go
singhharsh1708: The entry is kept deliberately when frontends established over other backend connections of the same agent remain: those are still live, and dropping the whole agentID entry (what the old code did on the first match) means later cleanup for them can no longer find their ProxyClientConnection. That is the leak this PR is fixing, covered by TestRemoveEstablishedForBackendConnPreservesOtherBackends. The entry is deleted as soon as the inner map is empty, so nothing lingers once the last frontend goes away. Added the log you suggested for the remaining-frontends case.

ipochi (approved):
I didn't realize this PR was already created before creating https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/894 and https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/895.

This looks good. thanks for the improvements @singhharsh1708. I'll close mine and merge this.

fixes: #894

PR #889, opened 2026-08-01
fix: make CloseHTTP idempotent to avoid close of closed channel

singhharsh1708:
The `CloseHTTP` callback for an HTTP-CONNECT connection wraps `conn.Close()` in `closeOnce`, but `close(closed)` sits just outside it:

```go
CloseHTTP: func() error {
    closeOnce.Do(func() { conn.Close() })
    close(closed)
    return nil
},
```

If `CloseHTTP` runs twice for the same connection, that second `close(closed)` panics with "close of closed channel" — and since it's unrecovered, it takes the whole proxy process down, not just the one tunnel. `conn.Close()` being inside `closeOnce` already shows the intent is for this cleanup to happen once; `close(closed)` just got left out of the guard. Moving it inside the same `closeOnce` makes the callback idempotent. Nothing changes on the normal single-close path.

### DIFF
--- a/pkg/server/tunnel.go
+++ b/pkg/server/tunnel.go
@@ -106,8 +106,10 @@ func (t *Tunnel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		Mode: ModeHTTPConnect,
 		HTTP: io.ReadWriter(conn), // pass as ReadWriter so the caller must close with CloseHTTP
 		CloseHTTP: func() error {
-			closeOnce.Do(func() { conn.Close() })
-			close(closed)
+			closeOnce.Do(func() {
+				conn.Close()
+				close(closed)
+			})
 			return nil
 		},
 		connected: connected,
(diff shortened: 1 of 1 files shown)


pkg/server/tunnel.go
```diff
@@ -106,8 +106,10 @@ func (t *Tunnel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		Mode: ModeHTTPConnect,
 		HTTP: io.ReadWriter(conn), // pass as ReadWriter so the caller must close with CloseHTTP
 		CloseHTTP: func() error {
-			closeOnce.Do(func() { conn.Close() })
-			close(closed)
+			closeOnce.Do(func() {
+				conn.Close()
```
cheftako, line 111: With this change, if something goes wrong with conn.Close(), close(closed) will potentially not get called.
How about
```
defer close(closed)
conn.Close()
```
It gets your desired ordering guaruntee without losing the call guaruntee.

cheftako (commented):
Minor tweak.

pkg/server/tunnel.go
singhharsh1708: Done, applied exactly that: defer close(closed) inside the Do, then conn.Close(). Keeps the ordering and guarantees the channel is closed.

ipochi (approved):
nit: Having a test would be great that exercises the actual production CloseHTTP callback and verifies:

Calling CloseHTTP() twice does not panic.
The underlying conn.Close() runs exactly once.
Ideally, concurrent calls are also safe; run that test with -race.

PR #891, opened 2026-08-06
Fix drain failure structured logging

kahirokunn:
The existing `klog.ErrorS` call passes a lone empty string as a key without a value. klog renders that as a missing structured value, producing malformed log output and losing useful client context.


ipochi (approved):
/lgtm

thank you for the fix @kahirokunn

PR #893, opened 2026-08-10
Fix agent DRAIN signaling blocked by stream Recv

kahirokunn:
When an agent starts draining while its proxy stream is idle, `Serve` is blocked in `Recv` and cannot observe `drainCh`. The DRAIN packet may never reach the proxy server, allowing new connections to be assigned to an agent that is shutting down.

This moves drain notification out of the receive loop so it can be sent as soon as `drainCh` closes.


ipochi:
/lgtm

thank you for the fix @kahirokunn

PR #895, opened 2026-08-11
server: preserve frontends on other backend connections

### LINKED ISSUE #894: Backend teardown can remove connections from an overlapping stream with the same agent ID
### Problem

  This is a reconnect/handover race, not something that occurs on every backend
  disconnect. It requires:

  1. Two backend streams with the same `agentID` to overlap on one server.
  2. Both streams to have established connections.
  3. One stream to disconnect while the other remains active.

  During cleanup, `removeEstablishedForBackendConn` finds connections belonging
  to the terminating backend but deletes the entire agent entry. Connections
  belonging to the surviving backend are therefore removed from routing state
  without being returned for cleanup.

  A same-process reconnect retains the agent ID and can create this overlap. An
  agent restart is only a likely trigger when a stable `--agent-id` or
  `PROXY_AGENT_ID` is reused; the default generated UUID normally changes on
  restart.

  ### Impact

  If the race occurs, connections on the surviving stream become untracked,
  subsequent DATA is treated as having a missing frontend, and the established
  connection metric under-reports. The scope is limited to connections sharing
  that agent ID on the affected server during the overlap.

  ### Expected behavior

  Remove only connections associated with the terminating backend. Delete the
  outer `agentID` entry only when no connections remain.

ipochi:
When backend streams with the same agent ID overlap, cleanup of one stream deletes the entire established connection map for that agent. This can untrack connections belonging to a surviving backend stream.

Remove only connections associated with the terminating backend, and delete the agent entry only when no connections remain. Add a regression test covering the overlapping-backend case.

Fixes: #894

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -456,11 +456,14 @@ func (s *ProxyServer) removeEstablishedForBackendConn(agentID string, backend *B
 	if !ok {
 		return nil, fmt.Errorf("can't find agentID %s in the established", agentID)
 	}
-	for _, frontend := range established {
+	for connID, frontend := range established {
 		if frontend.backend == backend {
-			delete(s.established, agentID)
+			delete(established, connID)
 			ret = append(ret, frontend)
 		}
+	}
+	if len(established) == 0 {
+		delete(s.established, agentID)
 	}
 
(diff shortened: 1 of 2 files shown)


ipochi:
closing in favor of https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/888

PR #896, opened 2026-08-11
server: queue HTTP CONNECT frontend writes

### LINKED ISSUE #881: HTTP-CONNECT head-of-line blocking: one slow frontend socket stalls all connections on an agent stream
In the kube-apiserver → konnectivity-server → agent (HTTP-CONNECT) topology, a single slow or unresponsive HTTP-CONNECT client can stall every connection multiplexed on the same agent stream. serveRecvBackend writes to the hijacked HTTP socket synchronously while consuming the shared per-agent receive channel, so one blocked write parks the only consumer of that channel. Unrelated DIAL_RSP, DATA, and control packets queue behind it, recvCh saturates, and backpressure propagates into gRPC and the agent.

Impact

- Webhook calls fail closed with context deadline exceeded at the configured deadline.
- Tunnel establishment for healthy, unrelated connections is delayed well beyond normal.
- One agent stream may multiplex hundreds/thousands of connections; a single slow reader degrades all of them.

Root cause

serveRecvBackend processes packets from one agent serially and calls frontend.send → HTTP.Write inline:

1. readBackendToChannel reads a packet from the backend gRPC stream.
2. It inserts into the bounded recvCh.
3. serveRecvBackend removes packets one at a time.
4. For DATA to an HTTP-CONNECT frontend it synchronously writes to the hijacked socket.

If that socket write blocks (slow/stalled client), step 4 never returns, the shared consumer stops, recvCh fills, and every other connection on that agent waits behind it. The slow connection and the victim connection are different connections — that's the head-of-line blocking.

Evidence

kube-apiserver, failing closed at the deadline:

```
W0717 23:21:05.212067  1 dispatcher.go:217] Failed calling webhook, failing closed validation.istio.io: failed to call webhook: Post "https://istiod-stable.istio-system.svc:443/validate?timeout=10s": context deadline exceeded
W0717 23:23:41.764863  1 dispatcher.go:217] Failed calling webhook, failing closed validation.istio.io: ... context deadline exceeded
W0717 23:26:51.024549  1 dispatcher.go:217] Failed calling webhook, failing closed validation.istio.io: ... context deadline exceeded
```

Corresponding konnectivity-server lifecycle — establishment for an unrelated connection completes long after the webhook has already timed out:

```
I0717 23:20:55.228454  1 tunnel.go:58]  "Received request for host" method="CONNECT" host="172.17.4.169:15017"
I0717 23:23:10.971401  1 server.go:934] "Proxy connection established" connectionID=25707 agentID="ba78a7a7-..."
I0717 23:23:10.971527  1 tunnel.go:159] "Connection established, sent 200 OK" connectionID=25707 agentID="ba78a7a7-..."
```

The CONNECT start times line up with the apiserver failures at the deadline; establishment then completes much later. Some attempts established quickly but still missed the deadline in the tunneled exchange. (The EOF from host for 25707 belongs to that connection, not others sharing the window — correlate by host + agentID + connectionID, not timestamp.)

Required invariant

▎ A blocked HTTP socket write for connection A must never prevent serveRecvBackend from processing a packet for connection B.

ipochi:
Queue HTTP-CONNECT frontend writes per connection so one blocked socket does not immediately stall the shared agent receive loop. The bounded queue preserves ordering and existing backpressure, while delaying—not eliminating—HOL blocking when capacity is reached. Add metrics to expose full queues and receive loops blocked on them.

Fixes: #881

### DIFF
--- a/cmd/server/app/options/options.go
+++ b/cmd/server/app/options/options.go
@@ -109,8 +109,9 @@ type ProxyRunOptions struct {
 	// Minimum TLS version for server connections.
 	// Accepted values: VersionTLS10, VersionTLS11, VersionTLS12, VersionTLS13.
 	// If empty, defaults to VersionTLS12.
-	TLSMinVersion  string
-	XfrChannelSize int
+	TLSMinVersion            string
+	XfrChannelSize           int
+	FrontendWriteChannelSize int
 
 	// Lease controller configuration
 	EnableLeaseController bool
@@ -162,6 +163,7 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.StringSliceVar(&o.CipherSuites, "cipher-suites", o.CipherSuites, "The comma separated list of allowed cipher suites. Has no effect on TLS1.3. Empty means allow default list.")
 	flags.StringVar(&o.TLSMinVersion, "tls-min-version", o.TLSMinVersion, "Minimum TLS version for server connections. Accepted values: VersionTLS10, VersionTLS11, VersionTLS12, VersionTLS13. Empty defaults to VersionTLS12.")
 	flags.IntVar(&o.XfrChannelSize, "xfr-channel-size", o.XfrChannelSize, "The size of the two KNP server channels used in server for transferring data. One channel is for data coming from the Kubernetes API Server, and the other one is for data coming from the KNP agent.")
+	flags.IntVar(&o.FrontendWriteChannelSize, "frontend-write-channel-size", o.FrontendWriteChannelSize, "The number of packets buffered for each HTTP CONNECT frontend before backend receive processing blocks.")
 	flags.BoolVar(&o.EnableLeaseController, "enable-lease-controller", o.EnableLeaseController, "Enable lease controller to publish and garbage collect proxy server leases.")
 	flags.StringVar(&o.LeaseNamespace, "lease-namespace", o.LeaseNamespace, "The namespace where lease objects are managed by the controller.")
 	flags.StringVar(&o.LeaseLabel, "lease-label", o.LeaseLabel, "The labels on which the lease objects are managed.")
@@ -211,6 +213,7 @@ func (o *ProxyRunOptions) Print() {
 	klog.V(1).Infof("CipherSuites set to %q.\n", o.CipherSuites)
 	klog.V(1).Infof("TLSMinVersion set to %q.\n", o.TLSMinVersion)
 	klog.V(1).Infof("XfrChannelSize set to %d.\n", o.XfrChannelSize)
+	klog.V(1).Infof("FrontendWriteChannelSize set to %d.\n", o.FrontendWriteChannelSize)
 	klog.V(1).Infof("GracefulShutdownTimeout set to %v.\n", o.GracefulShutdownTimeout)
 	klog.V(1).Infof("BackendDialTimeout set to %v.\n", o.BackendDialTimeout)
 }
@@ -340,6 +343,9 @@ func (o *ProxyRunOptions) Validate() error {
 	if o.XfrChannelSize <= 0 {
 		return fmt.Errorf("channel size %d must be greater than 0", o.XfrChannelSize)
 	}
+	if o.FrontendWriteChannelSize <= 0 {
+		return fmt.Errorf("frontend write channel size %d must be greater than 0", o.FrontendWriteChannelSize)
+	}
 	// validate the TLS min version
 	if o.TLSMinVersion != "" {
 		tlsVer, err := util.GetTLSVersion(o.TLSMinVersion)
@@ -417,6 +423,7 @@ func NewProxyRunOptions() *ProxyRunOptions {
 		CipherSuites:              make([]string, 0),
 		TLSMinVersion:             "",
 		XfrChannelSize:            10,
+		FrontendWriteChannelSize:  10,
 		EnableLeaseController:     false,
 		LeaseNamespace:            "kube-system",
 		LeaseLabel:                "k8s-app=konnectivity-server",
--- a/cmd/server/app/options/options_test.go
+++ b/cmd/server/app/options/options_test.go
@@ -64,6 +64,7 @@ func TestDefaultServerOptions(t *testing.T) {
 	assertDefaultValue(t, "CipherSuites", defaultServerOptions.CipherSuites, make([]string, 0))
 	assertDefaultValue(t, "TLSMinVersion", defaultServerOptions.TLSMinVersion, "")
 	assertDefaultValue(t, "XfrChannelSize", defaultServerOptions.XfrChannelSize, 10)
+	assertDefaultValue(t, "FrontendWriteChannelSize", defaultServerOptions.FrontendWriteChannelSize, 10)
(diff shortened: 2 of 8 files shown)


cheftako:
https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/release-0.36/pkg/server/server.go#L609 demonstrates we already use a queue to prevent frontend writes from blocking any channel type. Not sure how this is different.

ipochi:
> https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/release-0.36/pkg/server/server.go#L609 demonstrates we already use a queue to prevent frontend writes from blocking any channel type. Not sure how this is different.

@cheftako 

The cited queue handles packets that the server receives from a gRPC frontend before forwarding them to the agent. It is on the client-to-agent path and is only created by the gRPC Proxy() handler; HTTP-CONNECT does not traverse it.

The HOL addressed here is on the reverse path. One shared serveRecvBackend loop receives packets from an agent and writes them to many HTTP-CONNECT sockets. This PR adds a queue after that loop has demultiplexed the packet, so each HTTP connection gets an independent write lane.

cheftako:
> > https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/release-0.36/pkg/server/server.go#L609 demonstrates we already use a queue to prevent frontend writes from blocking any channel type. Not sure how this is different.
> 
> @cheftako
> 
> The cited queue handles packets that the server receives from a gRPC frontend before forwarding them to the agent. It is on the client-to-agent path and is only created by the gRPC Proxy() handler; HTTP-CONNECT does not traverse it.
> 
> The HOL addressed here is on the reverse path. One shared serveRecvBackend loop receives packets from an agent and writes them to many HTTP-CONNECT sockets. This PR adds a queue after that loop has demultiplexed the packet, so each HTTP connection gets an independent write lane.

+1 on we should ensure HTTP-CONNECT is going through the transfer buffer to prevent push back from a slow packet write. However I would like to standardize the code as much as possible which does that. I think we need to have HTTP-CONNECT use its own ProxyServer class to get the code to be more standard. https://docs.google.com/document/d/1_ZR30lfVuDCF3cYNZOuzsJUfi77JQfKh-veJoA67d3g/edit?resourcekey=0-A1quGsu2r6QlJ9YHvA5nbQ&tab=t.dpg2tsu90eiy#heading=h.618lv3qm4neb is a start on that.

ipochi:
@cheftako friendly ping :)

cmd/server/app/options/options.go
```diff
@@ -162,6 +163,7 @@ func (o *ProxyRunOptions) Flags() *pflag.FlagSet {
 	flags.StringSliceVar(&o.CipherSuites, "cipher-suites", o.CipherSuites, "The comma separated list of allowed cipher suites. Has no effect on TLS1.3. Empty means allow default list.")
 	flags.StringVar(&o.TLSMinVersion, "tls-min-version", o.TLSMinVersion, "Minimum TLS version for server connections. Accepted values: VersionTLS10, VersionTLS11, VersionTLS12, VersionTLS13. Empty defaults to VersionTLS12.")
 	flags.IntVar(&o.XfrChannelSize, "xfr-channel-size", o.XfrChannelSize, "The size of the two KNP server channels used in server for transferring data. One channel is for data coming from the Kubernetes API Server, and the other one is for data coming from the KNP agent.")
+	flags.IntVar(&o.FrontendWriteChannelSize, "frontend-write-channel-size", o.FrontendWriteChannelSize, "The number of packets buffered for each HTTP CONNECT frontend before backend receive processing blocks.")
```
cheftako, line 166: [Non-blocking] On a follow up it would be good to make this not HTTP-CONNECT specific. Would also be nice if setting a channel size of 0 disabled this behavior. Would be nice is the message indicated more clearly that this is a per connection buffer.

cmd/server/app/options/options_test.go
```diff
@@ -206,6 +207,16 @@ func TestValidate(t *testing.T) {
 			value:    -10,
 			expected: fmt.Errorf("channel size -10 must be greater than 0"),
 		},
+		"ZeroFrontendWriteChannelSize": {
```
cheftako, line 210: Would be nice if we used 0 to allow the old behavior.

pkg/server/http_connect_frontend_writer.go
```diff
@@ -0,0 +1,149 @@
...
+	fullGauge prometheus.Gauge // guarded by mu; keeps Inc and Dec on the same GaugeVec child
+	stopped   bool             // guarded by mu
+}
+
+// enqueueWrite gives each HTTP-CONNECT stream a bounded response buffer behind
+// the common Frontend.Send path. When it fills, Send blocks without dropping
+// packets or disconnecting a slow client. This delays, rather than eliminates,
+// head-of-line blocking of the shared agent receive loop.
+func (h *httpConnectStream) enqueueWrite(pkt *client.Packet) error {
+	select {
```
cheftako, line 46: Do we need this select? Seems like the next select contains this already, so I think this is superfluous.

pkg/server/http_connect_frontend_writer.go
```diff
@@ -0,0 +1,149 @@
...
+
+// serveFrontendWrites owns established socket writes. The HTTP handshake is
+// completed before this goroutine starts, and CLOSE_RSP follows queued DATA.
+// The socket is closed only on ordered close, an actual write error, or abort.
+func (h *httpConnectStream) serveFrontendWrites() {
+	defer close(h.writerDone)
+	defer h.close()
+
+	for {
+		select {
```
cheftako, line 83: Again not sure we need this select.

pkg/server/http_connect_frontend_writer.go
```diff
@@ -0,0 +1,149 @@
...
+		case <-h.closed:
+			return
+		default:
+		}
+
+		select {
+		case pkt := <-h.writeCh:
+			h.updateFrontendWriteQueueMetric()
+			if pkt.Type == client.PacketType_CLOSE_RSP {
+				return
```
cheftako, line 93: [Non-blocking] Would be good to log (v5?) we got a CLOSE_RSP.

pkg/server/tunnel.go
```diff
@@ -131,32 +131,46 @@ type httpConnectStream struct {
 	closed    chan struct{}
 	closeOnce sync.Once
 
+	// The writer starts only after the HTTP 200 response has been written.
+	// Send queues established DATA and CLOSE_RSP in wire order. Neither channel
+	// is replaced, and writeCh is never closed, so Close can race with Send.
+	writeCh      chan *client.Packet
```
cheftako, line 137: [Non-blocking] Would be good to rename this for clarity. connWriteCh? frontendWriteCh?

cheftako:
/lgtm
/approve
Merging the code but I think we should seriously consider making 0 adhere to the old behavior. We can that value to determine if we need the extra goroutine and which way to do the write (write to conn on 0 or write to queue on non zero)

ipochi:
> /lgtm /approve Merging the code but I think we should seriously consider making 0 adhere to the old behavior. We can that value to determine if we need the extra goroutine and which way to do the write (write to conn on 0 or write to queue on non zero)

Done in #918

PR #898, opened 2026-08-11
Use the server Mode constants instead of string literals

kahirokunn:
/kind cleanup

pkg/server defines `ModeGRPC`/`ModeHTTPConnect` and most of the code already compares against them. This converts the last few `"grpc"` literals: the frontend mode checks, the option default, and the dial path in pkg/server.

```release-note
NONE
```


ipochi (approved):
/lgtm

Thank you for the fix. @kahirokunn

PR #901, opened 2026-08-11
server: prefer less congested backends in random routing

### LINKED ISSUE #900: Random routing ignores backend receive-channel congestion
Default and default-route routing may select an agent whose receive channel is already congested, even when healthier agents are available. This can compound tunnel delays by continuing to send new connections to a pressured agent stream.

This makes a situation like #881 even worse.

ipochi:
Use receive-channel occupancy as a lightweight routing signal. Random routing samples two non-draining agents and selects the less occupied one, while preserving existing single-agent and all-draining fallback behavior. Destination-host routing is unchanged.

fixes: #900 , and helps in #881

### DIFF
--- a/pkg/server/backend_manager.go
+++ b/pkg/server/backend_manager.go
@@ -56,6 +56,10 @@ type Backend struct {
 
 	// draining indicates if this backend is draining and should not accept new connections
 	draining atomic.Bool
+
+	// recvCh is assigned before the backend is published and is immutable
+	// afterward. It is observed only to measure receive-path pressure.
+	recvCh <-chan *client.Packet
 }
 
 // IsDraining returns true if the backend is draining
@@ -66,6 +70,17 @@ func (b *Backend) IsDraining() bool {
 // SetDraining marks the backend as draining
 func (b *Backend) SetDraining() {
 	b.draining.Store(true)
+}
+
+// RecvChannelOccupancy reports how much of the backend's receive channel is
+// occupied. Channel length is safe to sample concurrently, but the result is
+// only a point-in-time routing signal. An unavailable sample is treated as
+// neutral.
+func (b *Backend) RecvChannelOccupancy() float64 {
+	if b.recvCh == nil || cap(b.recvCh) == 0 {
+		return 0
+	}
+	return float64(len(b.recvCh)) / float64(cap(b.recvCh))
 }
 
 // Retire marks the backend as draining and closes Done once.
--- a/pkg/server/backend_manager.go
+++ b/pkg/server/backend_manager.go
@@ -457,27 +457,68 @@ func ignoreNotFound(err error) error {
 	return err
 }
 
-// GetRandomBackend returns a random non-draining backend, falling back to a
-// random draining backend only when no non-draining agents are available.
+// lessOccupiedBackend returns the backend with less receive-channel pressure.
+// Equal occupancy is resolved randomly so idle backends continue to share load.
+// The caller must hold s.mu.
+func (s *DefaultBackendStorage) lessOccupiedBackend(first, second *Backend) *Backend {
+	firstOccupancy := first.RecvChannelOccupancy()
+	secondOccupancy := second.RecvChannelOccupancy()
+	if firstOccupancy < secondOccupancy {
+		return first
+	}
+	if secondOccupancy < firstOccupancy {
+		return second
+	}
+	if s.random.Intn(2) == 0 {
+		return first
+	}
+	return second
+}
+
+// GetRandomBackend samples two distinct non-draining agents and prefers the
+// primary backend with less receive-channel pressure. A sole non-draining
+// agent is used directly; draining agents are used only as a last resort.
+// The storage must maintain the random-routing index; otherwise this method
+// returns ErrNotFound even when backends are registered.
 func (s *DefaultBackendStorage) GetRandomBackend() (*Backend, error) {
+	// Selection may repair a state-list entry after observing a concurrently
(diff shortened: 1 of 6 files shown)


pkg/server/backend_manager.go
```diff
@@ -68,6 +72,17 @@ func (b *Backend) SetDraining() {
 	b.draining.Store(true)
 }
```
cheftako, line 74: Using receive channel occupancy as a lightweight load signal without requiring additional synchronization or coordination locks is a clean approach. Handling nil/unbuffered channels as neutral (0 occupancy) makes it safe across various test setups as well.

pkg/server/backend_manager.go
```diff
@@ -184,12 +199,20 @@ type BackendManager interface {
 	Backend(ctx context.Context) (*Backend, error)
 	// AddBackend adds a backend.
 	AddBackend(backend *Backend)
-	// RemoveBackend adds a backend.
+	// RemoveBackend removes a backend.
 	RemoveBackend(backend *Backend)
 	BackendStorage
 	ReadinessManager
 }
```
cheftako, line 209: I like the use of the unexported `backendDrainingMarker` interface here to notify managers maintaining routing indices without altering the public `BackendManager` interface.

pkg/server/backend_manager.go
```diff
@@ -367,43 +455,82 @@ func ignoreNotFound(err error) error {
...
+		return selected, nil
 	}
 
-	// All agents are draining, use one as fallback
-	if firstDrainingBackend != nil {
-		agentID := firstDrainingBackend.id
+	if len(s.drainingAgentIDs) > 0 {
+		agentID := s.drainingAgentIDs[s.random.Intn(len(s.drainingAgentIDs))]
+		backend := s.backends[agentID][0]
 		klog.V(3).InfoS("No non-draining backends available, using draining backend as fallback", "agentID", agentID)
```
cheftako, line 527: nit: We could use `selected.GetAgentID()` here to stay consistent with the accessor method used elsewhere in this file (e.g. lines 234, 241), though direct field access works fine within the package.
```suggestion
		klog.V(3).InfoS("Pick agent as backend", "agentID", selected.GetAgentID())
```
Fine as a follow-up or left as-is.

pkg/server/server.go
```diff
@@ -900,11 +911,12 @@ func (s *ProxyServer) Connect(stream agent.AgentService_ConnectServer) error {
 	}
 
 	klog.V(2).InfoS("Agent connected", "agentID", agentID, "serverID", s.serverID)
+	recvCh := make(chan *client.Packet, s.xfrChannelSize)
+	backend.recvCh = recvCh
```
cheftako, line 919: Assigning `backend.recvCh` prior to calling `s.addBackend(backend)` ensures the receive channel is initialized and published before any concurrent routing caller can observe the backend and sample its occupancy.

pkg/server/backend_manager.go
```diff
@@ -56,6 +56,10 @@ type Backend struct {
 
 	// draining indicates if this backend is draining and should not accept new connections
 	draining atomic.Bool
+
+	// recvCh is assigned before the backend is published and is immutable
+	// afterward. It is observed only to measure receive-path pressure.
+	recvCh <-chan *client.Packet
```
cheftako, line 62: I like the idea/approach. However as per https://docs.google.com/document/d/1_ZR30lfVuDCF3cYNZOuzsJUfi77JQfKh-veJoA67d3g/edit?resourcekey=0-A1quGsu2r6QlJ9YHvA5nbQ&tab=t.1s2vhr3sjg7g#bookmark=id.ful2x7g3263u we already have transfer channel for GRPC in ProxyServer. I don't want to add a second channel/queue to the path. I think we need to finalize adding a transfer channel for HTTP-CONNECT. We can then use those channels to achieve what you have here.

pkg/server/backend_manager.go
```diff
@@ -68,6 +72,17 @@ func (b *Backend) SetDraining() {
 	b.draining.Store(true)
 }
 
+// RecvChannelOccupancy reports how much of the backend's receive channel is
+// occupied. An unavailable occupancy sample is treated as neutral.
+func (b *Backend) RecvChannelOccupancy() float64 {
+	if b.recvCh == nil || cap(b.recvCh) == 0 {
+		return 0
+	}
+	return float64(len(b.recvCh)) / float64(cap(b.recvCh))
```
cheftako, line 83: Watch out len(b.recvCh) is NOT thread safe. I think that is ok in this case however it can cause problems with race testing. https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/master/pkg/agent/client.go#L84 is an example of how we protect our tests from failure for this.

pkg/server/backend_manager.go
ipochi: Thanks for the feedback @cheftako . This PR does not add another transfer channel. `backend.recvCh` points to the existing per-agent channel created in `ProxyServer.Connect`; it only makes that channel’s occupancy available during backend selection. This is part of the agent-to-frontend path and is shared by both gRPC and HTTP-CONNECT.

The gRPC-only Proxy.recvCh described in the Frontend-to-Agent section is in the opposite direction. PR #896 adds a separate queue after serveRecvBackend for HTTP-CONNECT writes. This PR is independent of #896

pkg/server/backend_manager.go
ipochi: ran the focused tests with -race, and they passed.

ipochi:
@cheftako  This can proceed without a discussion on #896 . What doyou think ?

PR #902, opened 2026-08-12
Reject malformed certificate blocks in CA bundles

kahirokunn:
Loading a CA bundle goes through AppendCertsFromPEM, which silently skips anything it cannot parse. A corrupted bundle is accepted as long as one certificate in it still parses, and the proxy then quietly runs with partial trust, rejecting peers it should have accepted.

This switches the parsing to client-go's certutil, [the same code kube-apiserver uses when it reloads its client CA](https://github.com/kubernetes/kubernetes/blob/0f29094e5b73085e3802ecc1298ecae13866bfe6/staging/src/k8s.io/apiserver/pkg/server/dynamiccertificates/dynamic_cafile_content.go#L269-L280). A CERTIFICATE block that does not parse now fails the whole load.

Non-certificate PEM blocks are still ignored, and bytes that never form a PEM block (a truncated block, for example) are still skipped, same as kube-apiserver. The tests pin down both sides of that line.

## Manual verification on kind

I built and deployed the proxy agent and server from these two revisions side by side:

- Baseline: `master`
- Patched: this PR

Each deployment used a CA bundle containing two valid CA certificates followed by a valid PEM `CERTIFICATE` block with a malformed DER payload.

### Results with the partially malformed bundle

| Revision | Agent | Server | Result |
| --- | --- | --- | --- |
| `master` | Running | Running | The malformed certificate block was silently skipped |
| This PR | CrashLoopBackOff | CrashLoopBackOff | Startup failed because the malformed certificate block was rejected |

Both the agent and server built from this PR rejected the bundle and exited with:

```text
failed to load CA cert: error creating pool from /tls/ca.crt: x509: malformed certificate
```

### Results with a fully valid bundle

After restoring the bundle so that it contained only valid CA certificates, the agent and server built from this PR both started successfully, and the agent connected to the server.

/kind bug

```release-note
CA bundle files containing a certificate block that fails to parse are now rejected with an error instead of the malformed block being silently skipped.
```

### DIFF
--- a/cmd/server/app/server.go
+++ b/cmd/server/app/server.go
@@ -19,14 +19,12 @@ package app
 import (
 	"context"
 	"crypto/tls"
-	"crypto/x509"
 	"fmt"
 	"net"
 	"net/http"
 	netpprof "net/http/pprof"
 	"os"
 	"os/signal"
-	"path/filepath"
 	"runtime"
 	runpprof "runtime/pprof"
 	"strconv"
@@ -41,6 +39,7 @@ import (
 	"google.golang.org/grpc/keepalive"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
+	certutil "k8s.io/client-go/util/cert"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/apiserver-network-proxy/cmd/server/app/options"
 	"sigs.k8s.io/apiserver-network-proxy/konnectivity-client/proto/client"
@@ -422,14 +421,9 @@ func (p *Proxy) getTLSConfig(caFile, certFile, keyFile string, cipherSuites []st
 		return &tls.Config{Certificates: []tls.Certificate{cert}, MinVersion: minVersion, CipherSuites: cipherSuiteIDs}, nil // #nosec G402
 	}
 
-	certPool := x509.NewCertPool()
-	caCert, err := os.ReadFile(filepath.Clean(caFile))
+	certPool, err := certutil.NewPool(caFile)
 	if err != nil {
-		return nil, fmt.Errorf("failed to read cluster CA cert %s: %v", caFile, err)
-	}
-	ok := certPool.AppendCertsFromPEM(caCert)
-	if !ok {
-		return nil, fmt.Errorf("failed to append cluster CA cert to the cert pool")
+		return nil, fmt.Errorf("failed to load CA cert: %w", err)
 	}
 
 	tlsConfig := &tls.Config{ // #nosec G402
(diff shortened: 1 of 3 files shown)


cmd/server/app/server.go
```diff
@@ -422,14 +421,9 @@ func (p *Proxy) getTLSConfig(caFile, certFile, keyFile string, cipherSuites []st
 		return &tls.Config{Certificates: []tls.Certificate{cert}, MinVersion: minVersion, CipherSuites: cipherSuiteIDs}, nil // #nosec G402
 	}
 
-	certPool := x509.NewCertPool()
-	caCert, err := os.ReadFile(filepath.Clean(caFile))
+	certPool, err := certutil.NewPool(caFile)
```
cheftako, line 424: Nice to have this unified with `certutil.NewPool` across both the server and client/agent paths instead of duplicating the manual pool construction.

pkg/util/certificates.go
```diff
@@ -18,31 +18,16 @@ package util //nolint:revive
...
-	return certPool, nil
-}
+	certutil "k8s.io/client-go/util/cert"
+)
 
 // GetClientTLSConfig returns tlsConfig based on x509 certs
 func GetClientTLSConfig(caFile, certFile, keyFile, serverName string, protos []string) (*tls.Config, error) {
-	certPool, err := getCACertPool(caFile)
+	certPool, err := certutil.NewPool(caFile)
 	if err != nil {
```
cheftako, line 29: Switching to `certutil.NewPool` here is a nice cleanup and aligns our CA loading behavior directly with `kube-apiserver`. Wrapping with `%w` keeps the underlying parsing error intact while providing good context.

pkg/util/certificates_test.go
```diff
@@ -0,0 +1,91 @@
...
+		name    string
+		bundle  []byte
+		wantErr bool
+	}{
+		{name: "valid bundle", bundle: caPEM},
+		{name: "non-certificate block is ignored", bundle: append(append([]byte(nil), caPEM...), nonCertificateBlock...)},
+		{name: "malformed certificate block is rejected", bundle: append(append([]byte(nil), caPEM...), malformedCertificatePEM()...), wantErr: true},
+		// Data that never forms a valid PEM block is skipped rather than
+		// rejected, the same boundary as kube-apiserver's dynamic CA reload.
+		{name: "truncated trailing block is skipped", bundle: append(append([]byte(nil), caPEM...), []byte("-----BEGIN CERTIFICATE-----\nMIIBtruncated\n")...)},
```
cheftako, line 46: Appreciate the test cases here pinning down the boundary between unparseable `CERTIFICATE` blocks (which fail fast) vs. incomplete PEM blocks (which are ignored by `pem.Decode`), matching `dynamic_cafile_content` in `kube-apiserver`. I verified the test suite passes locally.

cheftako (approved):
Thanks for the change.

PR #908, opened 2026-09-01
Add post-dial connection-close instrumentation

### LINKED ISSUE #410: Improve post-dial instrumentation

Hariharanpugazh:
## What this PR does

Extends the post-dial metrics (server `connection_duration_seconds`, added in #899)
with per-close-reason instrumentation on both the proxy server and the agent, so
operators can distinguish *how* established connections end.

### Server (`pkg/server`)
- New `connection_close_total{reason}` counter with reasons `frontend_close`,
  `backend_close`, `stream_shutdown`.
- Emitted from `removeEstablished`, `removeEstablishedForBackendConn`, and
  `removeEstablishedForStream`.
- Fixes `removeEstablishedForStream`, which previously did not observe
  `connection_duration_seconds` or refresh `established_connections`.

### Agent (`pkg/agent`)
- New `endpoint_connection_duration_seconds` histogram and
  `endpoint_connection_close_total{reason}` counter (`server_close`,
  `endpoint_close`, `agent_shutdown`), mirroring the server pattern.
- Records dial-success time and observes lifespan + close reason at teardown.

## Testing
- New unit tests cover every close reason on both server and agent.
- `go test ./pkg/server/... ./pkg/agent/... ./pkg/testing/metrics/...` passes.
- `go vet` and golangci-lint (v2.9.0) clean.

## Which issue(s) this PR fixes
Fixes #410


Hariharanpugazh:
Opened #908 for this.

It adds post-dial connection-close instrumentation on top of the
`connection_duration_seconds` metric from #899:

- **Server:** `connection_close_total{reason}` (`frontend_close`,
  `backend_close`, `stream_shutdown`), plus a fix so
  `removeEstablishedForStream` also records duration / refreshes
  `established_connections`.
- **Agent:** `endpoint_connection_duration_seconds` +
  `endpoint_connection_close_total{reason}` (`server_close`,
  `endpoint_close`, `agent_shutdown`).

One design point I'd like reviewer input on: I kept the duration
histograms unlabeled and added separate `*_close_total{reason}`
counters, rather than putting a `reason` label on
`connection_duration_seconds`, to avoid changing a metric that just
shipped. Happy to switch to a labeled histogram if that's preferred.

pkg/agent/metrics/metrics.go
```diff
@@ -93,6 +97,25 @@ func newAgentMetrics() *AgentMetrics {
 		},
 		[]string{"reason"},
 	)
+	connectionDuration := prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: Namespace,
```
cheftako, line 102: Why namespace? Not sure it makes sense in this context.

cheftako: Looks like the others also have it so not going to worry.

ipochi:
1. [P1] Synchronize close-reason reads with writes — pkg/agent/client.go:457 (https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/07635929e0a31a9ad7f9a52ee7785a3c052e96ad/pkg/agent/client.go#L457)

     When endpoint EOF races with a server CLOSE_REQ, remoteToSendChannel can execute cleanup while Serve writes eConn.closeReason through setCloseReason. cleanReasonOnce serializes setters, but this read uses a different sync.Once and has no synchronization with the write. I reproduced the
     data race between lines 457 and 92 using simultaneous endpoint and server closes. Selecting and recording the reason needs a shared synchronization path that includes endpoint-initiated cleanup.

  2. [P2] Don’t classify every CLOSE_RSP as a frontend close — pkg/server/server.go:434 (https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/07635929e0a31a9ad7f9a52ee7785a3c052e96ad/pkg/server/server.go#L434)

     removeEstablished also handles unsolicited CLOSE_RSP packets sent by the agent after endpoint EOF or a read error. The unconditional increment therefore reports frontend_close even when the frontend never requested closure, attributing endpoint disconnects to the client. A focused test
     feeding an unsolicited close through serveRecvBackend reproduced connection_close_total{reason="frontend_close"} 1. Track whether the frontend initiated closure before assigning this reason, and distinguish unsolicited endpoint closes.

pkg/server/server.go
```diff
@@ -431,6 +431,7 @@ func (s *ProxyServer) removeEstablished(agentID string, connID int64) *ProxyClie
 	}
 	metrics.Metrics.SetEstablishedConnCount(s.getCount(s.established))
 	metrics.Metrics.ObserveConnectionDuration(time.Since(ret.establishedAt))
+	metrics.Metrics.ObserveConnectionClose(metrics.ConnectionCloseFrontend)
```
cheftako, line 434: +1 Imran's comment. This should not be force ton close frontend in all cases.

Hariharanpugazh:
> 1. [P1] Synchronize close-reason reads with write

Fixed in 3499463. Replaced the setter-only `cleanReasonOnce` with a dedicated
`closeReasonMu` mutex and routed both sides through it: `setCloseReason`
(first-writer-wins) and a new synchronized `getCloseReason` used by `cleanFunc`.
That removes the unsynchronized read/write between endpoint-initiated cleanup
and the CLOSE_REQ path. Added `TestCloseReasonRace` to exercise concurrent
set/get under `go test -race`.

Hariharanpugazh:
> 2\. [P2] Don’t classify every CLOSE_RSP as a frontend close — pkg/server/server.go:434

Fixed in 3499463. The server now tracks whether the frontend actually initiated
the close: the frontend CLOSE_REQ handler calls `markFrontendClosed`, which sets
a `frontendClosed` flag on the ProxyClientConnection (under `fmu`).
`removeEstablished` records `frontend_close` only when that flag is set, and a
new `endpoint_close` reason for unsolicited CLOSE_RSPs from the agent.
`TestConnectionCloseMetric` now asserts the unsolicited path yields
`endpoint_close`.

Hariharanpugazh:
Rebased onto master. Resolved the conflict in removeEstablishedForStream
against #917 (kept its duration/count fix and layered the stream_shutdown
close-reason on top), and updated the test for the GrpcFrontend -> Frontend
rename from that PR. Build, vet, and tests pass locally.

pkg/agent/client_test.go
```diff
@@ -221,6 +224,91 @@ func TestClose_Client(t *testing.T) {
 
 }
 
+func TestConnectionCloseMetric_ServerClose(t *testing.T) {
```
cheftako: While `TestConnectionCloseMetric_ServerClose` successfully tests the `server_close` reason, and `TestCloseReasonRace` checks race conditions on `endpointConn` setters, there are no tests validating that `endpoint_close` or `agent_shutdown` metrics are correctly published and scraped on the agent. Adding tests for these scenarios on the agent side would ensure consistent, reliable metric verification across all reasons.

pkg/server/server.go
```diff
@@ -406,9 +410,33 @@ func (s *ProxyServer) removeEstablished(agentID string, connID int64) *ProxyClie
 	}
 	metrics.Metrics.SetEstablishedConnCount(s.getCount(s.established))
 	metrics.Metrics.ObserveConnectionDuration(time.Since(ret.establishedAt))
+	// A CLOSE_RSP is only attributed to the frontend when the frontend actually
+	// requested the close (CLOSE_REQ). Otherwise the agent sent an unsolicited
+	// CLOSE_RSP because the endpoint closed on its own.
+	closeReason := metrics.ConnectionCloseEndpoint
```
cheftako: In `removeEstablished`, if a `CLOSE_RSP` from the agent arrives concurrently with a gRPC client's stream closing/cancelling, a race condition occurs between `serveRecvBackend` (calling `removeEstablished`) and `removeEstablishedForStream`. Whichever processes first deletes the connection and records the metric. While this race is thread-safe and correct, if `removeEstablishedForStream` wins, the close is recorded as `stream_shutdown` instead of `frontend_close` or `endpoint_close`. Consider documenting this subtle metric attribution race behavior in a comment so future maintainers understand potential metric variance under heavy stream teardown conditions.

pkg/agent/client.go
```diff
@@ -431,6 +466,11 @@ func (a *Client) Serve() {
 				}
 				close(eConn.dataCh)
 				a.connManager.Delete(eConn.connID)
+				// The connection was established (conn is non-nil), so record its
```
cheftako: Inside this closure (which implements `eConn.cleanFunc`), the synchronization via `<-dialDone` (found higher up in the same closure, around line 434) is excellent because it guarantees `eConn.establishedAt` is set before being read. Consider adding a brief comment inline calling out that `<-dialDone` prevents reads on uninitialized `establishedAt` values, ensuring future maintainers do not accidentally remove or reorder it.

cheftako (commented):
Also need to get the master test passing

PR #914, opened 2026-09-14
The APIServer Network Proxy Server can work in either GRPC or HTTP-CO…

cheftako:
…NNECT mode. Packet handling for these two modes should be the same. Currently it isn’t. The GRPC mode is the more correct and I would like the code base to be standardized on the current GPRC packet handling. The primary entry point for new requests in GRPC mode is the Proxy method from the ProxyServer class. This can be found in the pkg/server/server.go file. Currently the Proxy method takes a client.ProxyService_ProxyServer object as input. To make this work with HTTP-CONNECT this would need to be converted to an interface which could be implemented by an adapter for HTTP-CONNECT. Once the connection is established the GRPC method which handles it is the Connect method in the ProxyServer Class. This can also be found in the pkg/server/server.go file. Currently the Connect method takes an agent.AgentService_ConnectServer object as input. To make this work with HTTP-CONNECT this would need to be converted to an interface which could be implemented by an adapter for HTTP-CONNECT.

The stack traces for the GRPC calls mentioned above are “sigs.k8s.io/apiserver-network-proxy/pkg/server.(*ProxyServer).Proxy(0x1e6801748a00, {0x102e03608, 0x1e6801495170})
	pkg/server/server.go:552 +0x3c

sigs.k8s.io/apiserver-network-proxy/konnectivity-client/proto/client._ProxyService_Proxy_Handler({0x102d6a4a0?, 0x1e6801748a00}, {0x102dffcb0, 0x1e68014d8100})
	vendor/sigs.k8s.io/apiserver-network-proxy/konnectivity-client/proto/client/client_grpc.pb.go:109 +0xdc

google.golang.org/grpc.(*Server).processStreamingRPC(0x1e6801688248, {0x102dfb348, 0x1e6801768ff0}, 0x1e6801556540, 0x1e680178a840, 0x102ebdd80, 0x0)
	vendor/google.golang.org/grpc/server.go:1729 +0xcc0

google.golang.org/grpc.(*Server).handleStream(0x1e6801688248, {0x102dfbb78, 0x1e68019801a0}, 0x1e6801556540)
	vendor/google.golang.org/grpc/server.go:1862 +0x778

google.golang.org/grpc.(*Server).serveStreams.func2.1()
	vendor/google.golang.org/grpc/server.go:1078 +0x70

created by google.golang.org/grpc.(*Server).serveStreams.func2 in goroutine 32
	vendor/google.golang.org/grpc/server.go:1089 +0x120”
And
“sigs.k8s.io/apiserver-network-proxy/pkg/server.(*ProxyServer).readBackendToChannel(0x3b01977da8a0?, 0x3b01975609c0, 0x3b01975430a0, 0x3b0197543110)
	pkg/server/server.go:957 +0xcc
sigs.k8s.io/apiserver-network-proxy/pkg/server.(*ProxyServer).Connect.func2({0x104ee3348?, 0x3b01977da8a0?})
	pkg/server/server.go:924 +0x24
runtime/pprof.Do({0x104ee3310?, 0x104fefe40?}, {{0x3b01974a38c0?, 0x0?, 0x0?}}, 0x3b01977da870)
	@v0.0.1-go1.26.2.darwin-arm64/src/runtime/pprof/runtime.go:57 +0x78
created by sigs.k8s.io/apiserver-network-proxy/pkg/server.(*ProxyServer).Connect in goroutine 97
	pkg/server/server.go:924 +0x918”

Once the HTTP-CONNECT method is set to use the Proxy and Connect methods, we need to make sure the code compiles, the tests pass and make sure the use case at https://github.com/kubernetes-sigs/apiserver-network-proxy#http-connect-client-using-mtls-proxy-with-dial-back-agent-either-curl-or-test-client works.


pkg/server/tunnel.go
```diff
@@ -77,190 +85,269 @@ func (t *Tunnel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-		klog.ErrorS(errBackendDialTimeout, "backend dial timed out before connection was established", "host", r.Host, "dialID", connection.dialID, "agentID", connection.agentID)
-		metrics.Metrics.ObserveDialFailure(metrics.DialFailureBackendDialTimeout)
-		dialFailureObserved = true
-		conn.Write([]byte(fmt.Sprintf("HTTP/1.1 504 Gateway Timeout\r\nContent-Type: text/plain; charset=utf-8\r\n\r\nBackend dial timeout: %v\r\n", errBackendDialTimeout)))
-		// The deferred cleanup will run when we return here.
-		return
+	select {
+	case <-h.connected:
+	case <-h.closed:
+		return nil, io.EOF
```
ipochi, line 187: blocking: An HTTP-CONNECT request can remain pending if its selected agent disconnects before sending DIAL_RSP. Backend teardown removes established connections but not pending dials, so this path neither removes the pending entry nor closes the HTTP stream. The socket, goroutines, and read buffer can therefore remain retained, creating a memory leak.

pkg/server/tunnel.go
```diff
@@ -77,190 +85,269 @@ func (t *Tunnel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	h.connectID = resp.ConnectID
+	if _, err := h.conn.Write([]byte(connectEstablished)); err != nil {
+		klog.ErrorS(err, "failed to send 200 connection established", "host", h.host,
+			"dialID", h.dialID, "connectionID", resp.ConnectID)
+		h.close()
+		return err
+	}
+	klog.V(3).InfoS("Connection established, sent 200 OK", "host", h.host,
+		"dialID", h.dialID, "connectionID", resp.ConnectID)
+	close(h.connected)
```
ipochi, line 257: blocking: This signals dial completion before the connection is added to established. The pending entry has already been removed, so cleanup can find the connection in neither map.

  https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/c03f684a969639062b597a1f6838d732ea735e74/pkg/server/tunnel.go#L251

  https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/c03f684a969639062b597a1f6838d732ea735e74/pkg/server/server.go#L972-L1007

pkg/server/server.go
```diff
@@ -292,9 +216,40 @@ type AgentTokenAuthenticationOptions struct {
...
+	return g.server.Proxy(stream)
+}
+
+type grpcAgentService struct {
+	server *ProxyServer
+}
+
+func (g *grpcAgentService) Connect(stream agent.AgentService_ConnectServer) error {
+	return g.server.Connect(stream)
+}
```
ipochi: nit: Could we keep Proxy’s original gRPC signature and move its processing code into a private proxy(ProxyStream) helper?

```
  func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
        return s.proxy(stream)
  }
```

HTTP-CONNECT could call that helper directly, while Connect keeps its original signature. This would avoid the extra gRPC service wrappers.

pkg/server/server.go
ipochi, line 469: HTTP-CONNECT now shares an existing gRPC metrics bug: removeEstablishedForStream does not update the connection count or record duration.

I'll create an issue and PR after this merges to add those updates, matching removeEstablished, with a regression test.

pkg/server/tunnel_test.go
```diff
@@ -0,0 +1,344 @@
...
+	// fromAgent is the agent's send queue, consumed by the mocked Recv.
+	fromAgent chan *client.Packet
+	conn      net.Conn
+}
+
+func newHTTPConnectFixture(t *testing.T, ctrl *gomock.Controller, host string) *httpConnectFixture {
+	t.Helper()
+
+	proxyServer := NewProxyServer(uuid.New().String(), []proxystrategies.ProxyStrategy{proxystrategies.ProxyStrategyDefault}, 1, &AgentTokenAuthenticationOptions{}, xfrChannelSize)
+	agentConn, backend := prepareAgentConnMD(t, ctrl, proxyServer, nil)
```
ipochi: prepareAgentConnMD already registers a backend, and serveAgent(line 100) registers another through Connect(in line 105).

Could we create just the mock here and let Connect register the backend before sending the HTTP request? That would ensure requests and disconnect cleanup use the same backend.

pkg/server/server.go
```diff
@@ -557,8 +512,9 @@ func (s *ProxyServer) SetBackendDialTimeout(timeout time.Duration) {
 	s.backendDialTimeout = timeout
 }
 
-// Proxy handles incoming streams from gRPC frontend.
-func (s *ProxyServer) Proxy(stream client.ProxyService_ProxyServer) error {
+// Proxy handles an incoming frontend stream, whether it arrived over gRPC or
+// over HTTP-CONNECT.
+func (s *ProxyServer) Proxy(stream ProxyStream) error {
 	metrics.Metrics.ConnectionInc(metrics.Proxy)
```
ipochi, line 510: ServeHTTP increments http_connections, then the shared Proxy method increments grpc_connections{service_method="Proxy"}.

HTTP-CONNECT now reaches these calls, so it is counted in both the HTTP and gRPC connection gauges. Could we move ConnectionInc and ConnectionDec into grpcProxyService.Proxy, so they count only gRPC frontend connections?

The bug is that HTTP connections are also counted as gRPC connections, not that the gauges leak.

pkg/server/tunnel.go
```diff
@@ -77,190 +85,269 @@ func (t *Tunnel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-		}
-		klog.V(3).InfoS("Connection established, sent 200 OK", "host", r.Host, "agentID", connection.agentID, "connectionID", connection.connectID)
+var _ ProxyStream = &httpConnectStream{}
+
+func newHTTPConnectStream(r *http.Request, conn net.Conn, bufrw *bufio.ReadWriter) *httpConnectStream {
+	// The frontend packet handling expects gRPC style incoming metadata. Give
+	// it the client information the CONNECT request carried. The stream
+	// lifetime is bound to the hijacked connection rather than to the request,
+	// so it does not inherit the request context.
+	ctx := metadata.NewIncomingContext(context.Background(),
```
ipochi: nit: Could we create a cancellable context here and cancel it in httpConnectStream.close? That would make Context() reflect the stream’s lifetime, as it does for gRPC.

pkg/server/tunnel.go
```diff
@@ -77,190 +85,269 @@ func (t *Tunnel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-					ConnectID: connection.connectID,
+	buf := *h.readBuf
+	for {
+		n, err := h.bufrw.Read(buf)
+		if n > 0 {
+			// The packet is queued for the backend and outlives this call, so
+			// it cannot reference the reusable read buffer. A read error that
+			// accompanied the data is reported by the next Recv; bufio retains
+			// it.
+			data := make([]byte, n)
```
ipochi, line 198: non-blocking: A local before/after HTTP-CONNECT upload benchmark shows around 6% lower bulk throughput and nearly double the allocated bytes—could we include the benchmark and investigate reducing this overhead while keeping queued payloads safe from buffer reuse? I'll create an issue later one with the test and comparisons.

ipochi:
Could we add a regression test to protect queued payloads from buffer reuse—the test below passes on `c03f684a` with `-race` and fails when the copy is replaced with `data := buf[:n]`, reporting `first packet changed after second read: got "second", want "first!"`?

<details>
<summary>Test to add to pkg/server/tunnel_test.go — no additional imports needed</summary>

```go
// TestHTTPConnectStreamRecvPreservesPreviouslyReturnedPayload verifies that a
// packet retained by the transfer queue is not overwritten by the next read.
func TestHTTPConnectStreamRecvPreservesPreviouslyReturnedPayload(t *testing.T) {
	conn, frontend := net.Pipe()
	t.Cleanup(func() { frontend.Close() })

	stream := newHTTPConnectStream(
		&http.Request{Host: "127.0.0.1:8080"},
		conn,
		bufio.NewReadWriter(bufio.NewReader(conn), bufio.NewWriter(conn)),
	)
	t.Cleanup(stream.release)
	if err := conn.SetReadDeadline(time.Now().Add(5 * time.Second)); err != nil {
		t.Fatal(err)
	}
	// Exercise established DATA reads independently of the dial handshake.
	stream.dialRequested = true
	stream.connectID = 42
	close(stream.connected)

	payloads := []string{"first!", "second"}
	writeDone := make(chan error, 1)
	go func() {
		for _, payload := range payloads {
			// net.Pipe keeps these writes separate, making buffer reuse deterministic.
			if _, err := io.WriteString(frontend, payload); err != nil {
				writeDone <- err
				return
			}
		}
		writeDone <- nil
	}()
	t.Cleanup(func() {
		frontend.Close()
		if err := <-writeDone; err != nil {
			t.Errorf("writing frontend payloads: %v", err)
		}
	})

	var first *client.Packet
	for i, want := range payloads {
		pkt, err := stream.Recv()
		if err != nil {
			t.Fatalf("Recv %d: %v", i, err)
		}
		if pkt.GetType() != client.PacketType_DATA {
			t.Fatalf("Recv %d: expected DATA, got %v", i, pkt.GetType())
		}
		if got := string(pkt.GetData().Data); got != want {
			t.Fatalf("Recv %d: expected %q, got %q", i, want, got)
		}
		if i == 0 {
			first = pkt
		}
	}
	if got := string(first.GetData().Data); got != payloads[0] {
		t.Fatalf("first packet changed after second read: got %q, want %q", got, payloads[0])
	}
}
```

Run with:

```sh
go test -race ./pkg/server -run '^TestHTTPConnectStreamRecvPreservesPreviouslyReturnedPayload$' -count=1
```

The copy was restored after the failing check; this is regression coverage for the existing correct behavior, not a new corruption bug in the PR.

</details>

ipochi (commented):
I found couple of blocking issues and a potential throughtput issue but looks good otherwise. Once this merges, I will rework #896 based on these changes.

pkg/server/tunnel.go
cheftako: Good catch, this is real and now fixed in d862b422.

The gap was that backend teardown in `serveRecvBackend` only cleaned `established` — the `TODO(#126)` pending-dial case was never handled — and `startPendingDialTimeout` does not cover it either, since its `<-backend.Context().Done()` branch returns without removing the entry or notifying the frontend (and the default dial timeout is 0, so often no timer exists at all). An HTTP-CONNECT `Recv` parked in `select { <-connected; <-closed }` then never woke up, retaining the socket, its goroutines and the pooled read buffer.

Backend teardown now removes the pending dials for that backend connection (`PendingDialManager.removeForBackend`) and fails each one with a `DIAL_RSP` carrying `errBackendClosedWhileDialing`, matching the shape the existing dial-timeout path already uses. The HTTP-CONNECT frontend turns that into a 502 instead of hanging, and gRPC frontends get the same failure they get on a dial timeout.

Covered by `TestBackendCloseFailsPendingDials` and `TestHTTPConnectTunnelAgentCloseWhilePendingDial`; both fail without the fix.

pkg/server/tunnel.go
cheftako: Agreed, and fixed in d862b422.

The root cause is the ordering in `serveRecvBackend` rather than in `close(h.connected)` itself: the dial is removed from `pendingDial`, then sent to the frontend, and only then added to `established`. Any frontend that acts on the `DIAL_RSP` inside that window — by sending, or by shutting down — is found in neither map, and the backend connection leaks. HTTP-CONNECT just makes the window easy to hit, since `close(h.connected)` releases `Recv` immediately.

`addEstablished` now runs before `frontend.send`, with `removeEstablished` on send failure. I also split the `resp.Error != ""` branch out, since a failed dial has no connection to establish, which removes the `dialErr` bookkeeping.

`TestDialResponseEstablishesBeforeReachingFrontend` asserts `getFrontend` succeeds at the moment the `DIAL_RSP` reaches the frontend, and fails against the old ordering.

pkg/server/server.go
cheftako: Done in d862b422 — this is nicer, thanks.

`Proxy` and `Connect` keep their generated gRPC signatures and delegate to private `proxy(ProxyStream)` / `connect(AgentStream)` helpers, which `Tunnel.ServeHTTP` calls directly. That drops both service wrapper types and `GrpcProxyService`/`GrpcAgentService`, so `cmd/server` goes back to registering the `ProxyServer` itself and downstream registration is unchanged.

pkg/server/server.go
cheftako: Agreed, and that one is pre-existing rather than introduced here, so I left it for your follow-up. Thanks for picking it up.

pkg/server/tunnel_test.go
cheftako: Fixed in d862b422.

You are right that it registered two backends for the same agent, so a request could be served by the fixture-registered backend while `Connect` served a different one, and disconnect cleanup would then not match.

`prepareAgentConnMD` is split: `prepareAgentConn` builds just the mock, and `prepareAgentConnMD` keeps its old behaviour for the tests that want a pre-registered backend. The HTTP-CONNECT fixture now uses `prepareAgentConn`, starts `Connect`, waits for the backend to be registered, and only then sends the CONNECT request, so requests and cleanup always use the same backend connection.

pkg/server/server.go
cheftako: Correct, and fixed in d862b422.

`ConnectionInc`/`ConnectionDec(metrics.Proxy)` moved into the gRPC-only `Proxy` method, which now wraps the shared `proxy(ProxyStream)` helper. HTTP-CONNECT frontends reach `proxy` directly and are counted only by `HTTPConnectionInc` in `Tunnel.ServeHTTP`, so `grpc_connections{service_method="Proxy"}` is once again gRPC frontends only.

pkg/server/tunnel.go
cheftako: Done in d862b422. `newHTTPConnectStream` builds the context with `context.WithCancel` before attaching the metadata, and `httpConnectStream.close` cancels it, so `Context()` ends with the stream the way a gRPC stream context does. `TestHTTPConnectStreamContextEndsWithStream` covers it.

pkg/server/tunnel.go
cheftako: Thanks — no change here, leaving the benchmark and the allocation reduction to the issue you are filing. The copy in `Recv` stays for now since queued payloads must survive buffer reuse, which the regression test below pins down.

cheftako:
Thanks for the thorough review @ipochi — all of it was legitimate. Pushed as d862b422 (force-push: the branch is rebased onto current master, so your inline comments may show as outdated).

- **Pending dial leak on agent disconnect** — fixed. Backend teardown now removes the pending dials for that backend connection and fails each with a `DIAL_RSP`, so the HTTP-CONNECT frontend gets a 502 instead of retaining its socket, goroutines and pooled buffer.
- **Dial completion signalled before establish** — fixed by reordering `serveRecvBackend` so `addEstablished` runs before `frontend.send`, with `removeEstablished` on send failure.
- **Keep the gRPC signatures** — done. `Proxy`/`Connect` take the generated types again and delegate to private `proxy(ProxyStream)`/`connect(AgentStream)`; the service wrappers are gone.
- **Double-counted connection gauge** — fixed. `ConnectionInc`/`Dec(Proxy)` moved into the gRPC-only `Proxy`.
- **Cancellable stream context** — done.
- **Fixture registering two backends** — fixed; `Connect` registers the only backend now.
- **`removeEstablishedForStream` metrics** and the **throughput/allocation regression** — left for the follow-up issues you mentioned.

Your buffer-reuse regression test is added verbatim as `TestHTTPConnectStreamRecvPreservesPreviouslyReturnedPayload`. Alongside it: `TestHTTPConnectTunnelAgentCloseWhilePendingDial`, `TestDialResponseEstablishesBeforeReachingFrontend`, `TestBackendCloseFailsPendingDials` and `TestHTTPConnectStreamContextEndsWithStream`. I verified the three new regression tests fail against the unfixed code.

Full unit suite with `-race`, the integration suite, `go vet` and `golangci-lint` all pass. PTAL.

ipochi (approved):
Thank you for addressing the requested changes.

/lgtm

cheftako:
> Thank you for addressing the requested changes.
> 
> /lgtm

Lets get #896 rebased and merged.

PR #917, opened 2026-09-18
server: update established metrics on frontend stream cleanup

ipochi:
Frontend stream cleanup removes established connections without updating the established-connection gauge or recording their duration. This already affects gRPC and also affects HTTP-CONNECT through the shared adapter path.

Observe each removed connection once and refresh the gauge under the existing lock, matching the other established-connection cleanup paths. Cover both frontend modes, unrelated streams, and repeated cleanup.


cheftako:
/lgtm
/approve
Thanks for the change.

PR #918, opened 2026-09-18
server: allow disabling the HTTP CONNECT writer queue

ipochi:
Treat --frontend-write-channel-size=0 as synchronous frontend writes. Skip allocating writer channels and starting the writer goroutine in this mode, while preserving the default queue size of 10 and rejecting negative flag values.

Keep backend-cancellation cleanup active and return socket write errors directly to the sender. Clarify the flag help and latency metric for the disabled-buffering case. Disabling the queue restores immediate shared stream HOL when a frontend write blocks.

Cover synchronous writes, ordered close, frontend/backend cancellation, socket errors, queue metrics, and option validation. Exercise clean agent EOF with and without CLOSE_RSP in buffered and synchronous modes. Retain zero-size integration coverage for requests, concurrent traffic, agent disconnect/reconnect, and multiple agent connections.


cheftako:
/lgtm
/approve
Thanks for turning this around so quickly.

PR #919, opened 2026-09-18
Updated client go.mod to match K/K. 

cheftako:
Updated client go.mod to match https://github.com/kubernetes/kubernetes/blob/release-1.34/go.mod.
A couple of places we are head of 1.34 (and shouldn't be).
Don't want to downgrade, so left those alone.

go get golang.org/x/net@v0.56.0
go mod tidy


cheftako:
Investigating the failure

ipochi:
@cheftako 

This matches Go issue #75031 (https://github.com/golang/go/issues/75031): after switching toolchains, coverage can invoke the wrong Go executable when processing packages without tests.

  I reproduced the coverage failure using Go 1.26.2; the same check passes with Go 1.26.4.

  make test GOTOOLCHAIN=go1.26.4

cheftako:
https://github.com/kubernetes/test-infra/pull/37892

ipochi (approved):
/lgtm. Thank you for fixing the issues and sending the patching in @k8s/test-infra

